Issue #1, opened 2021-11-12
Create cmake project


yhmo:
Create CMakeList.txt and prepare third-partities

PR #8, opened 2021-11-22
Initialize project

yhmo:

#1
#3 
#4


scripts/proto_gen.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
```
matrixji, line 1: I may introduce a cmake way to generate pb.cc from .proto file, after this is merged.

yhmo: Ok

Issue #16, opened 2021-11-24
Proposal discussion: a detail class graph for the sdk


yhmo:
- a proposal to define the client implementation classes
- discuss proposal in this issue
- put the proposal into the [MEP page](https://wiki.lfaidata.foundation/pages/viewpage.action?pageId=52265005) once the proposal is agreed by all collaborators

matrixji:
Do we possibly borrow the interface design from [milvus-sdk-java](https://github.com/milvus-io/milvus-sdk-java)?

yhmo:
> Do we possibly borrow the interface design from [milvus-sdk-java](https://github.com/milvus-io/milvus-sdk-java)?

I am also thinking about this question. @ArkaprabhaChakraborty mentioned we can create a similar orm similar to the python sdk. I think the sdk can be designed as two levels:
- the orm classes: ConnectionInstance/Collection/Partition/Index/Schema/Parameters, and maybe ConnectionPool
- the client implementation: a class to maintain grpc channel, a class to transfer parameters to rpc interface and with some logic like the pymilvus/client/grpc_handler.py

yhmo:
Basically, the client implementation can do all things. Orm classes can be an enhancement for usability.

ArkaprabhaChakraborty:
I'll add some initial orm classes. Their tests and their respective client implementations. I guess that should do. And then I can make further issues which I and @matrixji can divide between ourselves? Also I'll change the MEP. What do you say? My approach might be wrong so please do correct me if so.

yhmo:
@ArkaprabhaChakraborty 
Please attach your proposal here, we can discuss together. 
Just like the "collection alias" feature: https://github.com/milvus-io/milvus/issues/4812, firstly we have a raw proposal, we discuss it in the issue page, finally we put the final design into the MEP page: https://wiki.lfaidata.foundation/display/MIL/MEP+10+--+Support+Collection+Alias

ArkaprabhaChakraborty:
**Class Graph Proposal**

So I was proposing the following structure:

```
------client
         |
         ------- stub
                  |
                  ------- Milvus
         |
         ------- prepare
                  |
                  ------ Prepare
         |
         ------- types
                  |
                  ------  Status
                  |
                  ------- DataType
                  |
                  ------- RangeType
                  |          
                  ------- IndexType
         |
         ------- exceptions
                  |
                  -------  ParamError
                  |
                  ------- ConnectError
                  |
[...]

PR #21, opened 2021-11-29
Add a CrateCollection api for first step

yhmo:

#16

### DIFF
--- a/src/impl/MilvusClientImpl.cpp
+++ b/src/impl/MilvusClientImpl.cpp
@@ -18,8 +18,14 @@
 
 namespace milvus {
 
-void
-MilvusClientImpl::Dummy() {
+std::shared_ptr<MilvusClient>
+MilvusClient::Create() {
+    return std::make_shared<MilvusClientImpl>();
+}
+
+Status
+MilvusClientImpl::CreateCollection(const CollectionSchema& schema) {
+    return Status();
 }
 
 }  // namespace milvus
--- a/src/impl/Status.cpp
+++ b/src/impl/Status.cpp
@@ -1,0 +1,98 @@
+// Copyright (C) 2019-2020 Zilliz. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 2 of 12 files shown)


matrixji:
> We need to add a command to make automatic linting changes. This can be done with gradle, I guess it can be done with cmake too. :)

for cmake generated makefile, you could using:
```
make lint
make check-clang-format
make check-clang-tidy
```
for lint checks

and using:
```
make clang-format
```
for auto format c++ code with clang-format

src/impl/MilvusClientImpl.cpp
```diff
@@ -18,8 +18,15 @@
 
 namespace milvus {
 
-void
-MilvusClientImpl::Dummy() {
+std::shared_ptr<MilvusClient>
+MilvusClient::Create() {
+    return std::shared_ptr<MilvusClient>(new MilvusClientImpl());
```
matrixji, line 23: It's OK, but prefer to use std::make_shared instead of std::shared_ptr<>(new ). std::make_shared is exceptional safe.

ArkaprabhaChakraborty:
I guess this PR is failing for lint checks. Other than that it seems fine :). Also we can follow this format/approach for rest of the client section.

src/impl/Status.cpp
```diff
@@ -0,0 +1,98 @@
...
+namespace milvus {
+
+constexpr int CODE_WIDTH = sizeof(StatusCode);
+
+Status::Status(StatusCode code, const std::string& msg) {
+    // 4 bytes store code
+    // 4 bytes store message length
+    // the left bytes store message string
+    const uint32_t length = (uint32_t)msg.size();
+    auto result = new char[length + sizeof(length) + CODE_WIDTH];
```
matrixji, line 25: IMO, we could use 2 private members to hold the `code` and `msg`, then the compiler will automatically generate the default copy/move constructors.
Maybe we could also do the improvement with other PRs.

src/impl/MilvusClientImpl.cpp
yhmo: you are right

src/impl/Status.cpp
yhmo: This file is copied from the milvus project. I agree use two members to make it simple. Let's change it by another PR.

matrixji: /lgtm

yhmo:
@matrixji 
I just fix the lint warning, need "lgtm" label again.

PR #38, opened 2021-12-03
Introducing mock server for integration testing

matrixji:

### DIFF
--- a/test/it/mocks/MilvusMockedTest.cpp
+++ b/test/it/mocks/MilvusMockedTest.cpp
@@ -1,0 +1,12 @@
+
+#include "MilvusMockedTest.h"
+
+void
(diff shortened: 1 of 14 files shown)


test/it/mocks/MilvusMockedTest.cpp
```diff
@@ -0,0 +1,12 @@
...
+
+void
+milvus::MilvusMockedTest::SetUp() {
+    server_.Start();
+    client_ = milvus::MilvusClient::Create();
+}
+
+void
+milvus::MilvusMockedTest::TearDown() {
+}
```
yhmo, line 12: Should call server_.Stop() here?

matrixji: Yes, I think so, we need to add it while adding testing code later

PR #50, opened 2021-12-06
Handle server side error

yhmo:


matrixji:
/lgtm

Later I'll check if possible to update Status by proto::XyzStatus in MilvusConnection class.

PR #83, opened 2021-12-15
Add Field class.

matrixji:
- Dependency for adding Insert interface.


matrixji:
Prefer using Field from #84

PR #90, opened 2021-12-23
Declare all interfaces except CalcDistance

yhmo:


ArkaprabhaChakraborty:
That's one odd lint error. :)

PR #99, opened 2021-12-25
Implement Query interface

matrixji:
Issue: #81

### DIFF
--- a/src/impl/MilvusClientImpl.cpp
+++ b/src/impl/MilvusClientImpl.cpp
@@ -19,6 +19,7 @@
 #include <chrono>
 #include <thread>
 
+#include "TypeUtils.h"
 #include "common.pb.h"
 #include "milvus.grpc.pb.h"
 #include "milvus.pb.h"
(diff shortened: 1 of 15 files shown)


src/impl/MilvusClientImpl.cpp
```diff
@@ -462,7 +463,39 @@ MilvusClientImpl::Search(const SearchArguments& arguments, const SearchResults&
...
+    }
+
+    rpc_request.set_expr(arguments.Expression());
+    const auto& output_fields = arguments.OutputFields();
+    auto& fields = *(rpc_request.mutable_output_fields());
+    for (auto field : output_fields) {
+        fields.Add(std::move(field));
+    }
+
+    rpc_request.set_travel_timestamp(arguments.GuaranteeTimestamp());
```
yhmo, line 485: Here should be "TravelTimestamp()"

matrixji: Fixed with add argument matcher in testing code.

PR #116, opened 2022-01-10
Refine Search interface

matrixji:

### DIFF
--- a/src/impl/MilvusClientImpl.cpp
+++ b/src/impl/MilvusClientImpl.cpp
@@ -498,7 +498,24 @@ MilvusClientImpl::Delete(const std::string& collection_name, const std::string& 
 
 Status
 MilvusClientImpl::Search(const SearchArguments& arguments, SearchResults& results) {
-    auto pre = [&arguments]() {
+    std::string anns_field;
+    auto validate = [this, &arguments, &anns_field]() {
+        CollectionDesc collection_desc;
+        auto status = DescribeCollection(arguments.CollectionName(), collection_desc);
+        if (status.IsOk()) {
+            // check anns fields
+            auto& field_name = arguments.TargetVectors()->Name();
+            auto anns_fileds = collection_desc.Schema().AnnsFieldNames();
+            if (anns_fileds.find(field_name) != anns_fileds.end()) {
+                anns_field = field_name;
+            } else {
+                return Status{StatusCode::INVALID_AGUMENT, std::string(field_name + " is not a valid anns field")};
+            }
+        }
+        return status;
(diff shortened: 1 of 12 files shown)


src/include/types/CollectionSchema.h
```diff
@@ -107,6 +107,20 @@ class CollectionSchema {
...
     }
 
+    /**
+     * @brief Return Anns field name, if not exist return ""
+     */
+    std::string
+    AnnsFieldName() const {
+        for (const auto& field : fields_) {
+            auto data_type = field.FieldDataType();
+            if (data_type == DataType::BINARY_VECTOR || data_type == DataType::FLOAT_VECTOR) {
```
yhmo, line 117: Currently, each collection only supports one vector field, this function is no problem. But in later version(maybe 2.1), each collection can manage multiple vector fields. This function could be a potential bug in future.

src/impl/MilvusClientImpl.cpp
```diff
@@ -472,7 +472,17 @@ MilvusClientImpl::Delete(const std::string& collection_name, const std::string&
 
 Status
 MilvusClientImpl::Search(const SearchArguments& arguments, SearchResults& results) {
-    auto pre = [&arguments]() {
+    std::string anns_field;
+    auto validate = [this, &arguments, &anns_field]() {
+        CollectionDesc collection_desc;
+        auto status = DescribeCollection(arguments.CollectionName(), collection_desc);
+        if (status.IsOk()) {
+            anns_field = collection_desc.Schema().AnnsFieldName();
```
yhmo, line 480: I suggest add a method "SetAnnsFieldName()" to the SearchArguments class, or change the AddTargetVectors() to be:
`Status AddTargetVector(const std::string& field_name, const FloatVecFieldData::ElementT& vector)`

Then create FieldData with the name, if the FieldData already exists, just return an error. 
Now we only allow search one vector field, in someday maybe we can allow two vectors fields to search together with  different weight.

matrixji: Seems `Status AddTargetVector(const std::string& field_name, const FloatVecFieldData::ElementT& vector)` is more reasonable right now.

src/include/types/CollectionSchema.h
matrixji: I'd like refactor to `std::unorder_set<std::string> AnnsFieldNames()` and use it for validation.

yhmo: Ok, return `std::unordered_set<std::string>` is reasonable.

PR #128, opened 2022-01-21
Fix example bug

yhmo:

### DIFF
--- a/src/include/milvus/MilvusClient.h
+++ b/src/include/milvus/MilvusClient.h
@@ -111,11 +111,12 @@ class MilvusClient {
      *
      * @param [in] collection_name name of the collection
      * @param [in] progress_monitor set timeout to wait loading progress complete, set to ProgressMonitor::NoWait() to
-     * return instantly
+     * return instantly, set to ProgressMonitor::Forever() to wait until finished.
      * @return Status operation successfully or not
      */
     virtual Status
-    LoadCollection(const std::string& collection_name, const ProgressMonitor& progress_monitor) = 0;
+    LoadCollection(const std::string& collection_name,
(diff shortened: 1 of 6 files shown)


src/include/milvus/MilvusClient.h
```diff
@@ -142,13 +143,13 @@ class MilvusClient {
...
      * @param [in] progress_monitor set timeout to wait flush progress complete, set to ProgressMonitor::NoWait() to
-     * return instantly
+     * return instantly, set to ProgressMonitor::Forever() to wait until finished.
      * @param [out] collection_stat statistics of the collection
      * @return Status operation successfully or not
      */
     virtual Status
-    GetCollectionStatistics(const std::string& collection_name, const ProgressMonitor& progress_monitor,
-                            CollectionStat& collection_stat) = 0;
+    GetCollectionStatistics(const std::string& collection_name, CollectionStat& collection_stat,
```
matrixji, line 151: Opt1:
```
virtual Status GetCollectionStatistics(const std::string& collection_name, CollectionStat& collection_stat) = 0;
virtual Status GetCollectionStatistics(const std::string& collection_name, CollectionStat& collection_stat, const ProgressMonitor& progress_monitor) = 0;
```
Opt2:
```-google-default-arguments``` in .clang-tidy.

IMO, opt2 is easier, and "not using default arguments in virtual" may have problems with old legacy compilers, but we are already with c++11, it should be ok.

yhmo: Not sure why the "NOLINT" could not work: https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics

Will try the "-google-default-arguments"

PR #138, opened 2022-02-04
Add DescribeIndex Interface

### LINKED ISSUE #64: Implement DescribeIndex() interface

ArkaprabhaChakraborty:
fixes #64 
Signed-off-by: ArkaprabhaChakraborty <chakrabortyarkaprabha998@gmail.com>

### DIFF
--- a/src/impl/MilvusClientImpl.cpp
+++ b/src/impl/MilvusClientImpl.cpp
@@ -484,7 +484,34 @@ MilvusClientImpl::CreateIndex(const std::string& collection_name, const IndexDes
 Status
 MilvusClientImpl::DescribeIndex(const std::string& collection_name, const std::string& field_name,
                                 IndexDesc& index_desc) {
-    return Status::OK();
+    auto pre = [&collection_name, &field_name]() {
+        proto::milvus::DescribeIndexRequest rpc_request;
+        rpc_request.set_collection_name(collection_name);
+        rpc_request.set_field_name(field_name);
+        return rpc_request;
+    };
+
+    auto post = [&index_desc](const proto::milvus::DescribeIndexResponse& response) {
+        auto count = response.index_descriptions_size();
+        for (size_t i = 0; i < count; ++i) {
+            auto& field_name = response.index_descriptions(i).field_name();
+            auto& index_name = response.index_descriptions(i).index_name();
+            index_desc.SetFieldName(field_name);
+            index_desc.SetIndexName(index_name);
+            std::unordered_map<std::string, std::string> index_params;
+            auto index_params_size = response.index_descriptions(i).params_size();
+            index_params.reserve(index_params_size);
+
+            for (size_t j = 0; j < index_params_size; ++j) {
+                index_params.emplace(response.index_descriptions(i).params(j).key(),
+                                     response.index_descriptions(i).params(j).value());
+            }
+            index_desc.SetParams(index_params);
+        }
+    };
+
+    return apiHandler<proto::milvus::DescribeIndexRequest, proto::milvus::DescribeIndexResponse>(
+        pre, &MilvusConnection::DescribeIndex, post);
 }
 
 Status
--- a/test/it/TestDescribeIndex.cpp
+++ b/test/it/TestDescribeIndex.cpp
@@ -1,0 +1,57 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


ArkaprabhaChakraborty:
@matrixji I don't know what I am doing wrong. Any help?

ArkaprabhaChakraborty:
@matrixji What should I add to the tests to pass codecov test?

src/impl/MilvusClientImpl.cpp
```diff
@@ -484,7 +484,39 @@ MilvusClientImpl::CreateIndex(const std::string& collection_name, const IndexDes
...
+    auto pre = [&collection_name, &field_name]() {
+        proto::milvus::DescribeIndexRequest rpc_request;
+        rpc_request.set_collection_name(collection_name);
+        rpc_request.set_field_name(field_name);
+        return rpc_request;
+    };
+
+    auto post = [&index_desc](const proto::milvus::DescribeIndexResponse& response) {
+        auto count = response.index_descriptions_size();
+        if (count > 0) {
```
matrixji, line 496: no need to check `if (count > 0)`, the for statement already has such logic test.

src/impl/MilvusClientImpl.cpp
```diff
@@ -484,7 +484,39 @@ MilvusClientImpl::CreateIndex(const std::string& collection_name, const IndexDes
...
+            for (size_t i = 0; i < count; ++i) {
+                auto field_name = response.index_descriptions(i).field_name();
```
matrixji, line 498: Prefer using reference, `auto &field_name = ....`, and alao for index_name

src/impl/MilvusClientImpl.cpp
```diff
@@ -484,7 +484,39 @@ MilvusClientImpl::CreateIndex(const std::string& collection_name, const IndexDes
...
+                auto index_name = response.index_descriptions(i).index_name();
+                if (field_name.length() > 0 and index_name.length() > 0) {
```
matrixji, line 500: Prefering using `(! field_name.empty()) && (! index_name.empty())`

test/it/TestDescribeIndex.cpp
```diff
@@ -0,0 +1,52 @@
...
+    DescribeIndexRequest request;
+    request.set_collection_name(collection_name);
+    request.set_field_name(field_name);
+
+    EXPECT_CALL(service_, DescribeIndex(_,
+                                        AnyOf(Property(&DescribeIndexRequest::collection_name, collection_name),
+                                              Property(&DescribeIndexRequest::field_name, field_name)),
+                                        _))
+        .WillOnce([](::grpc::ServerContext*, const DescribeIndexRequest*, const DescribeIndexResponse*) {
+            return ::grpc::Status{};
```
matrixji, line 46: If you want to pass the test like `EXPECT_EQ(index_desc.FieldName(), ...)`, you need to set the expected response here.

matrixji:
> @matrixji any idea why the tests are failing? What mistake did I make?

Sorry for being late, as we have a long Spring Festival holiday, I may occasionally check my mail. Mainly you need to add some expected response in the mocked function to pass the test. I've dropped some comments on your PR.

test/it/TestDescribeIndex.cpp
ArkaprabhaChakraborty: How Do I set the response? There aren't any available methods to do that.

ArkaprabhaChakraborty:
I tried what you said ... But for some reason the field name is being passed as a null string

test/it/TestDescribeIndex.cpp
matrixji: `.WillOnce([](::grpc::ServerContext*, const DescribeIndexRequest*, const DescribeIndexResponse*) {` 
should be 
`.WillOnce([](::grpc::ServerContext*, const DescribeIndexRequest*, DescribeIndexResponse* response) {`

ArkaprabhaChakraborty: Okay got it .

ArkaprabhaChakraborty: But I still can't change the response as index_descriptions provided in the response is const type.

ArkaprabhaChakraborty: the proto files don't have any methods to set the index_descriptions in a response

matrixji: > index_descriptions

For changing proto generated member structure, you should use mutable_member(), like, response->mutable_index_descriptions().

>
> Bar* mutable_foo(): Returns a pointer to the mutable Bar object that stores the field's value. If the field was not set prior to the call, then the returned Bar will have none of its fields set (i.e. it will be identical to a newly-allocated Bar). After calling this, has_foo() will return true and foo() will return a reference to the same instance of Bar.
> 

more ref: https://developers.google.com/protocol-buffers/docs/reference/cpp-generated

ArkaprabhaChakraborty: I got it. It wasn't mutable methods but add methods. Since the response didn't have any it was setting an auto index_descriptor with empty values.

ArkaprabhaChakraborty:
@matrixji I think it's done

PR #167, opened 2022-03-15
Add milvus-proto as proto source

yhmo:
Signed-off-by: groot <yihua.mo@zilliz.com>


.github/workflows/main.yaml
```diff
@@ -65,6 +65,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Update submodule
```
matrixji, line 68: checkout action support with a `submodules`  parameter to enable submodule checkout. seems like:
```
- name: checkout
  uses: actions/checkout@v2
  with:
     submodules: 'true'
```

yhmo: Seems submodules: 'true' only support SSH mode.

yhmo:
The checkout step downloads/extracts a tar file from the repo, the .git file is not included in the tar file, "git" command cannot work.

```
Run actions/checkout@v2
/usr/bin/docker exec  72284533b12807da357c6933f3a7c1e3bdcf6c5355e4cab44fd94008f373ff95 sh -c "cat /etc/*release | grep ^ID"
Syncing repository: milvus-io/milvus-sdk-cpp
Getting Git version info
Deleting the contents of '/__w/milvus-sdk-cpp/milvus-sdk-cpp'
The repository will be downloaded using the GitHub REST API
To create a local Git repository instead, add Git 2.18 or higher to the PATH
Downloading the archive
(node:[16](https://github.com/milvus-io/milvus-sdk-cpp/runs/5563557039?check_suite_focus=true#step:3:16)) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
Writing archive to disk
Extracting the archive
/bin/tar xz -C /__w/milvus-sdk-cpp/milvus-sdk-cpp/6750b9f6-0cf3-4709-b86e-67[27](https://github.com/milvus-io/milvus-sdk-cpp/runs/5563557039?check_suite_focus=true#step:3:27)9ce60ffd -f /__w/milvus-sdk-cpp/milvus-sdk-cpp/6750b9f6-0cf3-4709-b86e-67279ce60ffd.tar.gz
Resolved version milvus-io-milvus-sdk-cpp-c90783e
```

matrixji:
Seems failed due to running checkout@v2 under the container

For action/checkout@v2 or v3:
> Fallback to REST API download
>   - When Git 2.18 or higher is not in the PATH, the REST API will be used to download the files
>   - When using a job container, the container's PATH is used

For checkout with submodules under the container, seems using action/checkout@v1 seems the simplest solution.

PoC Ref: 
- https://github.com/matrixji/milvus-sdk-cpp/runs/5572272549
- https://github.com/matrixji/milvus-sdk-cpp/commit/ed1be4f8c80a15f477a1d72dc6df6c1c11114fe1

src/proto-gen/proto-gen.cmake
```diff
@@ -17,24 +17,24 @@
 function(add_proto_source name)
     add_custom_command(
         OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/proto-gen/${name}.pb.cc
-        DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/proto/${name}.proto
+        DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/milvus-proto/${name}.proto
```
matrixji, line 20: There is still a subfolder `proto` under `milvus-proto`.

yhmo: right..... should be ok now

Issue #191, opened 2022-08-03
Promote proto to version 2.1


matrixji:
As milvus 2.1 was released, should update the proto submodule to version 2.1 in the main branch.

PR #192, opened 2022-08-03
Promote proto to latest version.

matrixji:
Fixes: #191


ArkaprabhaChakraborty:
Ayee! Back to the old days :)...

PR #206, opened 2022-12-04
Update proto and st with milvus 2.1.4

matrixji:


ArkaprabhaChakraborty:
Can we configure an automated bump action which does these changes? Whenever a new proto version is released the bump action can automatically make these bump PRs? Can we open a discussion for this? :).

ArkaprabhaChakraborty:
Can we configure an automated bump action which does these changes? Whenever a new proto version is released the bump action can automatically make these bump PRs? Can we open a discussion for this? :).

matrixji:
> Can we configure an automated bump action which does these changes? Whenever a new proto version/image tag is released the bump action can automatically make these bump PRs? Can we open a discussion for this? :).

Great Idea, I guess we could try to add some action on proto's repo. I think we could have more discussions on it.

Issue #204, opened 2022-09-21
Support connect server with user/password.


matrixji:
We need support if the server is configured with AUTH on.

- Support connect with user and password.
- ST for checking user/password works correctly with the real server.

PR #215, opened 2023-03-02
feat: support username with password

matrixji:
Fixes: #204


ArkaprabhaChakraborty:
The coverage went down... Is this a WIP?

matrixji:
> The coverage went down... Is this a WIP?

Just not covered by integration testing, Now it is fixed, and covered by both `it` and `st` cases.

Issue #266, opened 2024-06-12
Can't use system installed gRPC


nfrmtk:
Relates: #255 
Right now error is following:
```
  By not providing "Findgrpc.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "grpc", but
  CMake did not find one.

  Could not find a package configuration file provided by "grpc" with any of
  the following names:

    grpcConfig.cmake
    grpc-config.cmake

  Add the installation prefix of "grpc" to CMAKE_PREFIX_PATH or set
  "grpc_DIR" to a directory containing one of the above files.  If "grpc"
  provides a separate development package or SDK, be sure it has been
  installed.
```

yhmo:
The code change of #255 has been merged to this pr: https://github.com/milvus-io/milvus-sdk-cpp/pull/273

Issue #255, opened 2024-01-31
spelling mistake in cmake


ausk:
[ThirdPartyPackages](https://github.com/milvus-io/milvus-sdk-cpp/blob/master/cmake/ThirdPartyPackages.cmake#L48C8-L48C24)

```
if ("${MILVUS_WITH_GRPC}" STREQUAL "pakcage")
    find_package(grpc REQUIRED)
else ()
```

PR #267, opened 2024-06-12
fix cmake in case of preinstalled grpc

nfrmtk:
Relates: #266 #255 
After fixing it with `grpc` -> `gRPC`
I got new error about gRPC not being able to find Protobuf, fixed by adding line [49](https://github.com/milvus-io/milvus-sdk-cpp/compare/master...nfrmtk:cmake-grpc-fix?expand=1#diff-dba6098bfd74704b082d3dac547c8fd4f225d0b3e6456973de1d71551bf6718fR49)
And then generation started failing, so I've changed [this](https://github.com/milvus-io/milvus-sdk-cpp/compare/master...nfrmtk:cmake-grpc-fix?expand=1#diff-b22bac7c68896680fe0a55e1992c6190db0d0340716231738d855dfe3b4c038dR35) and [this](https://github.com/milvus-io/milvus-sdk-cpp/compare/master...nfrmtk:cmake-grpc-fix?expand=1#diff-b22bac7c68896680fe0a55e1992c6190db0d0340716231738d855dfe3b4c038dR39)

### DIFF
--- a/cmake/MilvusProtoGen.cmake
+++ b/cmake/MilvusProtoGen.cmake
@@ -32,11 +32,11 @@ set(PROTO_BINARY_DIR "${milvus_proto_BINARY_DIR}")
 set(PROTO_IMPORT_DIR "${milvus_proto_SOURCE_DIR}/proto")
 
 # resolve protoc, always use the protoc in the build tree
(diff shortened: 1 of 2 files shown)


Mohamad-Hussein:
Thanks @nfrmtk , I've tried building with an existing gRPC installation before, and this solution worked for me!

PR #283, opened 2025-06-01
Update to grpc 1.59.1

yhmo:


yhmo:
Build error in windows:
milvus-sdk-cpp\build\_deps\grpc-src\third_party\upb\upb\message\message.c(40): error C2099: initializer is not a constant

PR #288, opened 2025-06-09
add field data: json

hezheSt:


yhmo:
@hezheSt 
To use JSON field, you can switch to 2.4 branch, json field is already implemented.
Once the 2.4 branch is done, we will merge 2.4 back to master.

PR #290, opened 2025-06-18
support dynamic field

hezheSt:


yhmo:
Hi @hezheSt 
No need to add the member "isDyanmic" in FieldSchema.
In the server source code, the proxy always throws an error if a FieldSchema.IsDynamic is true:
https://github.com/milvus-io/milvus/blob/e2c85eec814767b8e28590ff19c25014501e5208/internal/proxy/util.go#L675

In pymilvus, the FieldSchema has a member "is_dynamic" for history reason, actually it is not necessary.
Just use the CollectionSchema.enable_dynamic_field to tell the server to add a dynamic field. The client no need to define a dynamic field.

yhmo:
Hi @hezheSt 
There are more comments about the MilvusClientImpl.Insert().
In Python SDK, the insert() interface only accepts row-based data. In Java SDK MilvusClientV1, the insert supports both row-based and column-based data.

In Python SDK, the row-based insert() creates a JSON field named "$meta" to pass all the redundant values: https://github.com/milvus-io/pymilvus/blob/d55ee6d54526480a614543f51fc8952bce459df2/pymilvus/client/prepare.py#L446

In Java SDK MilvusClientV1, the column-based insert(), user needs to explicitly input a Field named "$meta" to pass dynamic data: https://github.com/milvus-io/milvus-sdk-java/blob/f60245e6c3c9872f212293900ee551b7a0140778/sdk-core/src/main/java/io/milvus/param/ParamUtils.java#L625

Now the Cpp SDK insert() only supports column-based data, so we can follow the Java SDK to process dynamic field.
The major logic is:
1. for each FieldDataPtr, if its name is "$meta", convert it as a proto::schema::FieldData with dynamic=true.
2. for other FieldDataPtr, convert to normal proto::schema::FieldData and pass to server

PR #291, opened 2025-06-30
add GetLoadState interface

hezheSt:


yhmo:
@hezheSt 
The getLoadState() basically has two parameters: **collection_name** and **partition_name**.
partition_name is optional; if partition_name is empty, this api will get the collection's load state, else get a partition's load state.
This is python sdk:
https://github.com/milvus-io/pymilvus/blob/56ff408ccfcef71d29da6e8bc412c52d641debe0/pymilvus/milvus_client/milvus_client.py#L988

Please add a parameter **partition_name** to the interface.

PR #464, opened 2026-04-17
Run ci on macos-15

yhmo:


.github/workflows/main.yaml
```diff
@@ -173,9 +173,15 @@ jobs:
           python3 -m venv ~/.venv
           source ~/.venv/bin/activate
           sh scripts/install_deps.sh
-      - name: Unit Testing
+      - name: Unit Testing (no-conan)
+        if: ${{ matrix.macos == 14 }}
```
liliu-z: Test path selection is coupled to a hardcoded macOS version number (`matrix.macos == 14`). Adding a new macOS version will silently fall into the conan path. Use an explicit matrix field (e.g. `build_mode`) via `matrix.include` instead.

PR #467, opened 2026-04-22
Fix package workflow

yhmo:


.github/workflows/release.yaml
```diff
@@ -20,19 +20,13 @@ jobs:
...
-          - distro: Fedora
-            version: 38
-            image: fedora:38
-            key: fc38
           - distro: Fedora
             version: 39
             image: fedora:39
             key: fc39
     steps:
+      - name: Env
```
liliu-z, line 34: Dropping two distros from the release matrix reduces the set of officially published binaries. This should be called out in the PR description and release notes, not silently removed.

.github/workflows/release.yaml
```diff
@@ -20,19 +20,13 @@ jobs:
...
-            image: ubuntu:22.04
-            key: u2204
-          - distro: Fedora
-            version: 38
-            image: fedora:38
-            key: fc38
           - distro: Fedora
             version: 39
             image: fedora:39
             key: fc39
```
liliu-z, line 32: Fedora 39 is EOL (since 2024-11) just like the removed Fedora 38. It should be replaced with a currently supported version (e.g. Fedora 40 or 41).

PR #472, opened 2026-04-28
Add BatchDescribeCollections/DescribeReplicas interfaces

yhmo:


src/impl/MilvusClientV2Impl.cpp
```diff
@@ -34,6 +34,35 @@
 #include "utils/TypeUtils.h"
 
 namespace milvus {
+namespace {
+
+Status
```
liliu-z, line 39: `ConvertDescribeCollectionResponse` does not check the per-item `status` field of each nested `DescribeCollectionResponse`. When the server returns an error for a specific collection, the failed response is silently converted into an empty `CollectionDesc` instead of propagating the error. The check must handle both `status.code()` and the legacy `status.error_code()` fields (matching the logic in `StatusByProtoResponse`). Fix: add status validation at the top of `ConvertDescribeCollectionResponse` that checks both code paths and skips or reports failed items.

PR #474, opened 2026-04-29
Add RefreshLoad/Optimize interface

yhmo:


src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1946,6 +2040,253 @@ MilvusClientV2Impl::Compact(const CompactRequest& request, CompactResponse& resp
...
+                                         .WithTargetSize(target_size_mb);
+    CompactResponse compact_response;
+    status = Compact(compact_request, compact_response);
+    if (!status.IsOk()) {
+        return finish(status);
+    }
+    response.SetCompactionID(compact_response.CompactionID());
+
+    task.AddProgress("waiting for compaction");
+    for (;;) {
```
liliu-z, line 2232: `runOptimize()` polls compaction state in a loop (lines 2233-2250) but only exits on `EXECUTING` (continue) and `COMPLETED` (break). If the server returns `UNKNOWN` or any other unexpected state, the loop continues indefinitely. Combined with the default gRPC deadline of 0 (no timeout), this can permanently block the background thread. Add an explicit check for unrecognized states and exit with an error status after a bounded number of retries or elapsed time.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1946,6 +2040,253 @@ MilvusClientV2Impl::Compact(const CompactRequest& request, CompactResponse& resp
...
+        auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start);
+        auto remaining = request.TimeoutMs() - elapsed.count();
+        if (remaining <= 0) {
+            return Status{StatusCode::TIMEOUT, "Optimize timeout"};
+        }
+        timeout_ms = remaining;
+        return Status::OK();
+    };
+
+    auto wait_interval = [&check_cancelled, &check_timeout]() {
```
liliu-z: `TimeoutMs` set on `OptimizeRequest` (line 2112) is only checked between steps, but the 6 intermediate synchronous RPCs at lines 2141, 2159, 2187, 2226, 2236, 2262 each issue a gRPC call without any per-call deadline. Since the default gRPC deadline in `ConnectParam.h:346` is 0 (meaning no deadline), any single RPC can hang indefinitely, making the user-facing timeout ineffective. Either propagate a per-RPC deadline derived from the remaining timeout budget, or document that `TimeoutMs` only governs inter-step checks.

src/impl/types/OptimizeTask.cpp
```diff
@@ -0,0 +1,149 @@
...
+Status
+OptimizeTask::TaskStatus() const {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return status_;
+}
+
+void
+OptimizeTask::Start(Worker worker) {
+    auto self = shared_from_this();
+    worker_ = std::thread([self, worker]() {
```
liliu-z: `OptimizeTask::Start()` at line 104 constructs a `std::thread` which can throw `std::system_error` if the OS cannot create the thread. The caller `Optimize()` at `MilvusClientV2Impl.cpp:2048-2055` only catches `std::bad_weak_ptr`, so a thread creation failure will propagate an uncaught exception to the user. Every other SDK method returns errors via `Status`; this is the only path that can throw.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1942,12 +1989,331 @@ MilvusClientV2Impl::Compact(const CompactRequest& request, CompactResponse& resp
...
+
+    std::shared_ptr<MilvusClientV2Impl> self;
+    try {
+        self = shared_from_this();
+    } catch (const std::bad_weak_ptr&) {
+        return {StatusCode::UNKNOWN_ERROR, "MilvusClientV2Impl must be owned by std::shared_ptr to run Optimize"};
+    }
+
+    auto task_copy = task;
+    auto status = task->Start([self, request, task_copy](OptimizeResponse& response) {
```
liliu-z, line 2016: The new `Optimize` API spawns a detached background thread (`runOptimize`) that repeatedly invokes gRPC calls through `ConnectionHandler::apiHandler()`, which reads `connection_`, `ConnectParam::RpcDeadlineMs()`, and `retry_param_` without synchronization (ConnectionHandler.h L169-199). Meanwhile, the main thread can concurrently call `Disconnect()` (which does `stub_.reset()` at MilvusConnection.cpp L181), `Connect()`, `UseDatabase()`, `SetRpcDeadlineMs()`, or `SetRetryParam()` — all without locks. This is the first time the SDK itself creates a background thread, turning previously single-threaded-safe shared state into a real data race surface with potential use-after-free on `stub_`. Fix: at minimum, document that users must not call connection-mutating methods while an async Optimize is active; preferably add synchronization or check for active tasks in Disconnect/UseDatabase.

src/impl/types/OptimizeTask.cpp
```diff
@@ -0,0 +1,158 @@
...
+    } else {
+        cv_.wait(lock, [this] { return done_; });
+    }
+
+    response = response_;
+    return status_;
+}
+
+bool
+OptimizeTask::Cancel() {
```
liliu-z, line 46: Cancel() (L49-66) sets `done_=true` and `cancelled_=true` immediately, but the detached worker thread (started at L103-111) only checks `ShouldCancel()` between RPC steps. If `OptimizeRequest::TimeoutMs()` is 0 and `ConnectParam::RpcDeadlineMs()` is also 0 (the defaults), the current gRPC call has no deadline and can block indefinitely, keeping the thread and its `shared_ptr<MilvusClientV2Impl>` alive. The same issue affects sync mode: `GetResult()` at MilvusClientV2Impl.cpp L2016-2018 may return TIMEOUT while the background thread continues running. Users see 'completed/cancelled' but resources are not released. Fix: enforce a finite RPC deadline for all Optimize-internal calls, or do not set `done_=true` until the worker actually exits.

src/include/milvus/MilvusClientV2.h
```diff
@@ -273,6 +277,17 @@ class MilvusClientV2 {
...
+    /**
+     * @brief Refresh loaded collection data in query node.
+     * This API loads newly generated segments without changing the collection load options.
+     * If the request is sync mode, this API waits until the refresh is complete.
+     *
+     * @param [in] request input parameters
+     * @return Status operation successfully or not
+     */
+    virtual Status
+    RefreshLoad(const RefreshLoadRequest& request) = 0;
```
liliu-z, line 289: Adding pure virtual `RefreshLoad()` (L289) and `Optimize()` (L863) to the public abstract class `MilvusClientV2` is a source and ABI break. Any out-of-tree subclass, mock, or test double will fail to compile, and old binaries with a different vtable layout are incompatible. While this matches the project's existing pattern (PRs #470, #472 did the same), it is still a breaking change that should be acknowledged in release notes or handled via a major version bump if the project intends to provide compatibility guarantees.

src/impl/types/OptimizeTask.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+#include <chrono>
+#include <system_error>
+#include <utility>
+
+namespace milvus {
+
+OptimizeTask::OptimizeTask() = default;
+
+OptimizeTask::~OptimizeTask() {
```
liliu-z: The destructor's `if (worker_.joinable()) worker_.detach()` (L27-30) can never execute: `Start()` already calls `worker_.detach()` at L111, making `joinable()` permanently false. If thread construction throws (L112 catch), `worker_` remains default-constructed, also not joinable. This dead branch may mislead maintainers into thinking there is fallback cleanup logic.

src/impl/utils/MiscUtils.cpp
```diff
@@ -0,0 +1,101 @@
...
+    target_size_mb = 0;
+    normalized.clear();
+
+    auto text = Trim(target_size);
+    if (text.empty()) {
+        return Status::OK();
+    }
+
+    char* end = nullptr;
+    const auto number = std::strtold(text.c_str(), &end);
```
liliu-z: `std::strtold()` uses the current C locale for decimal parsing. Under locales that use comma as decimal separator (e.g. `de_DE`, `fr_FR`), input like `"1.5GB"` will be parsed as `number=1` with `end` pointing to `".5GB"`, which won't match any known unit suffix and returns `INVALID_ARGUMENT`. This makes `OptimizeRequest::TargetSize` behavior environment-dependent. Consider using `std::from_chars` (C++17, locale-independent) or `strtold_l` with an explicit C locale.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1942,12 +1989,334 @@ MilvusClientV2Impl::Compact(const CompactRequest& request, CompactResponse& resp
...
+        }
+        status = getCompactionState(state_request, state_response, rpc_timeout_ms);
+        if (!status.IsOk()) {
+            return finish(status);
+        }
+        if (state_response.State().FailedPlan() > 0) {
+            return finish({StatusCode::SERVER_FAILED, "Compaction failed"});
+        }
+
+        auto compaction_state = state_response.State().State();
```
liliu-z, line 2248: `getCompactionState` post handler (lines 2331-2340) maps all proto states except `Completed` and `Executing` to `UNKNOWN` (including legitimate queued/transitional states). `runOptimize` only allows 3 consecutive UNKNOWN states before returning `SERVER_FAILED`. For large-scale compaction where queue time may exceed 1.5 seconds, this causes premature failure. Additionally, if Milvus proto adds new states (e.g. `Queued`), the SDK will misinterpret them as errors. Consider increasing retry count significantly or treating UNKNOWN as a legitimate transitional state.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1942,12 +1989,334 @@ MilvusClientV2Impl::Compact(const CompactRequest& request, CompactResponse& resp
...
+
+    auto task_copy = task;
+    auto status = task->Start([self, request, task_copy](OptimizeResponse& response) {
+        return self->runOptimize(request, *task_copy, response);
+    });
+    if (!status.IsOk()) {
+        return status;
+    }
+
+    if (!request.Async()) {
```
liliu-z: When `Async()==false`, `GetResult(timeout_ms)` and the internal `runOptimize` check_timeout independently track the same `TimeoutMs()`. If `GetResult` times out first (due to thread launch delay), the caller receives TIMEOUT but the detached background thread continues executing compaction/refresh load operations. The overrun is bounded (~50ms + 1 RPC) since both use the same budget, but the semantic inconsistency means the caller believes the operation stopped when it hasn't.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1942,12 +1989,325 @@ MilvusClientV2Impl::Compact(const CompactRequest& request, CompactResponse& resp
...
+    if (!status.IsOk()) {
+        return finish(status);
+    }
+    if (load_state_response.State() == LoadState::LOAD_STATE_LOADED) {
+        status = check_task_active();
+        if (!status.IsOk()) {
+            return finish(status);
+        }
+        task.AddProgress("refreshing load");
+        status = remaining_rpc_timeout_ms(rpc_timeout_ms);
```
liliu-z, line 2287: When `OptimizeRequest::TimeoutMs() == 0` (no overall timeout), `runOptimize()` falls back to `DEFAULT_OPTIMIZE_RPC_TIMEOUT_MS = 60000` as the per-RPC deadline (line 2071), then passes this same value to `RefreshLoadRequest::TimeoutMs()` (lines 2282, 2290). `refreshLoad()` converts it to `ProgressMonitor{60}` (line 343). The root cause is semantic conflation: `rpc_timeout_ms` (single gRPC call deadline) is reused as the entire refresh operation timeout. Result: user requests "wait indefinitely" but refresh fails after 60s — a realistic scenario for large collections. Fix: decouple the per-RPC timeout from the ProgressMonitor timeout; when no overall timeout is set, use `ProgressMonitor::Forever()` for the refresh phase.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -353,6 +321,45 @@ MilvusClientV2Impl::LoadCollection(const LoadCollectionRequest& request) {
...
+    };
+
+    if (!request.Sync()) {
+        return connection_.InvokeWithRpcTimeout<proto::milvus::LoadCollectionRequest, proto::common::Status>(
+            rpc_timeout_ms, pre, &MilvusConnection::LoadCollection);
+    }
+
+    ProgressMonitor progress_monitor = ProgressMonitor::Forever();
+    if (request.TimeoutMs() > 0) {
+        progress_monitor = ProgressMonitor{static_cast<uint32_t>(request.TimeoutMs() + 999) / 1000};
```
liliu-z, line 345: When an overall timeout is set, `runOptimize()` tracks a millisecond-precision remaining budget (lines 2057, 2075), but `refreshLoad()` rounds it up to whole seconds for `ProgressMonitor` (line 345), and `WaitForStatus()` sleeps in 500ms ticks (ConnectionHandler.cpp:156). If only a few hundred milliseconds remain when entering refresh, the actual wait can exceed the budget by up to ~1 second. This is a separate issue from the 60s hard-limit bug — it affects the "has timeout" path's precision.

examples/src/v2/optimize.cpp
```diff
@@ -0,0 +1,154 @@
...
+    std::cout << "  Status: " << optimize_response.StatusText() << std::endl;
+    std::cout << "  Compaction ID: " << optimize_response.CompactionID() << std::endl;
+    std::cout << "  Progress:";
+    for (const auto& progress : optimize_response.ProgressHistory()) {
+        std::cout << " " << progress;
+    }
+    std::cout << std::endl;
+
+    std::cout << "========== Step 7: Query segment info (after optimize) ==========" << std::endl;
+    while (true) {
```
liliu-z, line 155: After Optimize completes, the example enters `while (true)` polling that only exits when `segment_count == 1`. The Optimize API does not guarantee consolidation to exactly 1 segment — compaction behavior depends on segment sizes, target_size, and server policy. If the final count is not 1, this example loops forever. As user-facing sample code, this sets incorrect expectations.

yhmo: it is expected

src/impl/MilvusClientV2Impl.cpp
yhmo: no need to fix

src/impl/types/OptimizeTask.cpp
```diff
@@ -0,0 +1,148 @@
...
+OptimizeTask::TaskStatus() const {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return status_;
+}
+
+Status
+OptimizeTask::Start(Worker worker) {
+    auto self = shared_from_this();
+    try {
+        worker_ = std::thread([self, worker]() {
```
liliu-z, line 97: The `worker(response)` call inside the detached thread (lines 96-99) has no try-catch. If any internal operation (`std::unordered_set::insert`, proto allocation, `std::string` copy, etc.) throws (e.g. `std::bad_alloc`), C++ guarantees `std::terminate` is called, crashing the entire process. Additionally, `Complete()` is never called, so any caller waiting on `GetResult()` without a timeout will block forever.

Fix: Wrap the lambda body in try-catch, and call `self->Complete(error_status, ...)` in the catch block.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1942,12 +1989,330 @@ MilvusClientV2Impl::Compact(const CompactRequest& request, CompactResponse& resp
...
+    if (!status.IsOk()) {
+        return finish(status);
+    }
+    status = check_task_active();
+    if (!status.IsOk()) {
+        return finish(status);
+    }
+
+    task.AddProgress("compacting");
+    CompactRequest compact_request = CompactRequest()
```
liliu-z, line 2216: The `runOptimize` method constructs a `CompactRequest` setting only `db/collection/target_size` (line 2216-2219) but never sets `ClusteringCompaction`. The underlying `compact()` supports `majorcompaction` via `CompactRequest::ClusteringCompaction()`, but `OptimizeRequest` has no corresponding field. As a result, `Optimize()` can only trigger regular compaction. If this is intentional scope limitation, it should be documented; otherwise the API capability gap needs to be closed.

test/ut/types/TestOptimizeTask.cpp
```diff
@@ -0,0 +1,44 @@
+// Licensed to the LF AI & Data foundation under one
```
liliu-z, line 1: Tests only cover timeout and cancel flag scenarios. Missing: (1) Start + normal async completion verifying GetResult returns correct result, (2) Cancel after Start then await completion, (3) Complete idempotency (multiple calls), (4) Cancel on already-completed task. These are critical concurrent interaction paths that are hard to catch without dedicated tests.

src/impl/MilvusClientV2Impl.cpp
yhmo: not a problem

test/ut/types/TestOptimizeTask.cpp
yhmo: not a problem

PR #480, opened 2026-05-07
Manually trigger release workflow

yhmo:


src/CMakeLists.txt
```diff
@@ -57,5 +57,6 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/
 # install libs
 install(TARGETS milvus_sdk
```
liliu-z, line 68: The main CI (`main.yaml`) builds Windows without setting `BUILD_SHARED_LIBS`, so `add_library(milvus_sdk ...)` at `src/CMakeLists.txt:24` defaults to static. The release workflow explicitly sets `-DBUILD_SHARED_LIBS=ON`. This means the new `RUNTIME DESTINATION` install rule at line 58-60 is never exercised in PR CI — it only takes effect for shared libraries on Windows. The daily CI has zero coverage of this PR's only CMake install change.

.github/workflows/main.yaml
```diff
@@ -6,6 +6,8 @@ on:
       - 'cmake/**'
       - 'examples/**'
       - 'scripts/**'
+      - '!scripts/verify_package_linux.sh'
```
liliu-z: Lines 9 and 43 of `main.yaml` explicitly exclude `scripts/**` from PR trigger paths. `release.yaml` only supports `workflow_dispatch`. As a result, `verify_package_linux.sh` and `verify_package_windows.ps1` have no automated CI entry point before merge — syntax errors or logic bugs in these scripts would only surface during a manual release run.

src/CMakeLists.txt
```diff
@@ -57,5 +57,6 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/
...
         EXPORT milvus_sdk
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
```
liliu-z, line 70: This PR sets `BUILD_SHARED_LIBS=ON` for Windows for the first time (main.yaml:247, release.yaml:254), but `milvus_sdk` has no `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`, `__declspec(dllexport)`, or `GenerateExportHeader` anywhere in the project. MSVC will produce a DLL with no exports → empty import library → all downstream targets (testing-ut, testing-it at test/CMakeLists.txt:38-50) fail with "unresolved external symbol". The main build step itself will fail; smoke tests never run.

Fix: add `set_target_properties(milvus_sdk PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)` after the `add_library` call.

scripts/install_deps.sh
```diff
@@ -161,11 +152,13 @@ if uname | grep -wq Linux ; then
...
         if grep -q 'Ubuntu 18.04' /etc/issue ; then
-            install_deps_for_ubuntu_1804
+            install_deps_for_ubuntu
         elif grep -q 'Ubuntu 20.04' /etc/issue ; then
-            install_deps_for_ubuntu_2004
+            install_deps_for_ubuntu
         elif grep -q 'Ubuntu 22.04' /etc/issue ; then
-            install_deps_for_ubuntu_2204
+            install_deps_for_ubuntu
+        elif grep -q 'Ubuntu 24.04' /etc/issue ; then
```
liliu-z, line 160: This PR adds `ubuntu_24_04` and `ubuntu_24_04_arm` release targets (release.yaml:104-107), but `install_deps_for_ubuntu_common()` at L98 installs `python2.7`, which does not exist in Ubuntu 24.04 (noble). `apt-get install` will fail and block the entire build.

The only reference to python2 in the repo is the shebang in `scripts/run_clang_format.py`, which is actually invoked via `${Python_EXECUTABLE}` (python3) from CMakeLists.txt:164. The release build path (`build.sh -f`) skips clang-format entirely. python2.7 is completely unnecessary.

Fix: remove `python2.7` from the apt install list at L98.

scripts/verify_package_windows.ps1
```diff
@@ -0,0 +1,69 @@
...
+    }
+    std::cout << version << std::endl;
+    return version.empty() ? 1 : 0;
+}
+'@ | Set-Content -Path $smokeCpp
+
+$buildCmd = Join-Path $smokeDir 'build-smoke.cmd'
+@"
+call "%ProgramFiles%\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+cl /std:c++14 /EHsc /I"$includeDir" "$smokeCpp" "$($libFile.FullName)" /OUT:"$smokeExe"
```
liliu-z: `/OUT:` is a linker option, not a compiler option. While cl.exe may pass unrecognized options through to the linker, the `/O` prefix conflicts with MSVC optimization flags (`/O1`, `/O2`, `/Ox`). This is currently masked by issue #1 (the link will fail regardless due to missing exports), but should be fixed alongside it.

Fix: change to `/Fe"$smokeExe"` or use `/link /OUT:"$smokeExe"`.

.github/workflows/release.yaml
```diff
@@ -1,56 +1,277 @@
...
-            key: u2004
-          - distro: Fedora
-            version: 39
-            image: fedora:39
-            key: fc39
+        os: ${{ fromJSON(needs.select-targets.outputs.linux) }}
     steps:
       - name: Env
         run: echo "${HOME}/.local/bin" >> $GITHUB_PATH
       - name: Checkout
```
liliu-z, line 146: All existing tags (up to v2.6.2) predate the CPack configuration, `make package` target, `build.sh -f` flag, verify scripts, and `cpio` dependency introduced in this PR. If an operator inputs an old tag, the workflow will fail at build time. The tag input description (L7) should note a minimum supported version to prevent confusion.

liliu-z:
**release.yaml:151** The Fedora release job builds with `BUILD_SHARED_LIBS=ON` (implicit default from `build.sh`), packages via `CPACK_GENERATOR "RPM"` (`CMakeLists.txt:294`), and verifies via the `rpm2cpio | cpio -idm` branch in `verify_package_linux.sh:31-33`. None of this runs in the PR CI — Ubuntu uses DEB, macOS uses TGZ, and `main.yaml:70` only tests static builds on Linux. The RPM branch in the verify script is newly added code with zero test coverage before merge. Additionally, the release job relies on `build.sh` defaulting `BUILD_SHARED_LIBS=ON` rather than setting it explicitly. Consider explicitly setting `BUILD_SHARED_LIBS` in the Fedora release job and adding a Fedora CI step that runs `make package` + verify to cover the RPM path.

liliu-z:
**verify_package_windows.ps1:59** The Windows smoke test compiles without `/DMILVUS_SDK_SHARED` and only calls `GetSDKVersion()`. While `WINDOWS_EXPORT_ALL_SYMBOLS` makes linking succeed, the smoke test does not validate the new `dllexport`/`dllimport` data symbol path (`extern const char*` constants) introduced by this PR. This leaves the DLL consumer import path for data symbols unverified.

liliu-z:
**main.yaml:8** The `pull_request.paths` filter in `main.yaml:8-10` and the `paths-filter` at `main.yaml:42-44` only include `.github/workflows/main.yaml`. Changes to `release.yaml`, `verify_package_linux.sh`, or `verify_package_windows.ps1` will not trigger the main CI. Future modifications to the release pipeline can be merged without any automated feedback.

liliu-z:
**test/st/TestDqlUtils.cpp:332** The `ConvertFilterTemplatesTest` (L332-334) is conditionally skipped when building as a Windows shared DLL. While call-chain analysis confirms the production path is unaffected (protobuf containers are created and filled within the DLL), the insert-then-fill refactoring introduced by this PR lacks direct automated test coverage on the target platform. This is a minor coverage gap, not a correctness issue.

PR #515, opened 2026-06-09
Support snapshot interfaces

yhmo:


yhmo (commented):
Discarding empty pending review created during posting.

src/include/milvus/request/snapshot/RestoreSnapshotRequest.h
```diff
@@ -0,0 +1,95 @@
...
+    SetTargetCollectionName(const std::string& collection_name);
+
+    RestoreSnapshotRequest&
+    WithTargetCollectionName(const std::string& collection_name);
+
+ private:
+    std::string snapshot_name_;
+    std::string source_db_name_;
+    std::string source_collection_name_;
+    std::string target_db_name_;
```
yhmo, line 81: The new request exposes `RewriteData()` / `SetRewriteData()`, but the field has no documentation at all, unlike the Java proto/service side where `rewrite_data` has specific semantics. Since this is a user-facing SDK request, please add a brief comment explaining what `true` vs `false` does and what the default means; otherwise callers have to discover it from the server proto or trial-and-error.

yhmo (commented):
Discarding empty pending review created during posting retry.

test/it/v1/TestSnapshotV2.cpp
```diff
@@ -0,0 +1,306 @@
...
+        .WithSnapshotName("snap")
+        .WithDescription("desc")
+        .WithCompactionProtectionSeconds(9);
+
+    auto status = client->CreateSnapshot(request);
+    EXPECT_TRUE(status.IsOk());
+}
+
+TEST_F(UnconnectMilvusMockedTest, CreateSnapshotInvalidArgument) {
+    auto client = CreateConnectedV2Client(service_, server_.ListenPort());
```
yhmo, line 97: The mocked snapshot integration suite only exercises client-side validation failures for `CreateSnapshot` and `UnpinSnapshotData`, but all of the RPC-backed happy paths (`DropSnapshot`, `ListSnapshots`, `DescribeSnapshot`, `RestoreSnapshot`, `GetRestoreSnapshotState`, `ListRestoreSnapshotJobs`, `PinSnapshotData`) still have no explicit gRPC failure-path coverage. Since these APIs all rely on `ConnectionHandler::Invoke(...)` / `MilvusConnection` transport mapping, please add at least one representative `UNAVAILABLE -> RPC_FAILED` assertion for the snapshot response-style calls as well, otherwise a transport/mapping regression can still slip through while the new tests stay green.

PR #516, opened 2026-06-11
Add Snapshot example

yhmo:


yhmo (commented):
I found one high-confidence issue in the PR diff.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1923,6 +1936,14 @@ MilvusClientV2Impl::Flush(const FlushRequest& request) {
                 return Status::OK();
             },
             progress_monitor);
+
+        // wait more 1 second to make sure the flushed segments are visible.
+        // there might be a small delay(on server-side) between the flush_ts and the time when the segments are actually flushed,
+        // if user calls createSnapshot immediately after flush returns, it might not find the flushed segments.
+        if (status.IsOk()) {
```
yhmo, line 1943: This adds an unconditional extra 1s delay to every successful synchronous `Flush()` call, even when the caller is unrelated to snapshots. That changes the SDK-wide latency/semantics of `Flush` to work around one timing-sensitive example. A narrower fix would keep `Flush` as the normal completion boundary and handle snapshot-specific sequencing in the example flow (or in a dedicated snapshot helper) instead of silently slowing all flush users.

yhmo: as designed

yhmo (commented):
I found one additional medium-severity example issue in the PR diff.

examples/src/v2/snapshot.cpp
```diff
@@ -0,0 +1,208 @@
...
+    std::cout << "Restore job state: state=" << std::to_string(job_info.State()) << ", progress=" << job_info.Progress()
+              << ", reason=" << job_info.Reason() << std::endl;
+
+    status = client->LoadCollection(milvus::LoadCollectionRequest().WithCollectionName(restore_collection_name));
+    util::CheckStatus("load restored collection: " + std::string(restore_collection_name), status);
+
+    auto source_row_count = getPersistedRowCount(client, collection_name);
+    auto target_row_count = getPersistedRowCount(client, restore_collection_name);
+    std::cout << "Source persisted row count=" << source_row_count << ", target query row count=" << target_row_count
+              << std::endl;
```
yhmo: Middle ---- The message says `target query row count`, but both `source_row_count` and `target_row_count` come from `getPersistedRowCount()` (`GetCollectionStats()`), so this example never actually verifies the restored collection through the query path after `LoadCollection()`. Either rename the message to `target persisted row count` or use `queryRowCount()` for the target-side check if query visibility is what you want to demonstrate.

examples/src/v2/snapshot.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    status = client->LoadCollection(milvus::LoadCollectionRequest().WithCollectionName(restore_collection_name));
+    util::CheckStatus("load restored collection: " + std::string(restore_collection_name), status);
+
+    auto source_row_count = getPersistedRowCount(client, collection_name);
+    auto target_row_count = getPersistedRowCount(client, restore_collection_name);
+    std::cout << "Source persisted row count=" << source_row_count << ", target persisted row count="
+              << target_row_count << std::endl;
+    if (source_row_count != target_row_count) {
+        std::cerr << "Restored row count mismatch: source=" << source_row_count << ", target=" << target_row_count
```
yhmo, line 192: Middle ---- This example treats `GetCollectionStats().Stats().RowCount()` as the final restore-success check immediately after restore/load. In this repo, nearby examples already note that collection stats can lag behind recent changes, so this check can fail even when the restore job completed normally. A query-based validation (or an explicit note/poll about stats lag) would make the example more reliable.

examples/src/v2/snapshot.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    auto client = milvus::MilvusClientV2::Create();
+
+    milvus::ConnectParam connect_param{"http://localhost:19530"};
+    auto status = client->Connect(connect_param);
+    util::CheckStatus("connect milvus server", status);
+
+    status = client->DropCollection(milvus::DropCollectionRequest().WithCollectionName(restore_collection_name));
+    status = client->DropCollection(milvus::DropCollectionRequest().WithCollectionName(collection_name));
+
```
yhmo, line 100: Middle ---- These two startup cleanup calls ignore the returned status completely. If `DropCollection()` fails for a real reason other than "not found" (for example, auth / connectivity / wrong DB context), the example will continue and later fail in `CreateCollection()` or `RestoreSnapshot()` with a much more misleading error. It would be better to explicitly tolerate only the expected "not found" case and surface other drop failures immediately.

PR #517, opened 2026-06-12
Add ExternalCollection interfaces

yhmo:


examples/src/v2/external_table.cpp
```diff
@@ -0,0 +1,212 @@
...
+#include "milvus/MilvusClientV2.h"
+
+namespace {
+
+const char* const collection_name = "CPP_V2_EXTERNAL_TABLE";
+const char* const field_id = "product_id";
+const char* const field_vector = "embedding";
+const char* const field_text = "content";
+const uint32_t dimension = 128;
+const char* const default_external_source = "s3://minio:9000/a-bucket/external_table_example_data/data.parquet";
```
yhmo: Medium ---- The Java reference flow uses `externalSource` as the object-storage prefix (`s3://minio:9000/<bucket>/<prefix>/`), not a single parquet object. Pointing the default here at `.../data.parquet` makes the C++ example diverge from the same MinIO setup and can cause `RefreshExternalCollection` to look at the wrong location. Please switch this default to the prefix (`.../external_table_example_data/`) and keep the nearby comment aligned with that contract.

PR #518, opened 2026-06-16
Struct field supports BINARY_VECTOR/FLOAT16_VECTOR/BFLOAT16_VECTOR/INT8_VECTOR

yhmo:


src/impl/utils/DqlUtils.cpp
```diff
@@ -1357,11 +1460,31 @@ ConvertSearchResults(const proto::milvus::SearchResults& rpc_results, const std:
...
+            const auto& element_indices = result_data.element_indices().data();
+            std::vector<int64_t> element_offsets;
+            element_offsets.reserve(item_topk);
+            auto begin = element_indices.begin();
+            std::advance(begin, offset);
+            auto end = begin;
+            std::advance(end, item_topk);
+            std::copy(begin, end, std::back_inserter(element_offsets));
+            item_fields_data.emplace_back(
+                std::make_shared<Int64FieldData>(element_offset_name, std::move(element_offsets)));
```
yhmo, line 1488: Medium ---- Could we also add the synthesized `element_offset` field name to the visible output-name set here? Right now `ConvertSearchResults()` appends an `Int64FieldData` for the offsets, but `SingleResult::OutputRows()` still filters through the original server `output_names_` set. That means the new `struct_field.cpp` example prints rows via `OutputRows()` and claims each result will contain `element_offset`, but the field is actually hidden unless callers bypass `OutputRows()` and fetch it manually from `OutputField("element_offset")`.

test/st/cases/TestCollection.cpp
```diff
@@ -226,8 +226,49 @@ TEST_F(MilvusServerTestCollectionOps, AddCollectionField) {
...
+        .AddField(milvus::FieldSchema("score", milvus::DataType::FLOAT))
+        .AddField(milvus::FieldSchema("embedding", milvus::DataType::FLOAT_VECTOR).WithDimension(4))
+        .AddField(milvus::FieldSchema("binary_embedding", milvus::DataType::BINARY_VECTOR).WithDimension(32))
+        .AddField(milvus::FieldSchema("float16_embedding", milvus::DataType::FLOAT16_VECTOR).WithDimension(16))
+        .AddField(milvus::FieldSchema("bfloat16_embedding", milvus::DataType::BFLOAT16_VECTOR).WithDimension(16))
+        .AddField(milvus::FieldSchema("int8_embedding", milvus::DataType::INT8_VECTOR).WithDimension(16));
+
+    auto status = client_->AddCollectionStructField(milvus::AddCollectionStructFieldRequest()
+                                                        .WithCollectionName(collection_name)
+                                                        .WithStructField(std::move(struct_field)));
```
yhmo, line 244: High ---- `MilvusClientV2Impl::AddCollectionStructField()` now rejects any struct field whose `IsNullable()` is false, but this test never sets `.WithNullable(true)` on `struct_field`. As written, the client returns `INVALID_ARGUMENT` before sending the RPC, so `ExpectStatusOK(status)` will fail consistently. Please mark the new struct field nullable here (as in the example) or relax the validation if non-nullable additions are intended.

src/impl/utils/DqlUtils.cpp
```diff
@@ -1362,6 +1465,28 @@ ConvertSearchResults(const proto::milvus::SearchResults& rpc_results, const std:
         item_fields_data.emplace_back(std::move(id_field));
         item_fields_data.emplace_back(std::move(score_field));
 
+        // set element offset field for vector search with payload, the element offset is the position
+        // of the returned vector in the original field data of the entity, which is useful for users to
+        // retrieve the corresponding payload data. For non-vector search or vector search without payload,
+        // the server will not return element offset, client will not set this field.
+        if (result_data.has_element_indices()) {
+            std::string element_offset_name = ELEMENT_OFFSET;
+            while (field_names.find(element_offset_name) != field_names.end() || element_offset_name == score_name) {
```
liliu-z, line 1474: The loop `while (field_names.find(element_offset_name) != field_names.end() || element_offset_name == score_name)` guards against requested-output and score-name collisions but not the real primary-key name, so when a collection's PK is named `element_offset` (and isn't a requested output) the synthetic offset field is given the PK's name; `GenGetters` builds a `std::map` whose `insert` keeps the first key and the PK is inserted before the offset field, so `OutputField("element_offset")`/`OutputRows()` resolve the PK and the real element offsets are silently dropped. Add `|| element_offset_name == real_pk_name` to the loop.

test/st/cases/TestGeneric.cpp
```diff
@@ -27,7 +27,7 @@ TEST_F(MilvusServerTestGeneric, GetServerVersion) {
     auto status = client_->GetServerVersion(version);
     std::cout << "Milvus version: " << version << std::endl;
     milvus::test::ExpectStatusOK(status);
-    EXPECT_THAT(version, testing::MatchesRegex("v?2.+"));
+    EXPECT_FALSE(version.empty());
```
liliu-z, line 30: The version check was relaxed from `EXPECT_THAT(version, MatchesRegex("v?2.+"))` to `EXPECT_FALSE(version.empty())` as collateral of switching the ST suite to the master server image, which drops all version-format validation so any non-empty string (including a malformed or non-2.x version) now passes. Restore a format-aware assertion so version-format coverage isn't silently lost.

PR #524, opened 2026-06-22
Support Highlighter for Search

yhmo:


src/include/milvus/types/Highlighter.h
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+ private:
+    void
+    SyncHighlightQueries() {
+        nlohmann::json queries = nlohmann::json::array();
+        for (const auto& query : highlight_queries_) {
+            queries.push_back({{"type", query.type}, {"field", query.field}, {"text", query.text}});
+        }
+        SetParam("highlight_queries", queries.dump());
```
yhmo: src/include/milvus/types/Highlighter.h line:167
High ---- The server reads lexical query config from the singular key "highlight_query", but SyncHighlightQueries() serializes it as "highlight_queries". That means AddHighlightQuery()/WithHighlightQueries() data is dropped on the wire, so lexical text-match highlighting never runs even though the new API appears configured.

src/include/milvus/types/Highlighter.h
```diff
@@ -0,0 +1,300 @@
...
+
+    LexicalHighlighter&
+    AddHighlightQuery(HighlightQuery query) {
+        highlight_queries_.emplace_back(std::move(query));
+        SyncHighlightQueries();
+        return *this;
+    }
+
+    LexicalHighlighter&
+    AddHighlightQuery(std::string type, std::string field, std::string text) {
```
yhmo: src/include/milvus/types/Highlighter.h line:102
Medium ---- AddHighlightQuery() accepts an arbitrary type string, but the server only accepts enum names like "TextMatch" here. This PR's own tests pass "term", so callers following that example will get INVALID_ARGUMENT at runtime unless the client validates or maps the accepted values.

test/ut/utils/TestDqlUtils.cpp
```diff
@@ -372,10 +386,65 @@ TEST_F(DqlUtilsTest, ConvertSearchRequestV2) {
     req.WithFilter("id > 0");
     req.AddOutputField("name");
 
+    auto highlighter = std::make_shared<milvus::LexicalHighlighter>();
+    highlighter->AddHighlightQuery("term", "text", "milvus")
```
liliu-z, line 390: The only demonstrated highlight-query type in the tests is `"term"`, which a real server rejects: the `HighlightQueryType` enum defines only `TextMatch`, and the proxy validates each query's `type` against it. The SDK API itself just forwards whatever type string the caller passes, so production code is unaffected, but this example/test value is invalid and would fail end-to-end — use `TextMatch`.

src/include/milvus/types/Highlighter.h
```diff
@@ -0,0 +1,300 @@
...
+ protected:
+    static std::string
+    BoolToString(bool value) {
+        return value ? "true" : "false";
+    }
+
+    template <typename T>
+    static std::string
+    NumberToString(T value) {
+        std::ostringstream oss;
```
liliu-z: NumberToString stringifies numeric params with `std::ostringstream` under the process-global locale, so a thousands-grouping locale turns e.g. `WithFragmentSize(16000)` into `16,000`/`16.000`. The server parses `fragment_size`/`fragment_offset`/`num_of_fragments` with `strconv.ParseInt` and `max_client_batch_size` with `strconv.Atoi`, hard-failing the whole search, and `threshold` gets a wrong decimal separator. Serialize locale-independently (e.g. `oss.imbue(std::locale::classic())`, or `std::to_string` for integers).

src/impl/utils/DqlUtils.cpp
```diff
@@ -1298,6 +1332,11 @@ ConvertSearchRequest(const T& request, const std::string& current_db, proto::mil
     // extra params offset/round_decimal/group_by/radius/range_filter/nprobe etc.
     SetExtraParams(request.ExtraParams(), rpc_request.mutable_search_params());
 
+    auto highlighter_status = SetSearchHighlighter(request, rpc_request);
```
liliu-z, line 1335: This change implements only the request side. SetSearchHighlighter is wired in here, but nothing parses the response's `SearchResultData.highlight_results` and the SearchResults type exposes no accessor for it. As shipped the feature returns no highlight data to the caller even after the request key is fixed, so add response parsing plus a result-type accessor unless that is a committed follow-up.

test/ut/utils/TestDqlUtils.cpp
```diff
@@ -372,10 +386,65 @@ TEST_F(DqlUtilsTest, ConvertSearchRequestV2) {
...
+        .WithNumOfFragments(2);
+    req.WithHighlighter(highlighter);
+
     milvus::proto::milvus::SearchRequest rpc_request;
     auto status = milvus::ConvertSearchRequest(req, "default", rpc_request);
     EXPECT_TRUE(status.IsOk());
     EXPECT_EQ(rpc_request.collection_name(), "test_coll");
+    EXPECT_TRUE(rpc_request.has_highlighter());
+    EXPECT_EQ(rpc_request.highlighter().type(), milvus::proto::common::HighlightType::Lexical);
+    EXPECT_EQ(rpc_request.highlighter().params_size(), 7);
```
liliu-z: The highlighter tests assert only `params_size()` and the type enum, never the actual param keys or values. That is exactly why the defects above ship undetected: the wrong `highlight_queries` key and the invalid `term` query type still yield the expected param count. Assert the serialized key set and the `highlight_query` JSON payload so these regressions are caught.

src/impl/types/Highlighter.cpp
```diff
@@ -0,0 +1,238 @@
...
+LexicalHighlighter::WithNumOfFragments(int64_t value) {
+    SetParam("num_of_fragments", std::to_string(value));
+    return *this;
+}
+
+void
+LexicalHighlighter::SyncHighlightQueries() {
+    nlohmann::json queries = nlohmann::json::array();
+    for (const auto& query : highlight_queries_) {
+        queries.push_back({{"type", query.type}, {"field", query.field}, {"text", query.text}});
```
liliu-z, line 117: LexicalHighlighter::SyncHighlightQueries serializes each query's `type` string verbatim with no validation, but the 2.6 server resolves it via an exact proto-enum-name lookup whose only entry is `TextMatch`, so any other string is rejected at runtime with "invalid highlight query type". The SDK's own unit tests demonstrate `"term"`/`"phrase"`, so a user copying the tested pattern builds a request the server refuses. Expose a typed constant/enum (or validate/map the string) for the accepted value, and fix the tests to use `TextMatch`.

src/include/milvus/types/SearchResults.h
```diff
@@ -118,6 +126,12 @@ struct MILVUS_SDK_API SingleResult {
     Status
     OutputRow(int i, EntityRow& row) const;
 
+    /**
+     * @brief Get highlight results of one row. Throw exception if the i is out of bound.
```
liliu-z: The doc comment says the accessor throws on an out-of-bound index, but the implementation returns a `Status` ({StatusCode::INVALID_ARGUMENT, "The row index is out of bound"}) and never throws; reword it to document the Status contract. The sibling `OutputRow` at line 124 carries the identical misleading wording (and also returns Status), so this was likely copied — fixing both keeps the file consistent.

src/impl/types/Highlighter.cpp
```diff
@@ -0,0 +1,238 @@
...
+SemanticHighlighter&
+SemanticHighlighter::AddPostTag(std::string tag) {
+    post_tags_.emplace_back(std::move(tag));
+    SyncPostTags();
+    return *this;
+}
+
+SemanticHighlighter&
+SemanticHighlighter::WithThreshold(float value) {
+    SetParam("threshold", std::to_string(value));
```
liliu-z, line 196: At line 196, `WithThreshold` formats the threshold without the SDK's `DoubleToString` helper that the rest of the codebase uses for doubles, so the serialized number can differ in format/precision and diverges from the established convention. Route the value through `DoubleToString` for consistent serialization.

src/impl/utils/DqlUtils.cpp
```diff
@@ -1357,6 +1396,29 @@ ConvertSearchResults(const proto::milvus::SearchResults& rpc_results, const std:
             score_name = "_" + score_name;
         }
 
+        std::vector<HighlightResults> item_highlight_results;
+        item_highlight_results.resize(item_topk);
+        for (const auto& rpc_highlight_result : result_data.highlight_results()) {
+            for (int j = 0; j < item_topk; ++j) {
+                const auto data_index = offset + j;
```
liliu-z, line 1403: The multi-query highlight indexing (`data_index = offset + j`, where `offset += item_topk` accumulates per query) is exercised only with a single query: every ConvertSearchResults test sets num_queries=1, so `offset` never advances past 0. Add a coverage case with nq>1 so a regression in mapping per-query highlight datas back to rows is caught.

src/include/milvus/types/Highlighter.h
```diff
@@ -0,0 +1,167 @@
...
+    HighlightType() const override;
+
+    LexicalHighlighter&
+    WithHighlightQueries(const std::vector<HighlightQuery>& queries);
+
+    LexicalHighlighter&
+    AddHighlightQuery(HighlightQuery query);
+
+    LexicalHighlighter&
+    AddHighlightQuery(std::string type, std::string field, std::string text);
```
yhmo, line 66: src/include/milvus/types/Highlighter.h line:66
Medium ---- This API still accepts an arbitrary lexical query `type` string, and `SyncHighlightQueries()` later sends it to the server unchanged. The 2.6 server validates this against enum names like `TextMatch`, but this PR's own tests now use values like `term` and `phrase`, so callers following the new API pattern can build a request that fails with `INVALID_ARGUMENT` at runtime unless the SDK validates or maps the accepted values here.

test/ut/types/TestHighlighter.cpp
```diff
@@ -0,0 +1,124 @@
...
+TEST_F(LexicalHighlighterTest, DefaultState) {
+    milvus::LexicalHighlighter highlighter;
+    EXPECT_EQ(highlighter.HighlightType(), "Lexical");
+    EXPECT_TRUE(highlighter.Params().empty());
+}
+
+TEST_F(LexicalHighlighterTest, SerializeParams) {
+    milvus::LexicalHighlighter highlighter;
+
+    auto& ref = highlighter.WithHighlightQueries({{"term", "title", "milvus"}, {"phrase", "body", "vector db"}})
```
yhmo, line 34: test/ut/types/TestHighlighter.cpp line:34
Medium ---- These lexical query `type` values are not server-valid. The request path forwards them unchanged as `highlight_query[*].type`, but Milvus 2.6 only accepts enum names like `TextMatch`, so a user following this tested pattern can build a search that fails with `INVALID_ARGUMENT` at runtime. Please switch the tests/examples to `TextMatch`, or validate/map the accepted values in the SDK.

yhmo: no need to fix

PR #526, opened 2026-06-23
Add DumpMessages/AlterRole/UpdateUser interfaces

yhmo:


src/impl/MilvusClientV2Impl.cpp
```diff
@@ -2507,6 +2507,28 @@ MilvusClientV2Impl::GetReplicateInfo(const GetReplicateInfoRequest& request, Get
...
+    rpc_request.set_start_timetick(request.StartTimeTick());
+    rpc_request.set_end_timetick(request.EndTimeTick());
+
+    auto callback = [&on_message](const proto::common::ImmutableMessage& rpc_message) {
+        DumpedMessage message;
+        ConvertImmutableMessage(rpc_message, message);
+        return on_message(message);
+    };
+
+    return connection_.GetConnection()->DumpMessages(rpc_request, GrpcOpts{connection_.GetRpcDeadlineMs()}, callback);
```
liliu-z: `DumpMessages` calls `GetConnection()->DumpMessages(...)` directly instead of routing through `Invoke`/`apiHandler` as every other method does, so it dereferences the connection `shared_ptr` while it is still null before `Connect()` has succeeded. This bypasses the `Invoke` guard that returns `NOT_CONNECTED`, breaking the not-connected invariant the rest of the client upholds, so calling `DumpMessages` before `Connect()` crashes the process on a null-pointer deref instead of returning a status. Route the call through `Invoke`/`apiHandler`.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -2507,6 +2507,28 @@ MilvusClientV2Impl::GetReplicateInfo(const GetReplicateInfoRequest& request, Get
...
 }
 
+Status
+MilvusClientV2Impl::DumpMessages(const DumpMessagesRequest& request,
+                                 const std::function<Status(const DumpedMessage&)>& on_message) {
+    proto::milvus::DumpMessagesRequest rpc_request;
+    rpc_request.set_pchannel(request.PChannel());
+    rpc_request.mutable_start_message_id()->set_id(request.StartMessageID().ID());
+    proto::common::WALName wal_name;
+    if (proto::common::WALName_Parse(request.StartMessageID().WalName(), &wal_name)) {
```
liliu-z: A mistyped WAL name (e.g. "Rocksmq" instead of the enum spelling "RocksMQ") makes WALName_Parse return false, so the if-guard skips set_wal_name and the request is sent with the default Unknown WAL and no error surfaced. Consider returning INVALID_ARGUMENT on parse failure, consistent with the SDK's other input validation, instead of silently degrading to Unknown. (In normal use the WalName round-trips from WALName_Name, so this only bites a hand-built/mistyped name.)

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -2507,6 +2507,28 @@ MilvusClientV2Impl::GetReplicateInfo(const GetReplicateInfoRequest& request, Get
...
+        rpc_request.mutable_start_message_id()->set_wal_name(wal_name);
+    }
+    rpc_request.set_start_timetick(request.StartTimeTick());
+    rpc_request.set_end_timetick(request.EndTimeTick());
+
+    auto callback = [&on_message](const proto::common::ImmutableMessage& rpc_message) {
+        DumpedMessage message;
+        ConvertImmutableMessage(rpc_message, message);
+        return on_message(message);
```
liliu-z, line 2545: The `on_message` callback is invoked without first checking that it is set, so an empty/`nullptr` `std::function` is accepted at registration. Invoking an empty `std::function` throws `std::bad_function_call`, so a caller that passes a null callback hits an uncaught exception on the first streamed message rather than getting an upfront validation error. Validate `on_message` is non-empty and return `InvalidArgument` before starting the stream.

src/impl/MilvusClientV2Impl.cpp
liliu-z: DumpMessages forwards the global per-RPC deadline (GrpcOpts{GetRpcDeadlineMs()}) to a potentially long-running stream, and MilvusConnection::DumpMessages applies it as the gRPC context deadline. For an open-ended dump (the example sets end_timetick to uint64_t::max(), not 0) this deadline terminates the stream mid-dump with DEADLINE_EXCEEDED/TIMEOUT, silently truncating output. It only triggers when the caller has set SetRpcDeadlineMs() (default rpc_deadline_ms_ is 0 = no deadline). Consider not applying the unary RPC deadline to streaming dumps, or overriding it for open-ended ranges.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -2740,13 +2788,28 @@ MilvusClientV2Impl::DescribeRole(const DescribeRoleRequest& request, DescribeRol
...
         for (const auto& entity : rpc_response.entities()) {
             desc.AddGrantItem({entity.object().name(), entity.object_name(), entity.db_name(), entity.role().name(),
                                entity.grantor().user().name(), entity.grantor().privilege().name()});
         }
+
+        proto::milvus::SelectRoleRequest role_request;
+        role_request.set_include_user_info(false);
+        role_request.mutable_role()->set_name(request.RoleName());
+        proto::milvus::SelectRoleResponse role_response;
+        auto status = connection_.GetConnection()->SelectRole(role_request, role_response,
```
liliu-z: Within `DescribeRole`, the embedded `SelectRole` call (`SelectRole(role_request, role_response, ...)`) executes as a post-step after `apiHandler`'s `Retry()` has already returned, so only the primary RPC is retried while `SelectRole` is one-shot. A transient failure on `SelectRole` therefore fails the whole `DescribeRole` with no retry, and this extra round-trip is issued unconditionally on every call. Move `SelectRole` inside the retried callback or wrap it in `Retry`.

examples/src/v2/cdc.cpp
```diff
@@ -0,0 +1,121 @@
...
+    const std::string target_pchannel = GeneratePChannels(kClusterBId).front();
+
+    milvus::GetReplicateInfoRequest replicate_info_request;
+    replicate_info_request.WithSourceClusterID(kClusterAId).WithTargetPChannel(target_pchannel);
+
+    milvus::GetReplicateInfoResponse replicate_info_response;
+    status = cluster_b_client->GetReplicateInfo(replicate_info_request, replicate_info_response);
+    util::CheckStatus("get replicate info from cluster B", status);
+
+    milvus::ReplicateMessageID start_message_id;
```
yhmo, line 150: examples/src/v2/cdc.cpp line:101
High ---- The proto commit pinned in this PR adds `salvage_checkpoint` specifically for data-salvage dumps, but the SDK response type still only exposes `Checkpoint()`, so this example can only resume from the regular replicated checkpoint. After a force-promote that misses the unsynchronized tail `DumpMessages()` is supposed to recover.

examples/src/v2/cdc.cpp
```diff
@@ -0,0 +1,121 @@
...
+    start_message_id.WithID(replicate_info_response.Checkpoint().MessageID().ID())
+        .WithWalName(replicate_info_response.Checkpoint().MessageID().WalName());
+
+    milvus::DumpMessagesRequest dump_request;
+    dump_request.WithPChannel(target_pchannel)
```
yhmo: examples/src/v2/cdc.cpp line:106
High ---- DumpMessages here runs against cluster A, but `target_pchannel` was built from `kClusterBId` and `GetReplicateInfo` documents the returned checkpoint pchannel as the source-cluster channel. This points the salvage dump at the downstream channel name, so on a force-promote recovery the example either reads the wrong stream or no stream at all.

PR #527, opened 2026-06-24
Add DumpMessages/AlterRole/UpdateUser interfaces

yhmo:


examples/src/v2/cdc.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    milvus::GetReplicateInfoResponse replicate_info_response;
+    status = cluster_b_client->GetReplicateInfo(replicate_info_request, replicate_info_response);
+    CheckStatus("get replicate info from cluster B", status);
+    PrintReplicateInfo(replicate_info_response);
+
+    const std::string dump_start_message_id = replicate_info_response.SalvageCheckpoint().MessageID().ID();
+
+    milvus::ReplicateMessageID start_message_id;
+    start_message_id.WithID(dump_start_message_id).WithWalName("RocksMQ");
```
yhmo, line 151: examples/src/v2/cdc.cpp line:151
Medium ---- `DumpMessages()` validates the WAL name string, but this example hard-codes `"RocksMQ"` instead of reusing `replicate_info_response.SalvageCheckpoint().MessageID().WalName()`. If the checkpoint came from a Pulsar/Kafka deployment, callers copying this sample will send the wrong WAL name and hit `Unknown WAL name`/mismatched replay behavior even though the server already returned the correct value.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3205,13 +3272,32 @@ MilvusClientV2Impl::DescribeRole(const DescribeRoleRequest& request, DescribeRol
...
+        proto::milvus::SelectRoleRequest role_request;
+        role_request.set_include_user_info(false);
+        role_request.mutable_role()->set_name(request.RoleName());
+        proto::milvus::SelectRoleResponse role_response;
+
+        auto connection = connection_.GetConnection();
+        if (connection == nullptr) {
+            return Status{StatusCode::NOT_CONNECTED, "Connection is not created!"};
+        }
+        auto status = connection->SelectRole(role_request, role_response, GrpcOpts{connection_.GetRpcDeadlineMs()});
```
liliu-z, line 3292: DescribeRole now issues a second SelectRole RPC inside the post-callback purely to fetch the role description. Unlike the primary SelectGrant call (wrapped in Retry), this supplementary call is a raw connection->SelectRole with no retry, and any failure returns early so the whole DescribeRole fails and the grant items already collected are discarded. Previously this path succeeded with a single SelectGrant RPC. Consider tolerating a SelectRole failure (leave the description empty) so grants are still returned, or routing it through the retry path.

PR #529, opened 2026-06-25
Align CreateRoleRequest/CreateUserRequest/SearchResponse/QueryResponse with other SDKs

yhmo:


src/include/milvus/response/dql/SearchResponse.h
```diff
@@ -43,8 +45,43 @@ class MILVUS_SDK_API SearchResponse {
...
+    SetCost(int64_t cost);
+
+    int64_t
+    ScannedRemoteBytes() const;
+
+    void
+    SetScannedRemoteBytes(int64_t scanned_remote_bytes);
+
+    int64_t
+    ScannedTotalBytes() const;
```
yhmo, line 67: src/include/milvus/response/dql/SearchResponse.h line:67
Medium ---- This PR adds `Cost()/ScannedRemoteBytes()/ScannedTotalBytes()/CacheHitRatio()` to the public `SearchResponse` API, but the only new RPC wiring in `MilvusClientV2Impl::Search/HybridSearch` is `session_ts()`. As written, these four getters will always return the sentinel defaults even when the server has real top-level search metadata, so the feature is only half implemented and callers cannot tell whether the values are truly unavailable or just never propagated. Please either wire the corresponding response fields end-to-end here or defer exposing these accessors until the SDK can actually populate them.

src/impl/utils/DqlUtils.cpp
```diff
@@ -61,6 +63,43 @@ SetSearchHighlighter(const T&, proto::milvus::SearchRequest&) {
...
+        try {
+            response.SetScannedTotalBytes(std::stoll(total_bytes_it->second));
+        } catch (...) {
+        }
+    }
+
+    const auto cache_hit_ratio_it = extra_info.find("cache_hit_ratio");
+    if (cache_hit_ratio_it != extra_info.end()) {
+        try {
+            response.SetCacheHitRatio(std::stof(cache_hit_ratio_it->second));
```
yhmo: src/impl/utils/DqlUtils.cpp line:97
Medium ---- `cache_hit_ratio` is serialized on the wire as dot-decimal text, but this code parses it with locale-sensitive `std::stof`. If a client process is running under a locale that expects commas, a valid server value like `"0.5"` will throw here and leave `CacheHitRatio()` at the sentinel `-1.0f`, so the new API becomes unreliable for otherwise valid responses. Please parse this protocol value with a locale-independent routine such as `std::from_chars` or an explicit classic-locale parser.

src/impl/utils/DqlUtils.cpp
```diff
@@ -61,6 +64,43 @@ SetSearchHighlighter(const T&, proto::milvus::SearchRequest&) {
 
 }  // namespace
 
+void
+FillSearchResponseExtraInfo(const proto::common::Status& status, SearchResponse& response) {
```
yhmo, line 68: src/impl/utils/DqlUtils.cpp line:68
Medium ---- `FillSearchResponseExtraInfo()` only overwrites a field when that key is present in `status.extra_info()`. Because `SearchResponse` now stores these values as mutable object state with sentinel defaults, reusing the same response object across two searches can leak stale metadata from the first call into the second one whenever the later response omits `report_value`, `scanned_*`, or `cache_hit_ratio`. Please reset all four metadata fields to their sentinel defaults before reading `extra_info`, so each RPC reflects only the current server response.

src/impl/utils/DqlUtils.cpp
```diff
@@ -61,6 +64,48 @@ SetSearchHighlighter(const T&, proto::milvus::SearchRequest&) {
...
+    response.SetScannedRemoteBytes(-1);
+    response.SetScannedTotalBytes(-1);
+    response.SetCacheHitRatio(-1.0f);
+
+    const auto& extra_info = status.extra_info();
+
+    const auto cost_it = extra_info.find("report_value");
+    if (cost_it != extra_info.end()) {
+        try {
+            response.SetCost(std::stoll(cost_it->second));
```
liliu-z, line 79: std::stoll at DqlUtils.cpp:79 (and the parallel sites :87, :95) stops at the first non-numeric character instead of rejecting the value, so a malformed field such as "123abc" is parsed as 123 rather than falling back to the -1 sentinel. The cache_hit_ratio field just below uses ParseFloatWithLocale, whose stream.eof() check rejects any unconsumed trailing input, so integer and float parsing are inconsistent. This is not reachable with the current server (which emits clean integers), so it is a low-severity robustness nit: prefer a strict integer parse that falls back to -1 on trailing characters, mirroring the float path.

PR #531, opened 2026-06-29
Support AddFunctionField/DropCollectionField/DropFunctionField interfaces

yhmo:


test/ut/request/TestCollectionRequests.cpp
```diff
@@ -327,6 +327,84 @@ TEST_F(AddCollectionStructFieldRequestTest, GettersAndSetters) {
...
+
+TEST_F(DropCollectionFieldRequestTest, GettersAndSetters) {
+    milvus::DropCollectionFieldRequest req;
+    req.WithCollectionName("drop_field_coll").WithDatabaseName("my_db");
+    EXPECT_EQ(req.CollectionName(), "drop_field_coll");
+    EXPECT_EQ(req.DatabaseName(), "my_db");
+    EXPECT_EQ(req.FieldName(), "");
+    EXPECT_EQ(req.FieldID(), 0);
+
+    auto& ref = req.WithFieldName("my_field").WithFieldID(101);
```
yhmo: The new client implementation rejects any `DropCollectionFieldRequest` that has both `FieldName()` and `FieldID()` set, but this new UT codifies exactly that impossible state and the new IT file never covers the `INVALID_ARGUMENT` branch. As written, the tests document the wrong public contract for this API and would not catch a regression in the XOR validation you just added. Could you replace this with a test that asserts one-selector-only succeeds and add an invalid-case test (both set / neither set) against `MilvusClientV2::DropCollectionField`?

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -37,6 +37,19 @@
 #include "utils/TypeUtils.h"
 
 namespace milvus {
+namespace {
+
+Status
+StatusFromAlterCollectionSchemaResponse(const proto::milvus::AlterCollectionSchemaResponse& response) {
```
liliu-z: This helper is not dead code — it is called by the post lambdas in DropCollectionField, AddFunctionField, and DropFunctionField. The real issue is redundancy: grpcCall already converts an AlterCollectionSchemaResponse through the new StatusByProtoResponse overload, so a server error becomes SERVER_FAILED before apiHandler ever reaches the post step, and on success the post lambda only re-derives an already-OK status. The helper plus all three post lambdas duplicate StatusByProtoResponse and can be removed, letting these handlers rely on the connection-layer conversion like every other RPC.

test/it/v1/TestAlterCollectionSchema.cpp
```diff
@@ -0,0 +1,232 @@
...
+            });
+
+    auto status = client->AddFunctionField(milvus::AddFunctionFieldRequest()
+                                               .WithCollectionName("coll")
+                                               .WithField(std::move(field))
+                                               .WithFunction(function));
+    EXPECT_EQ(status.Code(), StatusCode::SERVER_FAILED);
+}
+
+TEST_F(UnconnectMilvusMockedTest, AddFunctionField) {
```
yhmo, line 182: test/it/v1/TestAlterCollectionSchema.cpp line:158
Low ---- This new block only exercises the happy path plus transport/server-failure cases for AddFunctionField. The implementation also adds several caller-visible INVALID_ARGUMENT branches (null function, empty field name, output-field count != 1, mismatched output name), but none of them are covered here, so a regression in the new validation contract would slip through without a failing test. Could you add at least one invalid-input case alongside this block?

src/include/milvus/MilvusClientV2.h
```diff
@@ -491,6 +512,15 @@ class MILVUS_SDK_API MilvusClientV2 {
     virtual Status
     DropCollectionFunction(const DropCollectionFunctionRequest& request) = 0;
 
+    /**
+     * @brief Drop a function-backed output field from an existing collection.
```
yhmo: src/include/milvus/MilvusClientV2.h line:516
Low ---- The request type and implementation drop the function itself together with its output fields (`function_name` + `drop_function_output_fields=true`), but this docstring reads like the API only removes the output field. Since this is a public SDK surface, could you spell out that DropFunctionField deletes the function and its generated fields so callers understand the destructive scope?

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -705,6 +705,35 @@ MilvusClientV2Impl::AddCollectionField(const AddCollectionFieldRequest& request)
         pre, &MilvusConnection::AddCollectionField);
 }
 
+Status
+MilvusClientV2Impl::DropCollectionField(const DropCollectionFieldRequest& request) {
+    auto validate = [&request]() {
+        const bool has_field_name = !request.FieldName().empty();
+        const bool has_field_id = request.FieldID() != 0;
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:712
Medium ---- `FieldID()` is treated as valid whenever it is non-zero, so a negative ID takes the RPC path instead of being rejected locally. The sibling Java SDK only accepts `fieldId > 0`, and the server-side drop-field contract treats field IDs as positive identifiers, so this public C++ entrypoint currently accepts malformed-but-type-correct input and turns it into a remote error path. Could we reject `FieldID() <= 0` here and add a regression test for a negative ID?

PR #532, opened 2026-06-30
Move nlohmann headers to avoid conflicts with pre-installed nlohmann

yhmo:


src/include/milvus/types/FieldSchema.h
```diff
@@ -18,7 +18,7 @@
 
 #include <cstdint>
 #include <map>
-#include <nlohmann/json.hpp>
+#include <milvus/thirdparty/nlohmann/json.hpp>
```
yhmo, line 21: src/include/milvus/types/FieldSchema.h line:21
Medium ---- Moving the public headers off <nlohmann/json.hpp> is a source-compatibility break for downstream code that relied on the SDK-installed nlohmann headers. Before this change, consumers could include <nlohmann/json.hpp> from the SDK package; after this rename that path disappears entirely, so unchanged callers stop compiling. Please keep a forwarding header at the old install path or call this out as a breaking change.

test/ut/types/TestHighlighter.cpp
```diff
@@ -16,7 +16,7 @@
 
 #include <gtest/gtest.h>
 
-#include <nlohmann/json.hpp>
+#include <milvus/thirdparty/nlohmann/json.hpp>
```
yhmo, line 19: test/ut/types/TestHighlighter.cpp line:19
Medium ---- These updates only exercise the renamed header from the in-repo include tree. The risky path here is installed or Conan consumption, where <nlohmann/json.hpp> used to exist and now disappears, so please add a consumer-style build or install check that compiles against the packaged headers rather than only the source tree.

src/include/milvus/types/FieldSchema.h
yhmo: will add this to release notes

liliu-z:
**src/include/milvus/thirdparty/nlohmann/json.hpp:18** The relocated vendored header still uses the upstream include guard `INCLUDE_NLOHMANN_JSON_HPP_` and the unscoped `nlohmann` namespace, while 14 public headers expose `nlohmann::json` across the SDK's API (e.g. FieldSchema.h:305). Moving the install path removes the file-level collision with a pre-installed nlohmann (the PR's stated goal), but it does not isolate symbols: if a downstream translation unit includes both a Milvus header and its own `<nlohmann/json.hpp>` of a different version, the shared guard makes whichever is included first win and silently skips the other, producing confusing compile/link errors at the SDK's `nlohmann::json` API boundary. Full isolation needs a Milvus-specific namespace plus a renamed include guard. This is an inherited limitation, not introduced here — track as a non-blocking follow-up.

PR #536, opened 2026-07-13
Add MilvusClientV2Session

yhmo:


src/impl/utils/DqlUtils.cpp
```diff
@@ -2003,7 +2028,7 @@ AppendSearchResult(const SingleResult& from, SingleResult& to) {
 
 Status
 IsAmbiguousParam(const std::string& key) {
-    static std::set<std::string> s_ambiguous = {PARAMS, TOPK, ANNS_FIELD, METRIC_TYPE};
+    static std::set<std::string> s_ambiguous = {PARAMS, TOPK, ANNS_FIELD, METRIC_TYPE, CLUSTER_ID};
```
yhmo, line 2031: src/impl/utils/DqlUtils.cpp line:2031
Medium ---- Reserving `cluster_id` here and unconditionally filtering it in the three converters breaks the existing raw-extra-param path for ordinary client calls: `SearchRequest::AddExtraParam("cluster_id", ...)` now silently ignores the value, while Query/Hybrid values are removed and no replacement is added when the converter session ID is empty. The same caller code can therefore hit the default route without an error. Please preserve the request value for non-session calls, and only override or reject it when a non-empty session cluster ID is supplied; add a regression test for the plain-client path.

PR #537, opened 2026-07-13
Support aggregation for search()

yhmo:


src/impl/utils/DqlUtils.cpp
```diff
@@ -1630,6 +1684,150 @@ ConvertSearchResults(const proto::milvus::SearchResults& rpc_results, const std:
...
+    return bucket;
+}
+
+}  // namespace
+
+Status
+ConvertAggregationBuckets(const proto::milvus::SearchResults& rpc_results, AggregationBuckets& buckets) {
+    buckets.clear();
+    const auto& result_data = rpc_results.results();
+    if (result_data.agg_buckets_size() == 0) {
```
yhmo: src/impl/utils/DqlUtils.cpp line:1797
High ---- A valid aggregation search can return agg_topks with one zero per query and no agg_buckets (the server does this for no matches). This early return drops those per-query empty groups, so GetAggregationBuckets() has size 0 instead of num_queries and callers lose query alignment. Please treat only both fields being absent as “no aggregation response” and materialize an empty vector for each zero agg_topks entry; add a zero-bucket multi-query test.

src/impl/request/dql/SearchRequest.cpp
```diff
@@ -219,4 +219,35 @@ SearchRequest::WithHighlighter(const HighlighterPtr& highlighter) {
...
+Status
+SearchRequest::Validate() const {
+    auto status = SearchRequestBase::Validate();
+    if (!status.IsOk()) {
+        return status;
+    }
+    if (search_aggregation_ == nullptr) {
+        return Status::OK();
+    }
+    if (!GroupByField().empty()) {
```
yhmo, line 247: src/impl/request/dql/SearchRequest.cpp line:247
Medium ---- This only rejects group_by_field, but Milvus also defines nonzero offset, group_by_fields, group_size/strict_group_size, and a highlighter as incompatible with search_aggregation. These values are expressible through SearchRequest, so Validate() currently succeeds and performs an RPC that the server must reject. Please cover the typed/raw extra-param forms and GetHighlighter() here, with focused validation tests.

src/impl/utils/DqlUtils.cpp
```diff
@@ -1630,6 +1684,150 @@ ConvertSearchResults(const proto::milvus::SearchResults& rpc_results, const std:
...
+        return Status::OK();
+    }
+
+    size_t offset = 0;
+    for (auto size : result_data.agg_topks()) {
+        if (size < 0) {
+            return {StatusCode::UNKNOWN_ERROR, "Aggregation bucket count cannot be negative"};
+        }
+        auto end = std::min(offset + static_cast<size_t>(size), all_buckets.size());
+        buckets.emplace_back(all_buckets.begin() + offset, all_buckets.begin() + end);
```
yhmo: src/impl/utils/DqlUtils.cpp line:1822
Medium ---- This range constructor copies every fully converted AggregationBucket, including nested maps, JSON values, hits, and subgroups, while all_buckets remains alive. With aggregation results allowed to scale to thousands of entries, this causes a second deep copy and peak-memory spike on every response. Please move the range (for example with move iterators) or partition while converting.

src/include/milvus/types/AggregationBucket.h
```diff
@@ -0,0 +1,52 @@
...
+
+struct MILVUS_SDK_API AggregationBucket {
+    std::vector<AggregationBucketKey> key;
+    int64_t count{0};
+    std::unordered_map<std::string, nlohmann::json> metrics;
+    std::vector<AggregationHit> hits;
+    std::vector<AggregationBucket> sub_groups;
+};
+
+using AggregationBuckets = std::vector<std::vector<AggregationBucket>>;
```
yhmo, line 57: src/include/milvus/types/AggregationBucket.h line:50
Low ---- Please document that the outer vector is indexed by query order and the inner vector contains that query's buckets, including empty entries for queries with no buckets. Without that contract, this nested alias and GetAggregationBuckets() are ambiguous to public SDK callers, and the zero-result shape is especially easy to implement or consume incorrectly.

src/impl/utils/DqlUtils.cpp
```diff
@@ -1514,6 +1514,54 @@ ConvertSearchRequest(const T& request, const std::string& current_db, proto::mil
     return Status::OK();
 }
 
+void
+ConvertSearchAggregation(const SearchAggregation& from, proto::common::SearchAggregationSpec& to) {
+    for (const auto& field : from.Fields()) {
+        to.add_fields(field);
+    }
+    to.set_size(from.Size());
```
yhmo, line 1522: src/impl/utils/DqlUtils.cpp line:1522
Medium ---- SearchAggregationSpec in the pinned proto also has search_size, and the server uses it as the candidate-bucket budget before applying size. Since this model and converter never expose or serialize that field, C++ callers cannot widen the candidate set on skewed/distributed data and may get avoidably inaccurate top buckets. Please add SearchSize/WithSearchSize validation and map it with set_search_size(), with a wire-level test.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1804,7 +1807,16 @@ MilvusClientV2Impl::search(const SearchRequest& request, SearchResponse& respons
             }
         }
         auto status = ConvertSearchResults(rpc_response, pk_name, results);
+        if (!status.IsOk()) {
+            return status;
+        }
         response.SetResults(std::move(results));
+        AggregationBuckets aggregation_buckets;
+        status = ConvertAggregationBuckets(rpc_response, aggregation_buckets);
```
yhmo, line 1814: src/impl/MilvusClientV2Impl.cpp line:1815
Medium ---- response.SetResults() has already committed fresh data when ConvertAggregationBuckets() can still reject malformed agg_topks. On that error, a reused SearchResponse contains fresh Results() but stale aggregation/session/extra fields. Please convert both result sets into temporaries first and only mutate response after every conversion succeeds; add a mocked failure-path test.

src/impl/types/SearchAggregation.cpp
```diff
@@ -0,0 +1,267 @@
...
+}
+
+SearchAggregation&
+SearchAggregation::WithSubAggregation(const SearchAggregationPtr& sub_aggregation) {
+    SetSubAggregation(sub_aggregation);
+    return *this;
+}
+
+Status
+SearchAggregation::Validate() const {
```
yhmo, line 209: src/impl/types/SearchAggregation.cpp line:209
Medium ---- The patch coverage check is currently failing at 69.38% against the 80% target, and this new public Validate() logic has no focused tests for empty fields/size, invalid metrics/orders/top hits, or cycles. Please add direct validation tests so these hostile-input branches and their status codes are protected.

examples/src/v2/search_aggregation.cpp
```diff
@@ -0,0 +1,381 @@
...
+                       .AddOutputField(field_rating)
+                       .WithSearchAggregation(level1);
+    addQueryVectors(request);
+    executeSearch(client, "Case 5: three-level nested", request);
+}
+
+}  // namespace
+
+int
+main(int argc, char* argv[]) {
```
yhmo, line 363: examples/src/v2/search_aggregation.cpp line:363
Low ---- This new main is auto-built as sdk_search_aggregation_v2, but examples/README.md explicitly lists the available V2 executables and the new one is missing. Please add it there so users can discover and run the example consistently with the other recently added examples.

PR #539, opened 2026-07-14
Support CommitImport/AbortImport for BulkImport

yhmo:


src/impl/BulkImport.cpp
```diff
@@ -102,4 +102,46 @@ BulkImport::GetImportJobProgress(const std::string& url, const std::string& job_
     }
 }
 
+nlohmann::json
+BulkImport::CommitImport(const std::string& url, const std::string& job_id, const std::string& db_name,
```
yhmo, line 80: src/impl/BulkImport.cpp line:106
Medium ---- Commit and abort only apply to 2PC jobs created with options.auto_commit set to "false", but CreateImportJobs currently forwards options only when a non-null timeout is present. A normal caller passing only {"auto_commit":"false"} therefore silently creates an auto-committing job, leaving these new methods without a usable C++ creation path. Please forward the supplied options independently of timeout and add a create-to-commit/abort test.

test/ut/TestBulkImport.cpp
```diff
@@ -0,0 +1,78 @@
...
+    };
+    server.Post("/v2/vectordb/jobs/import/commit", handler);
+    server.Post("/v2/vectordb/jobs/import/abort", handler);
+
+    const auto port = server.bind_to_any_port("127.0.0.1");
+    ASSERT_GT(port, 0);
+    std::thread server_thread([&]() { server.listen_after_bind(); });
+
+    const auto url = "http://127.0.0.1:" + std::to_string(port);
+    auto commit_response = milvus::BulkImport::CommitImport(url, "commit-job", "commit-db", "commit-token");
```
yhmo: test/ut/TestBulkImport.cpp line:51
Low ---- The real REST handlers parse jobId as int64 and return code 1100 for nonnumeric values, so this happy-path mock currently accepts inputs Milvus rejects. Please use numeric strings such as "123" and "456" and assert those payloads so the test demonstrates a server-valid caller pattern.

src/include/milvus/BulkImport.h
```diff
@@ -47,6 +47,20 @@ class MILVUS_SDK_API BulkImport {
     static nlohmann::json
     GetImportJobProgress(const std::string& url, const std::string& job_id, const std::string& db_name = "default",
                          const std::string& api_key = "");
+
+    /**
+     * @brief Commit an import job by restful api.
```
yhmo: src/include/milvus/BulkImport.h line:52
Low ---- These public comments should state the 2PC effects: CommitImport makes staged imported data visible, while AbortImport discards it. The current wording only repeats the method names, so callers cannot learn the destructive lifecycle distinction from the C++ API documentation.

src/include/milvus/BulkImport.h
```diff
@@ -47,6 +47,20 @@ class MILVUS_SDK_API BulkImport {
     static nlohmann::json
     GetImportJobProgress(const std::string& url, const std::string& job_id, const std::string& db_name = "default",
                          const std::string& api_key = "");
+
+    /**
+     * @brief Commit a 2PC import job created with options.auto_commit=false, making its staged imported data visible.
```
yhmo, line 52: src/include/milvus/BulkImport.h line:52
Medium ---- `CreateImportJobs` accepts raw JSON, but Milvus decodes `options` as a string map and rejects a JSON boolean here; the test correctly uses `{"auto_commit", "false"}`. Writing `options.auto_commit=false` in the public contract encourages callers to pass boolean `false`, so please show the required string form explicitly and mirror it in the `AbortImport` comment.

PR #540, opened 2026-07-14
Support minhash function

yhmo:


examples/src/v2/minhash_function.cpp
```diff
@@ -0,0 +1,204 @@
...
+    auto request = milvus::SearchRequest()
+                       .WithCollectionName(collection_name)
+                       .WithAnnsField(signature_field)
+                       .AddEmbeddedText(text)
+                       .WithMetricType(milvus::MetricType::MHJACCARD)
+                       .AddExtraParam("mh_search_with_jaccard", "true")
+                       .AddExtraParam("refine_k", "50")
+                       .WithLimit(top_k)
+                       .AddOutputField(id_field)
+                       .AddOutputField(text_field)
```
yhmo, line 121: examples/src/v2/minhash_function.cpp line:121
Low ---- This requests the full text value for every hit, but deduplicateTexts only reads id_field and prints from the original texts vector. Since the schema allows texts up to 65,535 bytes and this search runs once per document, the example unnecessarily adds per-hit network and deserialization cost; please remove .AddOutputField(text_field).

PR #541, opened 2026-07-15
Support OrderBy for search/query

yhmo:


src/impl/utils/DqlUtils.cpp
```diff
@@ -62,6 +62,49 @@ SetSearchHighlighter(const T&, proto::milvus::SearchRequest&) {
...
+}
+
+const std::vector<OrderByField>&
+GetOrderByFields(const SearchRequest& request) {
+    return request.OrderByFields();
+}
+
+template <typename T>
+const std::vector<OrderByField>&
+GetOrderByFields(const T&) {
```
yhmo, line 79: src/impl/utils/DqlUtils.cpp line:77
Medium ---- `QueryIteratorRequest` and `SearchIteratorRequest` inherit the new `AddOrderByField`/`WithOrderByFields` APIs, but calls from their exact template instantiations select this generic overload instead of the `QueryRequest`/`SearchRequest` overloads, so a valid typed order-by setting is silently omitted from iterator RPCs. Please add exact iterator overloads (or otherwise constrain the fallback); if order-by is intentionally unsupported for iterators, reject or hide the inherited API rather than accepting and dropping it.

src/impl/utils/DqlUtils.cpp
```diff
@@ -62,6 +63,49 @@ SetSearchHighlighter(const T&, proto::milvus::SearchRequest&) {
     return Status::OK();
 }
 
+template <typename T>
+struct SupportsOrderByFields : std::integral_constant<bool, std::is_base_of<QueryRequest, T>::value ||
```
yhmo: src/impl/utils/DqlUtils.cpp line:67
High ---- `QueryIteratorRequest` and `SearchIteratorRequest` derive from these base types, so this trait serializes `order_by_fields` for both iterator APIs. The current Milvus query and search paths explicitly reject ORDER BY with iterators, which makes the new iterator tests describe requests that always fail. Please exclude iterator request types here (and reject inherited setters during iterator preparation) instead of emitting this parameter.

src/impl/utils/DqlUtils.cpp
```diff
@@ -1376,16 +1420,24 @@ ConvertQueryRequest(const T& request, const std::string& current_db, proto::milv
...
-    // limit/offet etc.
-    auto& params = request.ExtraParams();
-    for (auto& pair : params) {
-        if (pair.first == CLUSTER_ID) {
-            continue;
-        }
+    // limit/offset etc. make a copy of params here to erase some keys, because they are handled separately
+    auto extra_params = request.ExtraParams();
+    extra_params.erase(CLUSTER_ID);
+    extra_params.erase(ORDER_BY_FIELDS);
```
yhmo, line 1426: src/impl/utils/DqlUtils.cpp line:1426
Medium ---- Erasing this key unconditionally breaks the existing `AddExtraParam("order_by_fields", ...)` escape hatch: when the typed vector is empty, the caller's raw value is silently dropped. The same occurs in search conversion at line 1553. Please preserve the raw value unless typed fields are present, or return an explicit conflict error when both forms are supplied.

src/impl/types/OrderByField.cpp
```diff
@@ -0,0 +1,73 @@
...
+    SetDirection(direction);
+    return *this;
+}
+
+Status
+OrderByField::Validate() const {
+    if (field_name_.empty()) {
+        return {StatusCode::INVALID_ARGUMENT, "OrderByField field name cannot be empty"};
+    }
+    if (field_name_.find_first_of(",:") != std::string::npos) {
```
yhmo: src/impl/types/OrderByField.cpp line:64
Medium ---- Search ORDER BY supports dynamic JSON paths such as `metadata["a:b"]`, and the server parser deliberately ignores colons inside brackets. Rejecting every colon makes that valid search field impossible through the typed API (the new test even codifies the wrong rejection). Please validate delimiters in a request-aware or bracket-aware way rather than banning all colons.

src/impl/request/dql/SearchRequest.cpp
```diff
@@ -235,12 +235,40 @@ SearchRequest::WithSearchAggregation(const SearchAggregationPtr& aggregation) {
...
+    return *this;
+}
+
 Status
 SearchRequest::Validate() const {
     auto status = SearchRequestBase::Validate();
     if (!status.IsOk()) {
         return status;
     }
+    for (const auto& order_by_field : order_by_fields_) {
```
yhmo: src/impl/request/dql/SearchRequest.cpp line:266
Medium ---- `SearchRequest::Validate()` returns OK for `WithRerank(...)` plus order fields, although Milvus rejects this combination because both define result ordering. Since both values are locally available and this validator runs before the RPC, please reject the conflict here and cover it with a unit test.

src/include/milvus/request/dql/QueryRequest.h
```diff
@@ -180,10 +182,35 @@ class MILVUS_SDK_API QueryRequest : public DQLRequestBase<QueryRequest> {
...
+    OrderByFields() const;
+
+    /**
+     * @brief Set fields used to order query results.
+     */
+    void
+    SetOrderByFields(std::vector<OrderByField>&& order_by_fields);
+
+    /**
+     * @brief Set fields used to order query results.
```
yhmo, line 198: src/include/milvus/request/dql/QueryRequest.h line:198
Low ---- Query ORDER BY requires an explicit positive limit; without one the server rejects the request to avoid unbounded sort-buffer memory. The new public API does not mention this precondition, so callers can follow the generated docs and still construct an unusable request. Please document the `WithLimit` requirement here, and preferably validate it before serialization.

src/impl/utils/DqlUtils.cpp
yhmo: no one will call AddExtraParam("order_by_fields", ...)

src/impl/request/dql/SearchRequest.cpp
yhmo: no need

src/impl/utils/DqlUtils.cpp
yhmo, line 1426: src/impl/utils/DqlUtils.cpp line:1426
Medium ---- Erasing this key unconditionally silently drops the documented raw AddExtraParam("order_by_fields", ...) path whenever typed OrderByFields is empty; the same happens for search at line 1553. This is a compatibility regression and also removes the escape hatch for server options not represented by OrderByField, such as explicit null ordering. Please preserve the raw value when no typed fields are present, or reject a conflict explicitly when both forms are supplied.

yhmo: no need

PR #542, opened 2026-07-15
Support search by id

yhmo:


src/include/milvus/request/dql/SearchRequest.h
```diff
@@ -36,6 +37,40 @@ class MILVUS_SDK_API SearchRequest : public DQLRequestBase<SearchRequest>,
...
+     */
+    void
+    SetIDs(std::vector<std::string>&& id_array);
+
+    /**
+     * @brief Set integer primary keys whose vectors are used as search targets.
+     * Note: IDs and target vectors cannot be specified at the same time.
+     */
+    SearchRequest&
+    WithIDs(std::vector<int64_t>&& id_array);
```
yhmo, line 68: src/include/milvus/request/dql/SearchRequest.h line:65
Medium ---- Because SearchIteratorRequest publicly inherits SearchRequest, this new WithIDs API is exposed there too, but both iterator implementations still call CheckInput(args_.TargetVectors(), ...) instead of SearchRequest::Validate(). An IDs-only iterator therefore fails with "vector_data for search cannot be empty"; if a caller supplies both one vector and IDs, CheckInput passes and ConvertSearchRequest serializes the IDs, so multiple IDs can violate the iterator requirement that nq == 1. Please either hide or explicitly reject these methods for SearchIteratorRequest, or update iterator validation and conversion with a regression test to support exactly the intended ID case.

test/ut/utils/TestDqlUtils.cpp
```diff
@@ -538,6 +538,37 @@ TEST_F(DqlUtilsTest, ConvertSearchRequestV2) {
     EXPECT_EQ(highlight_query.at(0).at("text"), "milvus");
 }
 
+TEST_F(DqlUtilsTest, ConvertSearchRequestV2WithIntegerIDs) {
```
yhmo, line 549: test/ut/utils/TestDqlUtils.cpp line:541
Medium ---- These tests stop at protobuf construction, so they do not prove that Milvus accepts ID targets or that MilvusClientV2::Search() returns one public result group per requested ID. Please add a system Search() test that sends WithIDs through the client and asserts the returned SearchResponse (ideally for both integer and string primary keys); otherwise this user-facing request/response path can regress while all new tests still pass.

src/impl/request/dql/SearchRequest.cpp
```diff
@@ -23,6 +23,33 @@
...
 namespace milvus {
 
+const IDArray&
+SearchRequest::IDs() const {
+    return ids_;
+}
+
+void
+SearchRequest::SetIDs(std::vector<int64_t>&& id_array) {
+    ids_ = IDArray(std::move(id_array));
```
yhmo, line 33: src/impl/request/dql/SearchRequest.cpp line:33
Low ---- IDArray declares a copy constructor but no move assignment, so assigning this temporary selects copy assignment and copies the entire ID vector after the caller’s vector was moved into the temporary. For large string-ID searches this adds an avoidable O(n) allocation/copy before protobuf serialization; please add explicit copy/move assignment operators to IDArray (or otherwise move directly into ids_) so these rvalue setters preserve move semantics.

PR #544, opened 2026-07-16
Support nullable vector

yhmo:


src/impl/utils/DqlUtils.cpp
```diff
@@ -366,6 +366,46 @@ GetValidData(const google::protobuf::RepeatedField<bool>& proto_valid, size_t of
...
+void
+GetNullableVectorRange(const google::protobuf::RepeatedField<bool>& proto_valid, size_t offset,
+                       const std::vector<bool>& valid_data, size_t& vector_offset, size_t& vector_count) {
+    vector_offset = offset;
+    if (proto_valid.empty()) {
+        return;
+    }
+
+    const auto logical_offset = std::min(offset, static_cast<size_t>(proto_valid.size()));
+    vector_offset = static_cast<size_t>(std::count(proto_valid.begin(), proto_valid.begin() + logical_offset, true));
```
yhmo: src/impl/utils/DqlUtils.cpp line:378
Medium ---- ConvertSearchResults calls this helper once per output field for every query with increasing offsets; recounting valid_data from begin() to offset each time makes nullable-vector conversion quadratic in nq for uniform topk. Please compute packed offsets incrementally or precompute prefix counts so each validity entry is scanned only once.

src/impl/utils/DqlUtils.cpp
```diff
@@ -377,51 +417,78 @@ CreateMilvusFieldData(const proto::schema::FieldData& proto_data, size_t offset,
...
             field_data =
                 std::make_shared<BinaryVecFieldData>(std::move(name), std::move(vectors), std::move(valid_data));
             return Status::OK();
         }
         case proto::schema::DataType::FloatVector: {
             const auto& floats = proto_vectors.float_vector().data();
             std::vector<FloatVecFieldData::ElementT> vectors = BuildFieldDataVectors<float, float>(
-                proto_vectors.dim() * 4, floats.data(), floats.size(), offset, count);
+                proto_vectors.dim() * 4, floats.data(), floats.size(), vector_offset, vector_count);
+            auto status = ExpandNullableVectorData(name, valid_data, vectors);
```
yhmo, line 451: src/impl/utils/DqlUtils.cpp line:441
High ---- After nullable vectors are expanded and stored with valid_data here, iterator batching still drops that metadata: CopyFieldDataRange constructs sliced fields from Data() only, and AppendFieldData calls Append(src_data), which marks every appended element valid. QueryIteratorImpl::copyResults and SearchIteratorImpl::FetchPageFromCache use those helpers, so a null vector crossing a page slice or append is returned as a non-null empty vector. Please preserve, slice, and append ValidData alongside Data and add an iterator-oriented regression test.

test/ut/utils/TestDmlUtils.cpp
```diff
@@ -657,3 +658,119 @@ TEST_F(DmlUtilsTest, CreateProtoFieldDataWithNullableInt) {
...
+    schema.AddField(milvus::FieldSchema("float", milvus::DataType::FLOAT_VECTOR).WithDimension(2).WithNullable(true));
+    schema.AddField(milvus::FieldSchema("sparse", milvus::DataType::SPARSE_FLOAT_VECTOR).WithNullable(true));
+    schema.AddField(
+        milvus::FieldSchema("float16", milvus::DataType::FLOAT16_VECTOR).WithDimension(2).WithNullable(true));
+    schema.AddField(
+        milvus::FieldSchema("bfloat16", milvus::DataType::BFLOAT16_VECTOR).WithDimension(2).WithNullable(true));
+    schema.AddField(milvus::FieldSchema("int8", milvus::DataType::INT8_VECTOR).WithDimension(2).WithNullable(true));
+
+    auto binary = std::make_shared<milvus::BinaryVecFieldData>("binary");
+    binary->Add({1});
```
yhmo: test/ut/utils/TestDmlUtils.cpp line:697
Medium ---- This fixture always adds a non-null dense vector before `AddNull()`, so it misses the leading-null case. `FieldData::AddNull()` stores an empty `T{}` at `data_[0]`, and the next `Add(...)` compares its dimension with `data_[0].size()`, returning `DIMENSION_NOT_EQUAL` for binary/float/FP16/BF16/int8 columns. Nullable vectors must allow null in the first row, so please cover `AddNull()` followed by `Add(valid)` and adjust the dimension check to use the first valid vector (or the schema dimension).

src/impl/types/FieldData.cpp
```diff
@@ -200,6 +202,29 @@ FieldData<T, Dt>::Append(const std::vector<T>& elements) {
     return StatusCode::OK;
 }
 
+template <typename T, DataType Dt>
+StatusCode
+FieldData<T, Dt>::Append(const std::vector<T>& elements, const std::vector<bool>& valid_data) {
+    if (!valid_data.empty() && elements.size() != valid_data.size()) {
```
yhmo, line 223: src/impl/types/FieldData.cpp line:208
Medium ---- This only validates mask length. For vector specializations, equal-length input such as elements={{}, {1,2}} with valid_data={true,true}, or two valid vectors with different dimensions, returns OK; CreateProtoVectorField then emits fewer or misaligned values than the true bits advertise, so insert/upsert reaches the server with a malformed payload. Please validate every valid vector with the same non-empty and dimension invariants as Add() before mutating, and document the mask contract.

src/impl/types/FieldData.cpp
```diff
@@ -78,7 +78,8 @@ AddElement(const T& element, std::vector<T>& array) {
         return StatusCode::VECTOR_IS_EMPTY;
     }
 
-    if (!array.empty() && element.size() != array.at(0).size()) {
+    const auto first_valid = std::find_if(array.begin(), array.end(), [](const T& item) { return !item.empty(); });
```
yhmo: src/impl/types/FieldData.cpp line:81
Medium ---- This fixes leading nulls functionally, but it rescans the entire prefix on every later Add(). With N leading nulls followed by M valid vectors, construction performs O(N*M) empty checks instead of the previous constant-time dimension check. Please retain the first valid dimension or search from the end so the leading prefix is paid at most once.

src/impl/types/FieldData.cpp
yhmo: will fix it in another pr

src/impl/utils/DmlUtils.cpp
```diff
@@ -324,42 +324,65 @@ ParseSparseFloatVector(const nlohmann::json& obj, const std::string& field_name,
...
-    auto dim = data.front().size();
+    auto dim = schema_dim;
     auto& vectors_data = *(ret->mutable_float_vector()->mutable_data());
-    vectors_data.Reserve(static_cast<int>(data.size() * dim));
-    for (const auto& item : data) {
+    if (dim > 0) {
+        vectors_data.Reserve(static_cast<int>(data.size() * dim));
+    }
+    for (size_t i = 0; i < data.size(); ++i) {
+        if (nullable && field.IsNull(i)) {
```
liliu-z, line 403: In the column insert path nothing enforces `valid_data_.size() == Count()`: `CopyValidData` copies `valid_data_` verbatim while `CreateProtoVectorField` decides which rows to skip via `field.IsNull(i)`. A FieldData built with the data-only constructor (which leaves `valid_data_` empty) and then mutated with `AddNull()` gets a validity bitmap shorter than its rows — e.g. `FloatVecFieldData("vec", {{1,2}})` then `AddNull()` yields `Count()==2` but `ValidData()=={false}` — so `IsNull` reports the real vector at index 0 as null (dropping it) and treats trailing rows as valid, silently serializing corrupted null/row associations. Reject the mismatch (require `valid_data_.size() == Count()` or empty) before serializing; the same gap exists in all six vector overloads.

src/impl/utils/DmlUtils.cpp
```diff
@@ -324,42 +324,65 @@ ParseSparseFloatVector(const nlohmann::json& obj, const std::string& field_name,
...
+            continue;
+        }
+        const auto& item = data[i];
+        if (dim == 0) {
```
liliu-z, line 407: Unlike the row path (`CheckAndSetFloatVector` rejects `obj.size() != fs.Dimension()`), `CreateProtoVectorField` packs vector elements by total count without checking that each non-null row equals `schema_dim`. Ragged data built via the bulk constructor (which bypasses `Add`'s per-element dimension check) then serializes with wrong row boundaries whenever the lengths sum to `Count() * schema_dim` — e.g. `{{1},{2,3,4}}` at dim 2 emits 4 floats with `dim=2`, accepted server-side as two dim-2 rows. Because the PR now derives `dim` from `schema_dim` rather than the first row's length, this coincidental-sum case slips through where the old logic would have produced a server-rejected mismatch; validate each non-null item's length against `schema_dim`. Applies to all six vector overloads.

src/impl/utils/DmlUtils.cpp
```diff
@@ -324,42 +324,101 @@ ParseSparseFloatVector(const nlohmann::json& obj, const std::string& field_name,
...
+CreateProtoVectorField(const FloatVecFieldData& field, bool nullable, int64_t schema_dim) {
     auto ret = new proto::schema::VectorField{};
     auto& data = field.Data();
-    auto dim = data.front().size();
+    auto dim = schema_dim;
     auto& vectors_data = *(ret->mutable_float_vector()->mutable_data());
-    vectors_data.Reserve(static_cast<int>(data.size() * dim));
-    for (const auto& item : data) {
+    if (dim > 0) {
+        vectors_data.Reserve(static_cast<int>(data.size() * dim));
```
yhmo: src/impl/utils/DmlUtils.cpp line:394
Medium ---- This reserve still sizes the protobuf buffer as `data.size() * dim`, even though the loop skips null rows. A mostly/all-null high-dimensional column can therefore allocate a full dense matrix (potentially gigabytes) while serializing little or no vector data, turning valid nullable inserts into OOM/latency failures. Please reserve using the non-null row count (or let the buffer grow from the packed payload) instead.

PR #545, opened 2026-07-17
Support nullable vector

yhmo:


scripts/run_examples.sh
```diff
@@ -0,0 +1,79 @@
...
+if [[ -x "${EXTERNAL_TABLE_EXAMPLE}" ]]; then
+    echo "Skipping sdk_external_table_v2 because it requires pre-populated external object storage."
+fi
+
+readonly OPTIMIZE_EXAMPLE="${EXAMPLE_DIR}/sdk_optimize_v2"
+if [[ -x "${OPTIMIZE_EXAMPLE}" ]]; then
+    echo "Skipping sdk_optimize_v2."
+fi
+
+mapfile -d '' examples < <(find "${EXAMPLE_DIR}" -maxdepth 1 -type f -name "sdk_*_${version}" \
```
yhmo, line 54: scripts/run_examples.sh line:54
Medium ---- This new runner fails on the repository's supported macOS setup: the system Bash 3.2 does not provide `mapfile`, and BSD `sort` does not support `-z`, so the script exits before running any example. Please enumerate the executables with a Bash 3-compatible or glob-based loop so the documented macOS development environment can use it.

scripts/run_examples.sh
```diff
@@ -0,0 +1,79 @@
...
+    ! -name "sdk_cdc_v2" ! -name "sdk_optimize_v2" \
```
yhmo, line 55: scripts/run_examples.sh line:55
Low ---- The script reports `sdk_external_table_v2` as skipped above, but this filter only excludes CDC and Optimize. If that executable is present, the loop will run it and unexpectedly require pre-populated object storage; please also exclude `sdk_external_table_v2` here.

test/ut/utils/TestDmlUtils.cpp
```diff
@@ -657,3 +658,186 @@ TEST_F(DmlUtilsTest, CreateProtoFieldDataWithNullableInt) {
...
+
+    auto it = std::find_if(rpc_fields.begin(), rpc_fields.end(),
+                           [](const auto& field) { return field.field_name() == "vec"; });
+    ASSERT_NE(it, rpc_fields.end());
+    EXPECT_THAT(it->valid_data(), ElementsAre(false, false));
+    EXPECT_EQ(it->vectors().dim(), 2);
+    EXPECT_TRUE(it->vectors().float_vector().data().empty());
+}
+
+TEST_F(DmlUtilsTest, NullableVectorColumnRoundTrip) {
```
yhmo, line 706: test/ut/utils/TestDmlUtils.cpp line:706
Medium ---- This round-trip feeds the C++ encoder directly into the C++ decoder, so matching mistakes in the packed nullable-vector layout can cancel out. Because the system-test harness is pinned to Milvus v2.6.18, please add a `test/st` case that creates a nullable vector field, inserts mixed and all-null rows, and verifies query and search output from the real server.

scripts/run_examples.sh
yhmo: no need

PR #546, opened 2026-07-17
Support field-level operations for upsert

yhmo:


src/impl/utils/TypeUtils.cpp
```diff
@@ -150,6 +150,20 @@ FunctionTypeCast(proto::schema::FunctionType type) {
...
+FieldPartialUpdateOpTypeCast(FieldPartialUpdateOp::OpType type) {
+    switch (type) {
+        case FieldPartialUpdateOp::OpType::REPLACE:
+            return proto::schema::FieldPartialUpdateOp::REPLACE;
+        case FieldPartialUpdateOp::OpType::ARRAY_APPEND:
+            return proto::schema::FieldPartialUpdateOp::ARRAY_APPEND;
+        case FieldPartialUpdateOp::OpType::ARRAY_REMOVE:
+            return proto::schema::FieldPartialUpdateOp::ARRAY_REMOVE;
+        default:
+            return proto::schema::FieldPartialUpdateOp::REPLACE;
```
yhmo: src/impl/utils/TypeUtils.cpp line:163
Medium ---- `static_cast<FieldPartialUpdateOp::OpType>(99)` reaches this fallback, while `UpsertRequest::PartialUpdate()` treats the same value as non-`REPLACE` and enables partial update. The request therefore succeeds as a partial `REPLACE` instead of rejecting an invalid operation, which can overwrite data unexpectedly. Please validate the enum during request validation and return `INVALID_ARGUMENT` rather than silently coercing it.

src/include/milvus/types/FieldPartialUpdateOp.h
```diff
@@ -0,0 +1,62 @@
...
+#include "milvus/Export.h"
+
+namespace milvus {
+
+/**
+ * @brief Describes how a field value is applied during a partial upsert.
+ */
+class MILVUS_SDK_API FieldPartialUpdateOp {
+ public:
+    enum class OpType {
```
yhmo, line 38: src/include/milvus/types/FieldPartialUpdateOp.h line:30
Low ---- Please document the public contract for these values here: `ARRAY_APPEND`/`ARRAY_REMOVE` are array-only, append must stay within `max_capacity`, remove deletes every matching occurrence, and the named field must also be present in the upsert payload. Those rules exist in the pinned proto comments, but SDK users generating Doxygen from this header currently only see the names and otherwise learn the constraints from server errors.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1613,22 +1629,22 @@ MilvusClientV2Impl::Upsert(const UpsertRequest& request, UpsertResponse& respons
...
+            status = CheckAndSetRowData(rows, collection_desc->Schema(), true, request.PartialUpdate(), rpc_fields);
             if (!status.IsOk()) {
                 return status;
             }
         } else if (!fields.empty()) {
             // verify column-based data
             // if the collection is already recreated, some schema might be changed, we need to update the
             // collectionDesc cache and call CheckInsertInput() again.
-            status = CheckInsertInput(collection_desc, fields, true);
+            status = CheckInsertInput(collection_desc, fields, true, request.PartialUpdate());
```
yhmo, line 1655: src/impl/MilvusClientV2Impl.cpp line:1640
Medium ---- Please return any non-OK status from `CheckInsertInput` before calling `CreateProtoFieldDatas`. The new unequal-column-count check returns `INVALID_ARGUMENT`, but this path overwrites it with conversion success; for example, columns with counts 2 and 1 are sent with `num_rows = 2` instead of being rejected, so the helper test does not reflect `MilvusClientV2::Upsert` behavior.

src/impl/utils/DmlUtils.cpp
```diff
@@ -1489,13 +1497,16 @@ CheckAndSetFieldValue(const nlohmann::json& obj, const FieldSchema& fs, proto::s
...
 Status
-CheckAndSetRowData(const EntityRows& rows, const CollectionSchema& schema, bool partial_upsert,
+CheckAndSetRowData(const EntityRows& rows, const CollectionSchema& schema, bool is_upsert, bool partial_upsert,
                    std::vector<proto::schema::FieldData>& rpc_fields) {
     std::set<std::string> output_fields;
     GetOutputFields(schema, output_fields);
 
     const std::vector<FieldSchema>& normal_fields = schema.Fields();
     const std::vector<StructFieldSchema>& struct_fields = schema.StructFields();
+    if (partial_upsert && !struct_fields.empty()) {
```
yhmo: src/impl/utils/DmlUtils.cpp line:1507
High ---- This rejects every partial upsert as soon as the collection contains any struct field, including a request that omits the struct or replaces the whole struct. The server's partial-upsert path supports whole-struct REPLACE and rejects only struct sub-field updates, and the previous C++ flow serialized a supplied whole struct through ProcessStructFieldValues; this therefore breaks valid existing callers. Please keep whole-struct handling and only reject unsupported sub-field updates.

src/impl/utils/DmlUtils.cpp
```diff
@@ -1546,6 +1573,17 @@ CheckAndSetRowData(const EntityRows& rows, const CollectionSchema& schema, bool
         }
     }
 
+    if (partial_upsert) {
+        size_t expected_count = 0;
+        for (const auto& pair : field_counts) {
+            if (expected_count == 0) {
+                expected_count = pair.second;
+            } else if (pair.second != expected_count) {
```
yhmo: src/impl/utils/DmlUtils.cpp line:1581
Medium ---- Comparing field counts only with each other misses batches where all supplied fields occur equally fewer times than rows.size(); for example, [{id:1}, {tags:[1]}] gives both counts as 1, passes here, then sends num_rows=2 with one value in each column and fails server alignment validation. Please compare every non-empty field count with rows.size() and add the disjoint/empty-row case.

src/impl/utils/DmlUtils.cpp
```diff
@@ -1520,8 +1532,23 @@ CheckAndSetRowData(const EntityRows& rows, const CollectionSchema& schema, bool
...
         }
 
+        for (auto it = row.begin(); it != row.end(); ++it) {
+            const auto& name = it.key();
+            if (output_fields.find(name) != output_fields.end()) {
+                return {StatusCode::INVALID_ARGUMENT, "Function output field cannot be provided: " + name};
+            }
+            if (field_names.find(name) == field_names.end()) {
+                if (!schema.EnableDynamicField()) {
+                    return {StatusCode::INVALID_ARGUMENT, "Not a valid field: " + name};
```
liliu-z: When a row carries a field absent from the cached schema, the row path returns a terminal INVALID_ARGUMENT ("Not a valid field"), whereas the column path returns DATA_UNMATCH_SCHEMA for the identical case to drive a cache-refresh and re-check. This check runs in `validate` before the RPC is sent, so it also short-circuits the server-driven self-heal (stale schema_timestamp -> server SchemaMismatch -> cache evict -> retry) that the pre-PR row path relied on to recover. During an online schema migration that adds a field, a row-based writer with a stale cache now fails terminally where it previously succeeded after one retry. Return DATA_UNMATCH_SCHEMA (or let the RPC through so the server's SchemaMismatch drives the refresh) instead of a terminal error.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1656,6 +1708,11 @@ MilvusClientV2Impl::Upsert(const UpsertRequest& request, UpsertResponse& respons
         rpc_request.set_num_rows(static_cast<uint32_t>(row_count));
         rpc_request.set_schema_timestamp(collection_desc->UpdateTime());
         rpc_request.set_partial_update(request.PartialUpdate());
+        for (const auto& field_op : request.FieldOps()) {
+            auto* rpc_field_op = rpc_request.add_field_ops();
+            rpc_field_op->set_field_name(field_op.FieldName());
+            rpc_field_op->set_op(FieldPartialUpdateOpTypeCast(field_op.GetOpType()));
+        }
```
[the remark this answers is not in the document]
yhmo: no need

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1656,6 +1706,11 @@ MilvusClientV2Impl::Upsert(const UpsertRequest& request, UpsertResponse& respons
         rpc_request.set_num_rows(static_cast<uint32_t>(row_count));
         rpc_request.set_schema_timestamp(collection_desc->UpdateTime());
         rpc_request.set_partial_update(request.PartialUpdate());
+        for (const auto& field_op : request.FieldOps()) {
+            auto* rpc_field_op = rpc_request.add_field_ops();
```
liliu-z, line 1694: `MilvusClientV2Impl.cpp:1710` auto-sets `partial_update=true` and emits `field_ops`, but a 2.6.x server older than 2.6.17 — advertised as supported by the generic `| 2.6.x |` row in `README.md:20` — ignores the unknown `field_ops` field under proto3 while still honoring `partial_update`. That silently converts ARRAY_APPEND/REMOVE into a full REPLACE, wiping existing array contents with no error returned to the caller. Add a minimum-server-version guard (>= 2.6.17) before enabling field_ops, or document the required server version.

test/ut/utils/TestDmlUtils.cpp
```diff
@@ -438,12 +458,141 @@ TEST_F(DmlUtilsTest, CheckAndSetRowDataSimple) {
...
+    schema.AddField(milvus::FieldSchema("pk", milvus::DataType::INT64, "pk", true, false));
+    schema.AddField(milvus::FieldSchema("tags", milvus::DataType::ARRAY)
+                        .WithElementType(milvus::DataType::INT32)
+                        .WithMaxCapacity(10));
+
+    auto expect_inconsistent = [&schema](milvus::EntityRows rows) {
+        std::vector<milvus::proto::schema::FieldData> rpc_fields;
+        auto status = milvus::CheckAndSetRowData(rows, schema, true, true, rpc_fields);
+        EXPECT_EQ(status.Code(), milvus::StatusCode::INVALID_ARGUMENT);
+        EXPECT_EQ(status.Message(), "The row count of partial update fields is inconsistent");
```
liliu-z, line 761: The upsert cases at `test/ut/utils/TestDmlUtils.cpp:579` omit the PK field, so `Upsert` validation returns `"The field: pk is not provided."` and aborts before the row-count check the test is meant to exercise. The missing-PK error fires first on every run, so the assertion fails deterministically in CI and the intended path is never tested. Include the PK in each case's field data.

test/st/cases/TestDml.cpp
```diff
@@ -87,11 +119,92 @@ class MilvusServerTestDml : public MilvusServerTest {
...
+        nlohmann::json{{"id", 1}},
+        nlohmann::json{{"f_bool", true}},
+    };
+    status = client_->Upsert(milvus::UpsertRequest()
+                                 .WithCollectionName(collection_name)
+                                 .WithPartialUpdate(true)
+                                 .WithRowsData(std::move(rows)),
+                             response);
+    EXPECT_EQ(status.Code(), milvus::StatusCode::INVALID_ARGUMENT);
+    EXPECT_EQ(status.Message(), "The row count of partial update fields is inconsistent");
```
liliu-z, line 205: `test/st/cases/TestDml.cpp:205` repeats the unit test's defect: the upsert case omits the PK, so validation returns `"The field: pk is not provided."` before the intended row-count check and the system test fails deterministically in CI. Add the PK to this case's field data as well.

src/impl/MilvusClientImpl.cpp
```diff
@@ -908,14 +914,14 @@ MilvusClientImpl::Insert(const std::string& collection_name, const std::string&
...
             status = getCollectionDesc(collection_name, true, collection_desc);
             if (!status.IsOk()) {
                 return status;
             }
 
-            status = CheckInsertInput(collection_desc, fields, false);
+            status = CheckInsertInput(collection_desc, fields, false, false);
         }
 
         return CreateProtoFieldDatas(collection_desc->Schema(), fields, rpc_fields);
```
liliu-z, line 930: At `MilvusClientImpl.cpp:927` the V1 column `Insert` path discards the new unequal-column-count `INVALID_ARGUMENT` status instead of returning on it, then proceeds to send malformed field data to the server. V1 `Upsert` and both V2 paths already return early on this check, so only this path forwards a bad request when the column counts don't match. Return the status when it isn't OK before assembling the request.

test/it/v2/TestSearch.cpp
```diff
@@ -0,0 +1,244 @@
...
+    request.WithFilter("id > 0");
+    request.WithLimit(1);
+    request.AddFloatVector(std::vector<float>{0.1f, 0.2f, 0.3f, 0.4f});
+    return request;
+}
+
+}  // namespace
+
+TEST_F(UnconnectMilvusMockedTest, SearchResponseExtraInfoMetadata) {
+    auto client = CreateConnectedV2Client(service_, server_.ListenPort());
```
yhmo, line 70: test/it/v2/TestSearch.cpp line:70
Medium ---- This PR is scoped to field-level upsert operations, but it also relocates or adds hundreds of lines of unrelated search, hybrid-search, nullable-vector, and V2 test-directory coverage. That makes the DML behavior change harder to isolate and increases merge-conflict risk for unrelated work; could you split this test reorganization and coverage into a separate PR?

yhmo: no

src/impl/utils/DmlUtils.cpp
```diff
@@ -1520,14 +1538,42 @@ CheckAndSetRowData(const EntityRows& rows, const CollectionSchema& schema, bool
                     "The No." + std::to_string(i) + " input row is not a JSON dict object"};
         }
 
+        for (auto it = row.begin(); it != row.end(); ++it) {
+            const auto& name = it.key();
+            if (output_fields.find(name) != output_fields.end()) {
+                return {StatusCode::INVALID_ARGUMENT, "Function output field cannot be provided: " + name};
```
yhmo: src/impl/utils/DmlUtils.cpp line:1544
Medium ---- Returning INVALID_ARGUMENT here prevents the row-based callers from refreshing a stale collection schema, because they retry validation only for DATA_UNMATCH_SCHEMA. If a collection is recreated or altered so a field that the cache still marks as a function output is now a normal input, a valid insert or upsert fails before any RPC; the column path returns DATA_UNMATCH_SCHEMA for the same mismatch. Please preserve that refresh signal and cover this stale-function case.

src/impl/utils/GtsDict.cpp
```diff
@@ -79,7 +79,7 @@ GetNowMs() {
 int64_t
 MakeMktsFromNowMs() {
     auto ms = GetNowMs();
-    return ms << 18;
+    return (ms + 1000) << 18;
```
yhmo, line 82: src/impl/utils/GtsDict.cpp line:82
Medium ---- This helper supplies the iterator session_ts fallback when an older server returns zero, and the server uses that value directly as the continuation MvccTimestamp. Adding one second therefore moves the next page to a future snapshot: it can wait or exceed short RPC deadlines and can include writes that occurred after the first page, breaking stable iteration. Please keep the fallback at the observed current time or derive the first-page snapshot without a future offset.

yhmo: Align with pymilvus

yhmo, line 82: src/impl/utils/GtsDict.cpp line:82
Medium ---- This value is used as the next page guarantee_timestamp when an older server returns session_ts=0. Advancing it by one second makes every such Query/Search iterator wait for a future HLC (and can make sub-second RPC deadlines fail), even though this PR is unrelated to iterator consistency. Please keep the fallback at the current time unless the future timestamp is required and its wait/timeout behavior is handled explicitly.

test/it/v2/TestUpsert.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+    auto ids = std::make_shared<milvus::Int64FieldData>("id", std::vector<int64_t>{1});
+    milvus::UpsertRequest request;
+    request.WithCollectionName("partial_update_coll")
+        .WithColumnsData({ids})
+        .AddFieldOp(milvus::FieldPartialUpdateOp("tags", milvus::FieldPartialUpdateOp::OpType::ARRAY_APPEND));
+
+    milvus::UpsertResponse response;
+    auto status = client->Upsert(request, response);
+    EXPECT_TRUE(status.IsOk()) << status.Message();
```
yhmo: test/it/v2/TestUpsert.cpp line:205
Low ---- This mock reports success for ARRAY_APPEND on tags even though tags is absent from fields_data. The Milvus proxy rejects non-REPLACE field ops without a matching payload field, and the new public documentation says the field must be present, so this test records an impossible real-server success path. Please return/assert the server error here, or reject the request before issuing the RPC.

PR #547, opened 2026-07-22
Support field-level operations for upsert

yhmo:


src/impl/utils/GtsDict.cpp
```diff
@@ -79,7 +79,7 @@ GetNowMs() {
 int64_t
 MakeMktsFromNowMs() {
     auto ms = GetNowMs();
-    return ms << 18;
+    return (ms + 1000) << 18;
```
yhmo, line 82: src/impl/utils/GtsDict.cpp line:82
High ---- Adding 1000 ms makes the fallback MVCC timestamp one second ahead of the client clock. SearchIteratorImpl, SearchIteratorV2Impl, and QueryIteratorImpl reuse this value as guarantee_timestamp when old servers return session_ts == 0, and QueryNode waits until tSafe reaches it; later iterator pages can therefore stall for about a second or hit user RPC deadlines below 1 s. Please keep the fallback at the current time, or derive a server-safe timestamp without moving it into the future.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1483,6 +1517,11 @@ MilvusClientV2Impl::Upsert(const UpsertRequest& request, UpsertResponse& respons
         rpc_request.set_num_rows(static_cast<uint32_t>(row_count));
         rpc_request.set_schema_timestamp(collection_desc->UpdateTime());
         rpc_request.set_partial_update(request.PartialUpdate());
+        for (const auto& field_op : request.FieldOps()) {
```
yhmo, line 1520: src/impl/MilvusClientV2Impl.cpp line:1520
Medium ---- Please validate each field operation before building the RPC. The public type permits an empty field name and a cast unsupported enum value, and this loop currently forwards both so callers receive a remote SERVER_FAILED after a network round trip; the Java 2.6 sibling rejects the same malformed operations as invalid parameters. Returning INVALID_ARGUMENT from the validation phase would keep SDK error semantics deterministic.

src/impl/MilvusClientV2Impl.h
```diff
@@ -354,6 +354,30 @@ class MilvusClientV2Impl : public MilvusClientV2, public std::enable_shared_from
     RemovePrivilegesFromGroup(const RemovePrivilegesFromGroupRequest& request) final;
 
  private:
+    Status
+    insert(const InsertRequest& request, InsertResponse& response, bool allow_retry);
+
+    Status
+    upsert(const UpsertRequest& request, UpsertResponse& response, bool allow_retry);
+
+    Status
```
yhmo: src/impl/MilvusClientV2Impl.h line:363
Low ---- These six search/query declarations have no definitions or call sites on this branch and are unrelated to the insert/upsert retry helpers introduced here. Leaving declaration-only private paths creates misleading dead surface and future drift risk; please remove them until their implementations are introduced.

src/impl/utils/GtsDict.cpp
yhmo: as designed

src/impl/MilvusClientV2Impl.cpp
yhmo: no need

src/impl/utils/GtsDict.cpp
yhmo, line 82: src/impl/utils/GtsDict.cpp line:82
Medium ---- Following up on the resolved latency thread with a separate correctness consequence: iterator requests copy this fallback directly into both MvccTimestamp and GuaranteeTimestamp, so (ms + 1000) << 18 makes later pages read at an MVCC point a full second after the first-page fallback when an old server returns session_ts == 0. Writes or updates committed in that interval can appear only on later pages, causing inconsistent pagination and possible duplicates or misses. Please keep the fallback at the current client time, as the matching Java iterators do, unless the server returns a snapshot timestamp.

PR #548, opened 2026-07-22
Align collection schema evolution APIs with AlterCollectionSchema

sijie-ni-0214:
## Summary

- Route `AddCollectionField` through `AlterCollectionSchema`, with a fallback to the legacy RPC only when an older server returns gRPC `Unimplemented`.
- Extend `AddFunctionFieldRequest` with required bound-index metadata and serialize the index name and flat index parameters into `FieldInfo`.
- Validate BM25 and MinHash function output fields and reject missing, automatic, or conflicting bound-index declarations before sending the RPC.
- Mark the legacy add/drop collection-function APIs as deprecated in favor of function-field operations.
- Add request, mock integration, and example coverage for schema routing, fallback behavior, bound-index serialization, and error propagation.

issue: #549

## Compatibility

- New Milvus servers use `AlterCollectionSchema` for `AddCollectionField`.
- Older servers that do not implement `AlterCollectionSchema` continue to work through the legacy `AddCollectionField` RPC.
- Fallback is limited to the transport-level gRPC `Unimplemented` status; application errors and other transport errors are returned unchanged.
- External-collection AddField support relies on the corresponding upstream Milvus server support and is not special-cased in the SDK.

## Related Milvus work

- https://github.com/milvus-io/milvus/pull/51411
- https://github.com/milvus-io/milvus/issues/48983

## Test Plan

- [x] 739 unit tests
- [x] 249 mocked integration tests
- [x] Master E2E for Add/Drop Field, BM25, and MinHash function fields
- [x] Milvus 2.6 compatibility E2E verifying the legacy AddCollectionField fallback
- [x] `git diff --check`
- [x] clang-format check for all changed C++ files

### DIFF
--- a/src/impl/MilvusClientV2Impl.cpp
+++ b/src/impl/MilvusClientV2Impl.cpp
@@ -704,7 +704,24 @@ MilvusClientV2Impl::DropCollectionFieldProperties(const DropCollectionFieldPrope
 
 Status
 MilvusClientV2Impl::AddCollectionField(const AddCollectionFieldRequest& request) {
-    auto pre = [&request](proto::milvus::AddCollectionFieldRequest& rpc_request) {
+    auto alter_pre = [&request](proto::milvus::AlterCollectionSchemaRequest& rpc_request) {
+        rpc_request.set_db_name(request.DatabaseName());
+        rpc_request.set_collection_name(request.CollectionName());
+
+        auto* field_info = rpc_request.mutable_action()->mutable_add_request()->add_field_infos();
+        ConvertFieldSchema(request.Field(), *field_info->mutable_field_schema());
+        return Status::OK();
+    };
+
+    auto status =
+        connection_.Invoke<proto::milvus::AlterCollectionSchemaRequest, proto::milvus::AlterCollectionSchemaResponse>(
+            alter_pre, &MilvusConnection::AlterCollectionSchema);
+    // An older coordinator reports an internal gRPC UNIMPLEMENTED as Milvus ErrServiceUnimplemented (code 10).
+    if (status.RpcErrCode() != static_cast<int32_t>(::grpc::StatusCode::UNIMPLEMENTED) && status.ServerCode() != 10) {
+        return status;
+    }
+
+    auto legacy_pre = [&request](proto::milvus::AddCollectionFieldRequest& rpc_request) {
         rpc_request.set_db_name(request.DatabaseName());
         rpc_request.set_collection_name(request.CollectionName());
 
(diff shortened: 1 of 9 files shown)


src/impl/MilvusClientV2Impl.cpp
```diff
@@ -857,6 +873,45 @@ MilvusClientV2Impl::AddFunctionField(const AddFunctionFieldRequest& request) {
...
+                break;
+            case FunctionType::MINHASH:
+                if (request.Field().FieldDataType() != DataType::BINARY_VECTOR) {
+                    return Status{StatusCode::INVALID_ARGUMENT,
+                                  "MinHash function output field must be a binary vector."};
+                }
+                break;
+            default:
+                return Status{StatusCode::INVALID_ARGUMENT,
+                              "AddFunctionField only supports BM25 and MinHash functions."};
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:892
Low ---- This makes BM25/MinHash-only a public API constraint, but the public AddFunctionField and request docs still describe the function generically. Please document the supported function types there so callers, and future FunctionType additions, do not discover this restriction only at runtime.

sijie-ni-0214: Thanks. I removed the client-side BM25/MinHash function type and output data type restrictions and now rely on server-side semantic validation. This keeps the C++ API generic so future server-supported function types do not require a corresponding SDK change. The SDK still validates the client-owned bound-index contract.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -704,7 +704,23 @@ MilvusClientV2Impl::DropCollectionFieldProperties(const DropCollectionFieldPrope
...
+
+        auto* field_info = rpc_request.mutable_action()->mutable_add_request()->add_field_infos();
+        ConvertFieldSchema(request.Field(), *field_info->mutable_field_schema());
+        return Status::OK();
+    };
+
+    auto status =
+        connection_.Invoke<proto::milvus::AlterCollectionSchemaRequest, proto::milvus::AlterCollectionSchemaResponse>(
+            alter_pre, &MilvusConnection::AlterCollectionSchema);
+    if (status.RpcErrCode() != static_cast<int32_t>(::grpc::StatusCode::UNIMPLEMENTED)) {
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:719
Medium ---- This fallback misses the business-level "service unimplemented" response used in mixed-version deployments: a new proxy connected to an older coordinator returns alter_status.code == 10 with no gRPC error. The related Milvus implementation explicitly falls back for both that status and transport UNIMPLEMENTED; checking only RpcErrCode() makes AddCollectionField fail during rolling upgrades instead of using the legacy RPC. Please also recognize status.ServerCode() == 10 and cover that path.

sijie-ni-0214: Thanks, fixed. AddCollectionField now falls back for both transport-level gRPC UNIMPLEMENTED and the business-level service-unimplemented status (ServerCode 10) returned by a new proxy connected to an older coordinator. I added a mocked mixed-version fallback test and kept the existing ServerCode 12 non-fallback test because Milvus business code 12 means resource insufficient. All 8 AddCollectionField integration tests pass.

PR #550, opened 2026-07-23
Auto publish packages

yhmo:
Automates the release process when an exact vX.Y.Z tag is pushed.

  - Publishes the GitHub release immediately instead of creating a draft.
  - Triggers package builds for all supported Linux, macOS, and Windows targets.
  - Documents GRPC_VERBOSITY=ERROR for suppressing verbose gRPC example logs.
  - Documents the StorageV3 requirement for the add-field function demo.
  - Skips sdk_add_field_v2 in the batch example runner because it requires special server
    configuration.
  - Updates the Milvus image used by system tests.

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -1,4 +1,4 @@
-name: Create draft release
+name: Create release
 
 on:
   push:
@@ -7,17 +7,18 @@ on:
 
 permissions:
   contents: write
+  actions: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
   cancel-in-progress: true
 
 jobs:
-  # create a draft release for an exact release tag (vX.Y.Z).
+  # Create and publish a release for an exact release tag (vX.Y.Z).
   # Tags that do not match this format are ignored, which allows pushing other tags (e.g. vX.Y)
   # without triggering the release workflow.
-  draft-release:
-    name: Create draft release
+  release:
+    name: Create release
     runs-on: ubuntu-latest
     steps:
       - name: Prepare release metadata
@@ -26,7 +27,7 @@ jobs:
           RELEASE_TAG: ${{ github.ref_name }}
         run: |
           if ! echo "$RELEASE_TAG" | grep -Eq '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
-            echo "Tag $RELEASE_TAG is not an exact release tag in vX.Y.Z format. Skipping draft release."
+            echo "Tag $RELEASE_TAG is not an exact release tag in vX.Y.Z format. Skipping release."
             echo "is_release_tag=false" >> "$GITHUB_OUTPUT"
             exit 0
           fi
@@ -38,7 +39,7 @@ jobs:
           echo "version=${version}" >> "$GITHUB_OUTPUT"
           echo "release_date=${release_date}" >> "$GITHUB_OUTPUT"
           echo "milvus_minor=${milvus_minor}" >> "$GITHUB_OUTPUT"
-      - name: Create draft release
+      - name: Create release
         if: steps.release-info.outputs.is_release_tag == 'true'
         uses: ncipollo/release-action@v1
         with:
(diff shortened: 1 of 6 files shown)


examples/src/v2/add_field.cpp
```diff
@@ -153,6 +153,7 @@ main(int argc, char* argv[]) {
     }
 
     {
+        // StorageV3 must be enabled on the Milvus server before running this function-backed field demo.
```
yhmo: examples/src/v2/add_field.cpp line:156
Medium ---- StorageV3 alone is not sufficient for this demo on the Milvus image pinned in this PR. `AddFunctionField` also rejects the request unless `dataCoord.compaction.bumpSchemaVersion.enabled=true` (its default is false; `storageVersion.enabled` is also checked but defaults true), so a user following this comment still hits an error. Please document all non-default prerequisites here.

scripts/run_examples.sh
```diff
@@ -52,7 +52,7 @@ if [[ -x "${OPTIMIZE_EXAMPLE}" ]]; then
 fi
 
 mapfile -d '' examples < <(find "${EXAMPLE_DIR}" -maxdepth 1 -type f -name "sdk_*_${version}" \
-    ! -name "sdk_cdc_v2" ! -name "sdk_external_table_v2" ! -name "sdk_optimize_v2" \
+    ! -name "sdk_cdc_v2" ! -name "sdk_external_table_v2" ! -name "sdk_optimize_v2" ! -name "sdk_add_field_v2"\
```
yhmo: scripts/run_examples.sh line:55
Low ---- This silently excludes `sdk_add_field_v2`, while the three existing exclusions above each print a skip reason. The script can now end with “All v2 examples passed” without telling the operator this example was not run; please add the same explicit skip notice for this binary.

.github/workflows/create-release.yaml
```diff
@@ -47,8 +48,26 @@ jobs:
...
           omitNameDuringUpdate: true
           omitBodyDuringUpdate: true
           token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Trigger release package builds
+        if: steps.release-info.outputs.is_release_tag == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RELEASE_TAG: ${{ github.ref_name }}
+        run: |
+          gh workflow run release.yaml \
```
yhmo, line 63: .github/workflows/create-release.yaml line:62
High ---- This job has no checkout and does not set GH_REPO or pass --repo, so gh cannot determine which repository owns release.yaml and the command fails before dispatching any package builds. Because the preceding step has already published the release, every valid tag leaves a public release with no assets; please provide repository context, for example by setting GH_REPO to github.repository or passing --repo "$GITHUB_REPOSITORY".

liliu-z:
**.github/workflows/release.yaml:45** The release step sets `draft: false`, so GitHub publishes the release publicly as soon as the workflow reaches it rather than after every package has finished building and uploading. Combined with the job reporting success even when individual builds fail, each release is publicly visible but empty for the upload window (which can stretch to hours) and can land permanently incomplete while the workflow still shows green. Create the release as a draft (`draft: true`) and only publish it once all expected artifacts have been built and attached.

scripts/run_examples.sh
```diff
@@ -36,30 +36,41 @@ if [[ ! -d "${EXAMPLE_DIR}" ]]; then
...
+        sdk_add_field_v2)
+            echo "Skipping ${example_name} because its function-backed field demo requires additional server settings."
+            ;;
+        *)
+            examples+=("${example}")
+            ;;
+    esac
+done
+
+if [[ ${#examples[@]} -eq 0 ]]; then
```
liliu-z, line 69: When the example list is empty after the exclusion filter, the script now falls through to `exit 0` where it previously `exit 1`'d, so it reports success despite running no examples. A build that produced only the four excluded binaries would pass silently, hiding that nothing actually ran; the exit code is not currently consumed by automation and the trigger is contrived, but a run that executes nothing should not report success. Restore the non-zero exit when the post-filter list is empty.

PR #556, opened 2026-07-24
Fix stale collection schema cache after alias mutations

yhmo:
create_alias/alter_alias/drop_alias never touch collection_desc_cache, which getCollectionDesc keys by the caller-supplied name — an alias when the user passes one. After repointing an alias, getCollectionDesc resolve the primary key from the stale cached schema, so if the new target has a different PK/schema they build wrong requests against it; insert/upsert self-heal via the schema-timestamp mismatch retry and query() never consults the cache, the cache also never auto-expires. Evict the (db, alias) cache entry on every alias mutation.

### DIFF
--- a/src/impl/MilvusClientV2Impl.cpp
+++ b/src/impl/MilvusClientV2Impl.cpp
@@ -37,6 +37,31 @@
 #include "utils/TypeUtils.h"
 
 namespace milvus {
+namespace {
+
+// A gRPC deadline or ambiguous transport failure can be reported after Milvus has committed a metadata mutation.
+// Invalidate cached collection descriptions in that case, but retain them for definitive local or server rejections,
+// including retry exhaustion caused by server-side rate limiting.
+bool
+isAmbiguousMutationFailure(const Status& status) {
+    switch (status.RpcErrCode()) {
+        case ::grpc::StatusCode::OK:
+        case ::grpc::StatusCode::INVALID_ARGUMENT:
+        case ::grpc::StatusCode::NOT_FOUND:
+        case ::grpc::StatusCode::ALREADY_EXISTS:
+        case ::grpc::StatusCode::PERMISSION_DENIED:
+        case ::grpc::StatusCode::UNAUTHENTICATED:
+        case ::grpc::StatusCode::RESOURCE_EXHAUSTED:
+        case ::grpc::StatusCode::FAILED_PRECONDITION:
+        case ::grpc::StatusCode::OUT_OF_RANGE:
+        case ::grpc::StatusCode::UNIMPLEMENTED:
+            return false;
+        default:
+            return true;
+    }
+}
+
+}  // namespace
 
 std::shared_ptr<MilvusClientV2>
 MilvusClientV2::Create() {
@@ -156,6 +181,8 @@ MilvusClientV2Impl::CreateCollection(const CreateCollectionRequest& request) {
     };
 
     auto post = [this, &request, &schema](const proto::common::Status& rpc_response) {
+        removeCollectionDesc(request.DatabaseName(), request.CollectionName());
+
         if (request.Indexes().empty()) {
             return Status::OK();
         }
@@ -180,8 +207,12 @@ MilvusClientV2Impl::CreateCollection(const CreateCollectionRequest& request) {
         return LoadCollection(load_req);
     };
 
-    return connection_.Invoke<proto::milvus::CreateCollectionRequest, proto::common::Status>(
+    auto status = connection_.Invoke<proto::milvus::CreateCollectionRequest, proto::common::Status>(
         validate, pre, &MilvusConnection::CreateCollection, post);
+    if (isAmbiguousMutationFailure(status)) {
+        removeCollectionDesc(request.DatabaseName(), request.CollectionName());
+    }
+    return status;
 }
 
 Status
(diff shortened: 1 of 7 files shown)


src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3167,7 +3179,8 @@ MilvusClientV2Impl::getCollectionDesc(const std::string& db_name, const std::str
     // the reason is: describeCollection() could be limited by server-side(DDL request throttling is enabled)
     // we don't intend to allow too many threads run into describeCollection() in this method
     std::lock_guard<std::mutex> lock(collection_desc_cache_mtx_);
-    auto it = collection_desc_cache_.find(collection_name);
+    auto name = combineDbCollectionName(actual_db, collection_name);
+    auto it = collection_desc_cache_.find(name);
```
yhmo, line 3223: src/impl/MilvusClientV2Impl.cpp line:3183
High ---- This lookup activates the V2 cache for the first time: entries were already stored as `db|collection`, but lookup previously used only `collection`. That makes existing collection-identity mutations unsafe. For example, after `Compact("old")` caches ID 100, `RenameCollection("old", "new")` leaves `default|old` intact, and creating a new `old` also does not replace it; the next `Compact("old")` reuses ID 100 and can compact the renamed collection instead of the new one. Please invalidate the old key on rename and cover rename/recreate with a cache test, or keep this cache disabled until all identity-changing paths own invalidation.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -967,8 +975,12 @@ MilvusClientV2Impl::AlterAlias(const AlterAliasRequest& request) {
         return Status::OK();
     };
 
-    return connection_.Invoke<proto::milvus::AlterAliasRequest, proto::common::Status>(pre,
-                                                                                       &MilvusConnection::AlterAlias);
+    auto status =
+        connection_.Invoke<proto::milvus::AlterAliasRequest, proto::common::Status>(pre, &MilvusConnection::AlterAlias);
+    if (status.IsOk()) {
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:980
Medium ---- A non-OK transport status does not prove the alias mutation was rejected: Milvus can commit `AlterAlias` and the client can then receive `DEADLINE_EXCEEDED` or a connection failure. This branch then preserves an entry that never expires, so later schema/collection-ID requests keep using the old target. Please invalidate on ambiguous `TIMEOUT`/`RPC_FAILED` outcomes (or conservatively after every dispatched mutation), retaining the entry only for definitive pre-dispatch or server-rejection failures; the same issue applies to Create/Drop and V1.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -37,6 +37,36 @@
 #include "utils/TypeUtils.h"
 
 namespace milvus {
+namespace {
+
+// A timeout or ambiguous transport failure can be reported after Milvus has committed a metadata mutation.
+// Invalidate cached collection descriptions in that case, but retain them for definitive local or server rejections.
+bool
+isAmbiguousMutationFailure(const Status& status) {
+    if (status.Code() == StatusCode::TIMEOUT) {
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:46
Medium ---- `Retry()` also returns `StatusCode::TIMEOUT` after exhausting server-side rate-limit responses, with `RpcErrCode()==0` and server code 8/RateLimit. Those attempts are definitive rejections, but this branch treats them as post-commit ambiguity and evicts a valid schema cache, forcing an extra `DescribeCollection` precisely while the server is overloaded. Please distinguish a gRPC deadline (`RpcErrCode()==DEADLINE_EXCEEDED`) from retry-exhaustion timeouts before invalidating; the same helper in V1 has the same issue.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3167,7 +3224,8 @@ MilvusClientV2Impl::getCollectionDesc(const std::string& db_name, const std::str
     // the reason is: describeCollection() could be limited by server-side(DDL request throttling is enabled)
     // we don't intend to allow too many threads run into describeCollection() in this method
     std::lock_guard<std::mutex> lock(collection_desc_cache_mtx_);
-    auto it = collection_desc_cache_.find(collection_name);
+    auto name = combineDbCollectionName(actual_db, collection_name);
```
yhmo, line 3222: src/impl/MilvusClientV2Impl.cpp line:3227
Low ---- This line makes V2 cache hits database-qualified, but every new cache test uses the default database. A regression here can return the schema/collection ID for the same-named collection in another database. Please add a V2 mocked case that caches the same collection or alias name in two explicit databases and verifies both lookup and mutation eviction stay isolated.

test/st/cases/TestAlias.cpp
```diff
@@ -104,3 +106,63 @@ TEST_F(MilvusServerTestAlias, AlterAlias) {
...
+    ASSERT_NE(first_schema.Fields().end(), first_pk);
+    EXPECT_EQ(milvus::DataType::INT64, first_pk->FieldDataType());
+
+    status =
+        client_->AlterAlias(milvus::AlterAliasRequest().WithCollectionName(collection_name2).WithAlias(alias_name));
+    milvus::test::ExpectStatusOK(status);
+
+    milvus::DescribeCollectionResponse second_desc;
+    status =
+        client_->DescribeCollection(milvus::DescribeCollectionRequest().WithCollectionName(alias_name), second_desc);
```
yhmo, line 145: test/st/cases/TestAlias.cpp line:145
Low ---- This system test does not exercise `collection_desc_cache_`: public `DescribeCollection` always issues a fresh RPC, so the test passes even if the alias invalidation code is removed. Please warm and consume the cache through a cache-backed API such as `Compact`/`Get`/delete-by-ID before and after `AlterAlias`/`DropAlias` so the real-server test validates the regression.

PR #558, opened 2026-07-27
Fix misspelled RandomValue and RandomBools

ausk:


examples/src/v1/array.cpp
```diff
@@ -94,19 +94,19 @@ main(int argc, char* argv[]) {
     const int64_t row_count = 10;
     milvus::EntityRows rows;
     for (auto i = 0; i < row_count; ++i) {
-        auto cap = util::RandomeValue<int>(1, 5);
+        auto cap = util::RandomValue<int>(1, 5);
```
yhmo, line 97: examples/src/v1/array.cpp line:97
High ---- These call sites now use `RandomValue`/`RandomValues`/`RandomBools`, but `examples/utils/ExampleUtils.h` still defines only `RandomeValue`/`RandomeValues`/`RansomBools`. Since examples are built by default, all changed example targets fail name lookup. Please rename (or alias) the helper definitions in the same PR.

ausk: I forgot to sync ExampleUtils.h to codespace; now updated. All examples build successfully.

PR #559, opened 2026-07-27
Refine cache machinery

yhmo:
Refactor collection schema and DML timestamp caching for safer concurrent use and consistent session reads.
  - Introduce thread-safe global SchemaCache and CollectionTsCache.
  - Key cache entries by endpoint, database, and collection.
  - Deduplicate concurrent schema loads.
  - Add consistent cache invalidation for DML, DDL, alias, and database operations.
  - Preserve session timestamps across collection renames.
  - Support renaming collections across databases.
  - Fix query iterator result caching bug when ReduceStopForBest is true.
  - Add comprehensive unit, mocked integration, and end-to-end tests.


src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3838,36 +3839,19 @@ MilvusClientV2Impl::getCollectionDesc(const std::string& db_name, const std::str
...
-
-void
-MilvusClientV2Impl::cleanCollectionDescCache() {
-    std::lock_guard<std::mutex> lock(collection_desc_cache_mtx_);
-    collection_desc_cache_.clear();
+    const auto endpoint = connection_.CurrentEndpoint();
+    return SchemaCache::GetInstance().GetOrLoad(
+        endpoint, actual_db, collection_name, force_update,
+        [this, &actual_db, &collection_name, rpc_timeout_ms](CollectionDescPtr& loaded) {
+            auto request = DescribeCollectionRequest().WithDatabaseName(actual_db).WithCollectionName(collection_name);
```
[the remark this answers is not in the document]
yhmo: as design

src/impl/MilvusClientImpl.cpp
```diff
@@ -2094,38 +2103,25 @@ MilvusClientImpl::RemovePrivilegesFromGroup(const std::string& group_name, const
...
-    collection_desc_cache_.clear();
+    const auto endpoint = connection_.CurrentEndpoint();
+    const auto db_name = connection_.CurrentDbName("");
+    return SchemaCache::GetInstance().GetOrLoad(
+        endpoint, db_name, collection_name, force_update,
+        [this, &collection_name](CollectionDescPtr& loaded) {
+            CollectionDesc desc;
+            auto status = DescribeCollection(collection_name, desc);
+            if (status.IsOk()) {
+                loaded = std::make_shared<CollectionDesc>(std::move(desc));
```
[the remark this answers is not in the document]
yhmo: as design

src/impl/utils/cache/SchemaCache.cpp
```diff
@@ -0,0 +1,215 @@
...
+#include <algorithm>
+
+namespace milvus {
+
+SchemaCache::SchemaCache(size_t capacity) : capacity_(capacity) {
+}
+
+SchemaCache&
+SchemaCache::GetInstance() {
+    static SchemaCache instance;
```
yhmo, line 29: src/impl/utils/cache/SchemaCache.cpp line:28
Medium ---- Making this cache process-wide means schema entries survive client destruction and reconnect. If the same endpoint/database/collection is dropped and recreated by another process, or the server behind the endpoint is replaced, a fresh client hits this old entry and iterator preparation can send a stale collection ID and primary-key schema without forcing a refresh. Please scope the entry to a client/connection generation or add an explicit expiry/invalidation boundary.

src/impl/utils/cache/SchemaCache.cpp
```diff
@@ -0,0 +1,215 @@
...
+        return finish_load(Status::OK(), current_desc);
+    }
+
+    const auto load_generation = mutation_generation_.load(std::memory_order_acquire);
+    CollectionDescPtr loaded;
+    auto status = loader(loaded);
+    if (status.IsOk()) {
+        std::unique_lock<std::shared_timed_mutex> lock(mutex_);
+        // Do not repopulate an entry invalidated while the RPC was in flight.
+        if (mutation_generation_.load(std::memory_order_relaxed) == load_generation) {
```
yhmo: src/impl/utils/cache/SchemaCache.cpp line:103
Low ---- `mutation_generation_` covers the entire cache, so setting or invalidating any unrelated collection while this DescribeCollection call is in flight causes the successful load here not to be cached. Under multi-collection churn, requests for collection A can repeatedly issue DescribeCollection merely because collection B changes, defeating this cache and increasing DDL-throttling pressure. Please track invalidation per key, or otherwise suppress insertion only when this key was invalidated.

src/impl/utils/cache/SchemaCache.cpp
yhmo: as design

test/ut/utils/TestCollectionTsCache.cpp
```diff
@@ -0,0 +1,109 @@
...
+TEST(CollectionTsCacheTest, ConcurrentReadsAndMonotonicWrites) {
+    milvus::CollectionTsCache cache;
+    constexpr int kThreadCount = 8;
+    constexpr int kWritesPerThread = 1000;
+
+    std::vector<std::thread> threads;
+    threads.reserve(kThreadCount);
+    for (int thread_id = 0; thread_id < kThreadCount; ++thread_id) {
+        threads.emplace_back([&cache, thread_id]() {
+            for (int i = 1; i <= kWritesPerThread; ++i) {
```
liliu-z, line 220: Confirmed against the Windows CI log: `TestCollectionTsCache.cpp(96): error C3493: 'kWritesPerThread' cannot be implicitly captured`. Under `-std:c++14` MSVC treats the `constexpr int kWritesPerThread` used inside the thread lambda `[&cache, thread_id]` as needing an explicit capture (GCC/Clang correctly do not). This is a hard compile error, not a warnings-as-errors case — the compile flags are `/W3` with no `/WX`. Add `kWritesPerThread` to the lambda's capture list and apply the identical fix to the twin at TestSchemaCache.cpp:150. Merge blocker: the Windows job is red.

test/ut/utils/TestSchemaCache.cpp
```diff
@@ -0,0 +1,258 @@
...
+    constexpr int kThreadCount = 8;
+    std::atomic<int> entered{0};
+    std::atomic<int> load_count{0};
+    std::mutex gate_mutex;
+    std::condition_variable gate_cv;
+
+    auto loader = [&](milvus::CollectionDescPtr& desc) {
+        load_count.fetch_add(1);
+        std::unique_lock<std::mutex> lock(gate_mutex);
+        gate_cv.wait(lock, [&entered]() { return entered.load() == kThreadCount; });
```
liliu-z: Same MSVC C3493 as the TestCollectionTsCache.cpp:96 twin: the nested lambda `[&entered]() { return entered.load() == kThreadCount; }` references `constexpr int kThreadCount` without capturing it, which MSVC under `-std:c++14` rejects as a hard error. It is absent from the current CI log only because Ninja stopped at the first failure (TestCollectionTsCache.cpp:96); once that is fixed this line fails identically. Add `kThreadCount` to the inner lambda's capture list. This is a hard C3493, not a warnings-as-errors issue.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3838,36 +3839,19 @@ MilvusClientV2Impl::getCollectionDesc(const std::string& db_name, const std::str
...
-    }
-    return status;
-}
-
-void
-MilvusClientV2Impl::cleanCollectionDescCache() {
-    std::lock_guard<std::mutex> lock(collection_desc_cache_mtx_);
-    collection_desc_cache_.clear();
+    const auto endpoint = connection_.CurrentEndpoint();
+    return SchemaCache::GetInstance().GetOrLoad(
```
liliu-z, line 4011: This PR makes the V2 schema cache functional (the old code stored under `db|collection` but looked up by `collection_name`, so it never hit) and keys it on the user-supplied `collection_name`, which may be an alias. CreateAlias/AlterAlias/DropAlias have no post-callback and never invalidate the cache, so after AlterAlias re-points an alias, `getCollectionDesc(alias)` keeps returning the old target's cached schema (PK field name, fields, collection ID). The insert/upsert paths partly self-heal via the DATA_UNMATCH_SCHEMA force-refresh, and the DML/DQL RPCs carry the alias name which the server re-resolves at request time, so data is not misrouted; but the search-result PK-name fallback (old servers) and the SearchIterator metric-type lookup use the stale schema with no retry. Invalidate the cache on Create/Alter/Drop alias in both V1 and V2.

src/impl/utils/cache/CollectionTsCache.cpp
```diff
@@ -0,0 +1,115 @@
...
+    if (cache_.size() <= capacity_) {
+        return;
+    }
+
+    const auto oldest = std::min_element(cache_.begin(), cache_.end(), [](const auto& left, const auto& right) {
+        return left.second->last_access_.load(std::memory_order_relaxed) <
+               right.second->last_access_.load(std::memory_order_relaxed);
+    });
+    if (oldest != cache_.end()) {
+        cache_.erase(oldest);
```
liliu-z: The session write-timestamp store changed from the old unbounded `GtsDict` map (which never evicted) to an LRU cache capped at DEFAULT_CAPACITY=4096; `evictIfNeededLocked` drops the least-recently-used entry once the cap is exceeded. After eviction `CollectionTsCache::Get` returns 0 and `DeduceGuaranteeTimestamp` (DqlUtils.cpp:1401) falls back to `guarantee_ts = 1`, the weakest bound, so a SESSION-consistency read of a collection that was written but then evicted silently degrades to eventually-consistent with no error or signal. This is reachable under many-collection multi-tenant load and is a behavioral regression versus the previous unbounded store. Surface the downgrade or refresh/persist the session timestamp rather than silently dropping to 1.

src/impl/utils/cache/SchemaCache.cpp
```diff
@@ -0,0 +1,241 @@
...
+void
+SchemaCache::Invalidate(const std::string& endpoint, const std::string& db_name, const std::string& collection_name) {
+    const auto key = CollectionCacheKey::Create(endpoint, db_name, collection_name);
+    invalidateLoad(key);
+    std::unique_lock<std::shared_timed_mutex> lock(mutex_);
+    cache_.erase(key);
+}
+
+void
+SchemaCache::InvalidateDb(const std::string& endpoint, const std::string& db_name) {
```
liliu-z, line 146: Cache invalidation is wired into DropCollection, TruncateCollection, and the DML failure path, but not into other schema/name-mutating DDLs — even though the PR added `InvalidateDb` specifically for this, and `InvalidateDb` has no production caller (only a unit test). DropDatabase leaves stale entries for every collection in the dropped db; RenameCollection and AddCollectionField likewise never invalidate. Most cases self-heal (a dropped/renamed target makes the follow-up RPC fail, clearing the entry), so severity is low, but this is the same class as the alias gap and InvalidateDb is dead code until wired. Wire DropDatabase to InvalidateDb and the rename/schema-evolution paths to Invalidate.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1643,12 +1685,14 @@ MilvusClientV2Impl::insert(const InsertRequest& request, InsertResponse& respons
         results.SetInsertCount(static_cast<uint64_t>(rpc_response.insert_cnt()));
         response.SetResults(std::move(results));
 
+        auto actual_db = connection_.CurrentDbName(request.DatabaseName());
+        const auto endpoint = connection_.CurrentEndpoint();
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:1689
Medium ---- apiHandler snapshots the MilvusConnection used for the RPC, but this post callback reads endpoint/database from the mutable ConnectionHandler after the RPC. If another thread reconnects or switches databases while the insert is in flight, a timestamp returned by connection A can be stored under the process-wide cache key for connection B; subsequent SESSION reads on B may use a timestamp from a different Milvus cluster. Please carry the endpoint/database snapshot associated with the actual RPC through validation/post-processing; the same race also affects schema loading and the sibling DML/DDL callbacks.

yhmo: > Concurrent connection-state mutations are outside the client’s thread-safety contract. A connected client
  > supports concurrent RPC API calls, but Connect, Disconnect, UseDatabase, and destruction must not run
  > concurrently with in-flight RPCs. Under this contract, the endpoint and database remain stable throughout
  > validation, RPC execution, and post-processing, so cache entries cannot be associated with another
  > connection. We will document this limitation rather than add connection-generation propagation throughout
  > every callback.

src/impl/utils/cache/SchemaCache.cpp
```diff
@@ -0,0 +1,241 @@
...
+    bool is_loader = false;
+    {
+        std::lock_guard<std::mutex> lock(loading_mutex_);
+        auto it = loading_.find(key);
+        if (it == loading_.end()) {
+            load_state = std::make_shared<LoadState>();
+            loading_.emplace(key, load_state);
+            is_loader = true;
+        } else {
+            load_state = it->second;
```
yhmo, line 59: src/impl/utils/cache/SchemaCache.cpp line:53
Medium ---- `Invalidate()` only marks the existing `LoadState` and leaves it in `loading_`, so a caller that starts after a successful schema mutation can still attach here and receive the pre-mutation descriptor at lines 60-63. This regresses the old mutex ordering: with a delayed `DescribeCollection`, then `AddCollectionField` or drop/recreate, an iterator started before the delayed response returns can use a stale collection ID/PK schema. Please detach/replace invalidated load states (or make waiters retry when `invalidated_` is set) and cover a waiter that begins after invalidation.

src/impl/utils/cache/CollectionTsCache.cpp
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    uint64_t ts = 0;
+    auto it = cache_.find(old_key);
+    if (it != cache_.end()) {
+        ts = it->second;
+    }
+
+    cache_.erase(old_key);
+    cache_.erase(new_key);
```
yhmo: src/impl/utils/cache/CollectionTsCache.cpp line:97
High ---- Rename unconditionally deletes any timestamp already stored for the destination and then installs only the old-name value. Because RPCs are documented as concurrent, a write through the new collection name can finish after the server-side rename but before this callback runs; its newer timestamp is then replaced (or erased when the old key is absent), so a following SESSION read can miss that successful write. Please preserve the maximum of the old and destination timestamps, and cover this DDL/DML ordering in a concurrency test.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1195,8 +1283,16 @@ MilvusClientV2Impl::AlterAlias(const AlterAliasRequest& request) {
         return Status::OK();
     };
 
-    return connection_.Invoke<proto::milvus::AlterAliasRequest, proto::common::Status>(pre,
-                                                                                       &MilvusConnection::AlterAlias);
+    auto post = [this, &request](const proto::common::Status&) {
+        auto actual_db = connection_.CurrentDbName(request.DatabaseName());
+        const auto endpoint = connection_.CurrentEndpoint();
+        CollectionTsCache::GetInstance().Invalidate(endpoint, actual_db, request.Alias());
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:1289
High ---- This post-success invalidation can erase a timestamp from a concurrent write through the alias. After AlterAlias has taken effect server-side, another thread may complete an insert, upsert, or delete through that alias and call CollectionTsCache::Set before this callback acquires the cache lock; this Invalidate then drops the new timestamp, allowing a subsequent SESSION read to miss the write. The same ordering exists in the new CreateAlias callbacks and V1 counterparts, so the invalidation needs to remove only the pre-mutation state or otherwise use generation-aware updates.

yhmo: rare case, no need to handle

src/include/milvus/types/IteratorArguments.h
```diff
@@ -95,7 +95,7 @@ class MILVUS_SDK_API QueryIteratorArguments : public IteratorArguments, public Q
     SetReduceStopForBest(bool reduce_stop_for_best);
 
  private:
-    bool reduce_stop_for_best_{false};
+    bool reduce_stop_for_best_{true};
```
liliu-z, line 98: Flipping the QueryIteratorArguments reduce_stop_for_best_ default to true lets the server return more than batch_size rows to the query iterator, which for the first time exercises the cache-cutting branch in QueryIteratorImpl::Next() (QueryIteratorImpl.cpp:51-58). That branch is broken: line 54 copies the leftover cached rows into temp_results (overwriting the batch just placed there at line 46) instead of into the new_cache declared at line 53, then assigns the still-empty new_cache to cache_. Because copyResults overwrites its target, the current batch is lost, cache_ is emptied, and updateCursor advances past the leftover rows, so iteration silently skips rows and returns fewer than match. Fix line 54 to copy into new_cache.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -765,9 +795,15 @@ MilvusClientV2Impl::DropCollectionField(const DropCollectionFieldRequest& reques
         return Status::OK();
     };
 
+    auto post = [this, &request](const proto::milvus::AlterCollectionSchemaResponse&) {
+        const auto actual_db = connection_.CurrentDbName(request.DatabaseName());
+        SchemaCache::GetInstance().Invalidate(connection_.CurrentEndpoint(), actual_db, request.CollectionName());
```
liliu-z: The schema-mutation post-lambdas invalidate SchemaCache only for request.CollectionName(). If a client reached the collection through an alias, its schema was cached under the alias key and survives this invalidation, so DQL issued through the alias keeps using the pre-mutation schema. There is no TTL (the cache is LRU-bounded), and DML self-heals via the SchemaMismatch retry that invalidates the name actually used, but DQL through the alias stays stale until the entry is evicted. The same by-name-only invalidation is repeated across every alter/add/drop field and function post-lambda here and in the V1 client.

test/it/v1/TestCacheInvalidation.cpp
```diff
@@ -0,0 +1,184 @@
...
+    EXPECT_CALL(service_, RenameCollection(_, _, _))
+        .WillOnce([](::grpc::ServerContext*, const milvus::proto::milvus::RenameCollectionRequest*,
+                     milvus::proto::common::Status*) { return ::grpc::Status{}; });
+
+    auto status = client_->RenameCollection("old", "new");
+    EXPECT_TRUE(status.IsOk());
+    ExpectSchemaNotCached(endpoint, "default", "old");
+    ExpectSchemaNotCached(endpoint, "default", "new");
+    EXPECT_EQ(milvus::CollectionTsCache::GetInstance().Get(endpoint, "default", "old"), 0);
+    EXPECT_EQ(milvus::CollectionTsCache::GetInstance().Get(endpoint, "default", "new"), 100);
```
liliu-z: After renaming old->new with both timestamps cached, CollectionTsCache::Rename keeps max(100,200)=200 under the new name, as the RenamesCollection unit test asserts. The two rename integration tests instead expect the moved timestamp to be 100 — test/it/v1/TestCacheInvalidation.cpp:127 and test/it/v2/TestCacheInvalidation.cpp:138 — so both fail on every run. Since the integration test dirs are GLOB_RECURSE'd into the build, CI is deterministically red with no underlying product defect (these are cached session timestamps, not status codes). Change both expected values to 200.

src/impl/MilvusClientV2Impl.cpp
yhmo: The delayed-response ordering described here is possible when rename and DML overlap on the same collection. We do not guarantee semantic ordering for concurrent collection-lifecycle/schema DDL and DML/DQL operations. The thread-safety documentation on both MilvusClient and MilvusClientV2 now explicitly requires these conflicting operations to be serialized. Concurrent DML/DQL remains supported, with the normal rule that an overlapping DQL is not guaranteed to observe a DML until that DML completes. Generation-aware redirection would add substantial lifecycle complexity and the same limitation currently exists in the C++ SDK and PyMilvus, so we are handling this by defining the supported concurrency contract rather than adding cache-level ordering machinery.

src/impl/utils/cache/SchemaCache.cpp
```diff
@@ -0,0 +1,246 @@
...
+    // A different loader might have populated the cache between the first lookup and
+    // registration in loading_. For force refresh, a different cached pointer means
+    // another refresh completed after this call's initial lookup.
+    CollectionDescPtr current_desc;
+    if (getCached(key, current_desc) && (!force_update || current_desc != initial_desc)) {
+        return finish_load(Status::OK(), current_desc);
+    }
+
+    CollectionDescPtr loaded;
+    auto status = loader(loaded);
```
yhmo: src/impl/utils/cache/SchemaCache.cpp line:98
Medium ---- Once this key is registered in `loading_`, an exception from `loader` bypasses `finish_load`, so same-key waiters remain blocked forever and the stale load state is never erased. This is reachable because `ConvertCollectionSchema` intentionally throws `nlohmann::json::parse_error` for malformed `external_spec`. Please make every exit after registration complete/erase the load state (for example with a scope guard or catch path) and cover a throwing loader with a concurrent waiter.

src/include/milvus/request/dql/QueryIteratorRequest.h
```diff
@@ -51,7 +51,7 @@ class MILVUS_SDK_API QueryIteratorRequest : public IteratorArguments, public Que
     WithReduceStopForBest(bool reduce_stop_for_best);
 
  private:
-    bool reduce_stop_for_best_{false};
+    bool reduce_stop_for_best_{true};
```
yhmo, line 66: src/include/milvus/request/dql/QueryIteratorRequest.h line:54
Medium ---- Is this public default change intentional as part of the cache refactor? Existing C++ callers that omit `WithReduceStopForBest` now send `true` instead of the `false` default that has existed since query iterators were introduced; the mirrored V1 field changes too. Java/PyMilvus parity may justify it, but please either keep the compatible default or explicitly document/changelog the behavior change and add end-to-end evidence for unchanged callers. This is separate from the cache-cutting bug fixed elsewhere in this PR.

yhmo: align with pymilvus

src/impl/types/SearchIteratorV2Impl.cpp
```diff
@@ -149,7 +149,8 @@ SearchIteratorV2Impl<T>::executeSearch(const T& args, proto::milvus::SearchResul
         args.DatabaseName().empty() ? connection_->GetConnectParam().DbName() : args.DatabaseName();
 
     proto::milvus::SearchRequest rpc_request;
-    auto status = ConvertSearchRequest<T>(args, current_db, rpc_request, cluster_id_);
+    auto status =
+        ConvertSearchRequest<T>(args, current_db, rpc_request, cluster_id_, connection_->GetConnectParam().Uri());
```
yhmo, line 153: src/impl/types/SearchIteratorV2Impl.cpp line:153
Medium ---- ConvertSearchRequest now injects the cached DML timestamp for SESSION, but the unchanged branch below overwrites it with 0 whenever session_ts_ == 0, including the first real page. That page can read a pre-write snapshot and then store its older session_ts_, pinning the rest of the iterator to stale data. Please preserve the deduced timestamp for non-probe requests and add a test that seeds CollectionTsCache and asserts the first-page guarantee_timestamp.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1170,8 +1244,16 @@ MilvusClientV2Impl::CreateAlias(const CreateAliasRequest& request) {
         return Status::OK();
     };
 
-    return connection_.Invoke<proto::milvus::CreateAliasRequest, proto::common::Status>(pre,
-                                                                                        &MilvusConnection::CreateAlias);
+    auto post = [this, &request](const proto::common::Status&) {
+        auto actual_db = connection_.CurrentDbName(request.DatabaseName());
+        const auto endpoint = connection_.CurrentEndpoint();
+        CollectionTsCache::GetInstance().Invalidate(endpoint, actual_db, request.Alias());
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:1250
Medium ---- Separate from the resolved concurrent-invalidation thread, this also breaks a fully serialized session flow: after a write through `request.CollectionName()`, `CreateAlias` clears the alias timestamp instead of inheriting the target collection timestamp, so `DeduceGuaranteeTimestamp(alias)` falls back to `1` and an immediate `SESSION` read through the alias may miss that completed write. `AlterAlias` and the V1 hooks have the same issue; please invalidate the old alias value and then seed it from the target collection timestamp, with a canonical-write → alias → SESSION-read test.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3824,75 +4069,34 @@ MilvusClientV2Impl::getFlushState(const std::string& db_name, const std::vector<
...
-    // if connection is connected to "default", the input db_name is "" or "default", actual_db is "default"
-    // if connection is connected to "A" but the input db_name is "B", actual_db is "B"
-    // if connection is connected to "A" but the input db_name is "", actual_db is "A"
-    // if connection is connected to "A" but the input db_name is "A", actual_db is "A"
-    auto actual_db = connection_.CurrentDbName(db_name);
-
-    auto name = combineDbCollectionName(actual_db, collection_name);
-    std::lock_guard<std::mutex> lock(collection_desc_cache_mtx_);
-    collection_desc_cache_.erase(name);
+    return SchemaCache::GetInstance().GetOrLoad(
```
yhmo, line 4011: src/impl/MilvusClientV2Impl.cpp line:4075
Medium ---- The earlier process-wide staleness thread was dispositioned as design, but this is a narrower boundary the client can handle: neither V1 nor V2 CreateCollection invalidates this completed cache entry after a successful create. If another process drops a collection and this client recreates the same name with a different schema, the next insert, delete-by-ID, compact, or iterator can reuse the previous CollectionDesc (including collection ID and primary-key schema) until a later server error happens to refresh it. Please invalidate SchemaCache for the created endpoint/database/name on successful creation and add a recreate regression test.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -107,6 +107,37 @@ ConnectionHandler::CurrentDbName(const std::string& overwrite_db_name) const {
     return "";
 }
 
+std::string
+ConnectionHandler::CurrentEndpoint() const {
```
yhmo, line 111: src/impl/utils/ConnectionHandler.cpp line:111
Low ---- This helper is newly added but has no call sites; all changed paths use RpcContext::cache_.endpoint_ instead. Keeping a second unused way to read endpoint state expands the internal API without a distinct contract, so please remove it until a caller needs it.

src/impl/utils/cache/CacheContext.h
```diff
@@ -0,0 +1,37 @@
...
+namespace milvus {
+
+// Immutable cache identity captured from the connection used by an RPC.
+struct CacheContext {
+    std::string endpoint_;
+    std::string database_name_;
+    const void* load_scope_{nullptr};
+
+    CollectionCacheKey
+    Key(const std::string& collection_name) const {
```
yhmo: src/impl/utils/cache/CacheContext.h line:32
Low ---- This convenience method is never used; the cache implementations and call sites all construct CollectionCacheKey directly. It therefore adds dead API surface and another key-construction path that can drift, so please remove it or use it consistently.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3824,75 +4069,34 @@ MilvusClientV2Impl::getFlushState(const std::string& db_name, const std::vector<
...
+        context.cache_.endpoint_, context.cache_.database_name_, collection_name, force_update,
+        context.cache_.load_scope_, [this, &context, &collection_name, rpc_timeout_ms](CollectionDescPtr& loaded) {
+            auto request = DescribeCollectionRequest()
+                               .WithDatabaseName(context.cache_.database_name_)
+                               .WithCollectionName(collection_name);
+            DescribeCollectionResponse response;
+            auto status = describeCollection(context, request, response, rpc_timeout_ms);
```
liliu-z: SchemaCache coalesces concurrent first-time loads for the same (endpoint, db, collection) within one client's load_scope onto a single in-flight RPC; the leader's loader captures its own rpc_timeout_ms and joiners block on it via an unbounded condition_variable wait. So when runOptimize feeds a shrinking remaining_rpc_timeout_ms into getCollectionDesc, a concurrent same-client caller with a larger budget is bounded by the leader's shorter deadline (and, conversely, a joiner with a short deadline waits past its own on the untimed CV). Narrow window — only runOptimize passes a non-default per-call timeout. (The reviewer cited SearchIteratorV2Impl.cpp:150, which is unrelated; the real code path is MilvusClientV2Impl::getCollectionDesc / SchemaCache::GetOrLoad.)

yhmo: no need

src/impl/utils/cache/CollectionTsCache.cpp
```diff
@@ -0,0 +1,137 @@
...
+    transfer(CollectionCacheKey::Create(endpoint, old_db_name, old_collection_name),
+             CollectionCacheKey::Create(endpoint, new_db_name, new_collection_name), true);
+}
+
+void
+CollectionTsCache::Copy(const std::string& endpoint, const std::string& source_db_name,
+                        const std::string& source_collection_name, const std::string& target_db_name,
+                        const std::string& target_collection_name) {
+    transfer(CollectionCacheKey::Create(endpoint, source_db_name, source_collection_name),
+             CollectionCacheKey::Create(endpoint, target_db_name, target_collection_name), false);
```
yhmo, line 94: src/impl/utils/cache/CollectionTsCache.cpp line:94
Medium ---- `Copy` only snapshots the timestamp at alias creation. Every later Insert/Upsert/Delete calls `Set` for only the name used by that request, so a completed write through the collection leaves the alias on the old timestamp (and vice versa); a following SESSION read through the other name can miss that write. Please make aliases share/propagate the latest timestamp and add a create-alias → later-DML → SESSION-read regression test.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -535,15 +543,28 @@ MilvusClientV2Impl::DescribeReplicas(const DescribeReplicasRequest& request, Des
...
+        CollectionTsCache::GetInstance().Move(endpoint, source_db, request.CollectionName(), target_db,
+                                              request.NewCollectionName());
+        SchemaCache::GetInstance().Invalidate(endpoint, source_db, request.CollectionName());
+        SchemaCache::GetInstance().Invalidate(endpoint, target_db, request.NewCollectionName());
         return Status::OK();
     };
 
     return connection_.Invoke<proto::milvus::RenameCollectionRequest, proto::common::Status>(
-        pre, &MilvusConnection::RenameCollection);
+        pre, &MilvusConnection::RenameCollection, post);
```
yhmo, line 569: src/impl/MilvusClientV2Impl.cpp line:567
Medium ---- This cache update runs only when `Invoke` reports success. A deadline or UNAVAILABLE response can arrive after Milvus has already committed the rename, leaving both schema keys and the timestamp key in their pre-rename state indefinitely. Please invalidate the affected keys on ambiguous transport outcomes as well; the same issue applies to the other new DDL invalidation hooks.

src/impl/utils/cache/CollectionCacheKey.h
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+ private:
+    static std::string
+    NormalizeEndpoint(const std::string& endpoint) {
+        if (endpoint.empty()) {
+            return endpoint;
+        }
+
+        const auto uri = ParseURI(endpoint);
```
yhmo: src/impl/utils/cache/CollectionCacheKey.h line:48
Medium ---- `ParseURI` can throw from `std::stoi` for a malformed port. After `Connect` returns an error, a caller with an explicit database can reach this normalization from a cache-using API and receive an exception instead of a `Status`. Please make cache-key normalization non-throwing (for example, fall back to the raw endpoint) and cover the failed-connect follow-up path.

src/impl/utils/cache/CollectionTsCache.cpp
yhmo: as design

src/impl/MilvusClientV2Impl.cpp
yhmo: not in this pr's scope

liliu-z:
**src/impl/MilvusClientImpl.cpp:921** `QueryIterator` inherits the base iterator's `WithIDs` primary-key filter but never forwards it into the query request it builds, so the ID list is silently dropped. Callers that pass `WithIDs` expect results restricted to those keys; instead the iterator issues an unfiltered query, returning far more rows than requested and performing a full-collection scan — a correctness bug plus a performance hit. Forward `WithIDs` into the iterator's query expression/request, or reject the combination.

liliu-z:
**src/impl/MilvusClientImpl.cpp:958** Under SESSION consistency the first page sends a nonzero guarantee timestamp `T`, the server replies with `session_ts=0`, and the SDK then overwrites its stored guarantee timestamp with the local wall clock instead of retaining `T`. Later pages must reuse the first page's snapshot to stay consistent, so substituting `now()` yields a different — often future — timestamp, and pages after the first read from an inconsistent snapshot, inconsistently including or dropping rows relative to page one. When the response carries `session_ts==0`, keep the previously used `T` rather than replacing it with wall-clock time.

src/impl/MilvusClientImpl.cpp
```diff
@@ -116,8 +120,14 @@ MilvusClientImpl::CreateCollection(const CollectionSchema& schema, int64_t num_p
...
+    auto post = [&endpoint, &database_name, &schema](const proto::common::Status&) {
+        CollectionTsCache::GetInstance().Invalidate(endpoint, database_name, schema.Name());
+        SchemaCache::GetInstance().Invalidate(endpoint, database_name, schema.Name());
+        return Status::OK();
+    };
+
     return connection_.Invoke<proto::milvus::CreateCollectionRequest, proto::common::Status>(
-        validate, pre, &MilvusConnection::CreateCollection, nullptr);
+        validate, pre, &MilvusConnection::CreateCollection, post);
 }
```
liliu-z, line 134: `Connect` retains a connection built from a malformed URI, and the newly added top-of-function `CurrentDbName("")` path reaches `std::stoi` on an empty/invalid port string; `std::stoi("")` throws `std::invalid_argument`, which is never caught. A malformed URI (e.g. missing port) therefore escapes as an unhandled exception out of the SDK instead of a returned error Status, crashing or surprising the caller. Validate the port before `std::stoi` (or catch and return an error Status), and don't retain a connection created from a malformed URI.

yhmo: no need

PR #560, opened 2026-07-31
Refine cache machinery

yhmo:


src/impl/utils/cache/CollectionTsCache.h
```diff
@@ -0,0 +1,72 @@
...
+
+    void
+    InvalidateDb(const std::string& endpoint, const std::string& db_name);
+
+    // Move the latest timestamp to a renamed collection and remove the old key.
+    void
+    Move(const std::string& endpoint, const std::string& old_db_name, const std::string& old_collection_name,
+         const std::string& new_db_name, const std::string& new_collection_name);
+
+    // Copy the latest timestamp to an alias while retaining the collection key. The target is
```
yhmo, line 52: src/impl/utils/cache/CollectionTsCache.h line:52
Medium ---- Copy only synchronizes the alias at creation/alter time, but later DML calls Set with only the name used by that request. After creating alias for collection, an insert through collection advances only the collection key, so a SESSION query through alias sends the older timestamp and no longer guarantees visibility (and vice versa). The alias and canonical name need shared timestamp ownership or write propagation, with a test that performs DML after alias creation.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1786,9 +1891,46 @@ MilvusClientV2Impl::HybridSearch(const HybridSearchRequest& request, HybridSearc
...
+Status
+MilvusClientV2Impl::query(const std::string& endpoint, const std::string& database_name, const QueryRequest& request,
+                          QueryResponse& response) {
+    auto pre = [this, &endpoint, &database_name, &request](proto::milvus::QueryRequest& rpc_request) {
+        const auto id_count = request.IDs().GetRowCount();
+        if (!request.Filter().empty() && id_count != 0) {
+            return Status{StatusCode::INVALID_ARGUMENT, "Filter and IDs cannot be set at the same time"};
+        }
+
+        if (id_count == 0) {
```
yhmo, line 1908: src/impl/MilvusClientV2Impl.cpp line:1908
Medium ---- id_count == 0 treats an explicitly supplied empty ID array exactly like “IDs were never selected.” For example, QueryRequest().WithIDs({}).WithLimit(10) now takes the ordinary no-filter path and can query arbitrary rows (or get a different server error) instead of producing an empty ID query. Please preserve whether ID mode was selected—e.g. with an explicit flag/optional—and handle an empty selected array as empty input rather than falling through.

src/impl/utils/cache/CollectionTsCache.h
yhmo: not in this pr scope

src/impl/MilvusClientV2Impl.cpp
yhmo: align with pymilvus

PR #564, opened 2026-08-01
Support client-side bloom filter blobs for bloom_match

### LINKED ISSUE #562: Support client-side bloom filter blobs for bloom_match
## Background

`bloom_match(field, {blob})` is a new filter expression that tests approximate set membership against a **client pre-built** Split-Block Bloom Filter. It shipped server-side in milvus-io/milvus#51140 (merged 2026-07-28) and is on `master`.

The point is request size. A membership set of 10M int64 values is roughly 90 MB as a raw `in [...]` list, which blows past the proxy's gRPC receive limit. The same set as a bloom filter is about 14 MB at the default false-positive rate — the client builds the filter, ships the compact blob as an expression template parameter, and the proxy embeds it verbatim after validating the envelope. **The server never rebuilds the filter**, so every SDK must produce byte-identical blobs.

Design doc: [`docs/design-docs/design_docs/20260707-bloom-filter-expression.md`](https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260707-bloom-filter-expression.md)

## What to implement

1. A public helper that turns a membership set (int64 or UTF-8 string) plus a false-positive rate into a blob.
2. A `bytes` branch in the SDK's existing expression-template-value conversion, so the blob travels as protobuf `TemplateValue.bytes_val` rather than being base64-inflated through a string field (the body is not valid UTF-8, so a string round trip corrupts it besides).
3. Unit tests that check the produced blob against the shared golden vectors.

## Wire format

The blob is a 32-byte **MBF1** envelope followed by a Parquet Split-Block Bloom Filter body. All integers little-endian.

| offset | size | field |
|---|---|---|
| 0  | 4 | magic `MBF1` |
| 4  | 2 | version = 1 |
| 6  | 2 | algo = 1 (parquet_sbbf_xxh64) |
| 8  | 8 | n_declared (uint64, informational) |
| 16 | 8 | fpr_declared (float64, informational) |
| 24 | 4 | num_blocks (uint32; body length must equal num_blocks × 32) |
| 28 | 1 | domains bitmask: `0x01` = int64, `0x02` = utf8 |
| 29 | 3 | reserved, must be 0 |
| 32 | … | body |

The body is bit-identical to Arrow C++'s `parquet::BlockSplitBloomFilter` and thus to the parquet-format `BloomFilter.md` spec:

- The body is a power-of-two number of 32-byte blocks; each block is eight little-endian `uint32` words.
- Values are hashed with **XXH64, seed 0**. An integer hashes its **8-byte little-endian two's-complement encoding**; a string hashes its **raw UTF-8 bytes**. (This matches Parquet plain encoding for INT64 / BYTE_ARRAY.)
- Block index is the multiply-shift reduction `((hash >> 32) * num_blocks) >> 32`.
- Within the block, for each word `i` in `0..7`, set bit `1 << ((uint32(hash) * SALT[i]) >> 27)`, where

  ```
  SALT = [0x47b6137b, 0x44974d91, 0x8824ad5b, 0xa2b7289d,
          0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31]
  ```

  Note `uint32(hash)` is the **low** 32 bits, the multiply wraps mod 2³², and the shift is **logical**.

Sizing mirrors Arrow's `BlockSplitBloomFilter::OptimalNumOfBytes`:

```
bits = -8 * n / ln(1 - fpr^(1/8))
bits = clamp(bits, 32*8, 128*1024*1024*8)   # MIN_FILTER_BYTES=32, MAX_FILTER_BYTES=128 MiB
bits = round_up_to_power_of_two(bits)
num_bytes = bits / 8
```

Bounds: `fpr ∈ [0.0001, 0.05]`, default `0.005`. An empty member set is legal — it records `domains = 0` and matches nothing.

**The `domains` bitmask matters.** The two hash domains share one XXH64 output space, so an 8-byte string and the int64 with the same byte image hash identically. Recording which domains were actually inserted is what keeps a probe from aliasing across domains, and it lets the proxy reject a blob built for the wrong field type instead of silently returning fewer rows. Build the blob from the **same value domain as the target field**: integer fields hash int64, VARCHAR fields hash UTF-8.

## Reference implementations

Both are already merged and produce identical bytes:

- **Go** — [`client/sbbf/sbbf.go`](https://github.com/milvus-io/milvus/blob/master/client/sbbf/sbbf.go) (builder + parser) and [`client/milvusclient/bloom_filter.go`](https://github.com/milvus-io/milvus/blob/master/client/milvusclient/bloom_filter.go) (public API + template wiring). This is the cleanest one to port from.
- **Python** — [`pymilvus/client/bloom_filter.py`](https://github.com/milvus-io/pymilvus/blob/master/pymilvus/client/bloom_filter.py) (milvus-io/pymilvus#3690).

## Acceptance criteria

- [ ] Blob is **byte-identical** to the shared golden vectors: [`client/sbbf/testdata/golden_vectors.json`](https://github.com/milvus-io/milvus/blob/master/client/sbbf/testdata/golden_vectors.json) (3 cases: int64-only, string-only, mixed-domain) and [`client/sbbf/testdata/cpp_generated_100_int64.json`](https://github.com/milvus-io/milvus/blob/master/client/sbbf/testdata/cpp_generated_100_int64.json) (100 int64 values generated by Arrow C++ itself). Copy both files into the SDK's test fixtures and compare full hex — this is the real interop check, and it is worth more than any self-consistent round-trip test.
- [ ] `fpr` outside `[0.0001, 0.05]`, NaN, and infinity are rejected.
- [ ] Members of mixed type are rejected by the convenience API (one call, one domain).
- [ ] The blob reaches the server as `TemplateValue.bytes_val`, verified by a test on the template-conversion function.
- [ ] **An incremental builder**, so a membership set that does not fit in memory as a list can be streamed in. All four SDKs that have shipped this provide one (`sbbf.Builder` in Go, `BloomFilterUtils.Builder` in Java, `BloomFilterBuilder` in Node and pymilvus), and it is not a nicety: the filter is a fixed-size bit array sized from `n` and `fpr` alone — never from the values — and insertion is order-independent and idempotent, so members can come from a cursor or a file and be dropped as they go. Measured on pymilvus, a 10M-member build drops from 559 MB to 160 MB peak RSS for integers and from 882 MB to 160 MB for strings, at identical speed and identical bytes. Without it the caller is forced to materialise the whole set, which for 10M members costs far more than the 16 MiB filter itself. Note `n` only drives sizing: inserting more is allowed and simply raises the effective false-positive rate, so a rough estimate suffices.
- [ ] A size-estimate helper (`n`, `fpr` → exact blob size) so callers can check against `proxy.maxBloomFilterSize` (64 MiB default) before building. Bodies are powers of two, so a count just past a tier boundary doubles the blob and a slightly higher `fpr` usually brings it back under.

## Protobuf dependency

`TemplateValue.bytes_val = 6` was added in milvus-io/milvus-proto#632 (commit `7efdf25`). SDKs pinned before that commit must bump milvus-proto and regenerate.

```proto
message TemplateValue {
  oneof val {
    bool bool_val = 1;
    int64 int64_val = 2;
    double float_val = 3;
    string string_val = 4;
    TemplateArrayValue array_val = 5;
    bytes bytes_val = 6;   // <-- needed
  };
}
```

## Status across SDKs

| SDK | status |
|---|---|
| Go | ✅ merged with the server PR (milvus-io/milvus#51140) |
| Python | 🔄 milvus-io/pymilvus#3690 open, awaiting review |
| Java | 📋 milvus-io/milvus-sdk-java#1981 |
| Node | 📋 milvus-io/milvus-sdk-node#597 |
| C++ | **this issue** |
| Rust | 📋 milvus-io/milvus-sdk-rust#131 |

## C++ specifics

**Landing spots**

- `ConvertFilterTemplates` — declared `src/impl/utils/DqlUtils.h:75-77`, defined `src/impl/utils/DqlUtils.cpp:1472-1508`. It dispatches on `nlohmann::json` type predicates, with a validation gate at `:1483` that rejects anything not boolean / integer / float / string / array.
- Callers: `DqlUtils.cpp:1534` (query), `:1623` (search), `:2068` (hybrid search).

**Suggested API shape: carry the blob as `nlohmann::json::binary`, and keep the format invisible to callers.**

The public API carries templates as `std::unordered_map<std::string, nlohmann::json>` (`src/include/milvus/types/QueryArguments.h:108-115`, `src/include/milvus/types/SearchRequestBase.h:65,70-77`). That map does not need to change: the vendored nlohmann is **3.11.3**, which has a first-class binary value type (`json::binary_t`, a `std::vector<uint8_t>` subclass, used for the CBOR / MessagePack binary subtypes). So a blob has a perfectly natural `nlohmann::json` representation, and the wiring is genuinely just:

- `ConvertFilterTemplates` gets one more branch, `else if (temp.is_binary()) value.set_bytes_val(...)` over `temp.get_binary()`;
- the validation gate at `DqlUtils.cpp:1483` gets `is_binary()` added to the allowed set.

No public signature changes, no new overload of the templates map. (The one caveat is that `json::binary_t` throws on `dump()` to text JSON, since binary has no JSON text representation — irrelevant here, because template values are converted to protobuf, never dumped.)

On top of that, the MBF1 format should not be something users think about. Rather than handing back raw bytes for the caller to wrap, return the ready-to-use value:

```cpp
auto bf = milvus::BloomFilterTemplate(user_ids);            // nlohmann::json, binary-wrapped
// or milvus::BloomFilterTemplate(user_ids, /*fpr=*/0.005);

query_arguments.AddFilterTemplate("bf", bf);                // existing map, unchanged
query_arguments.SetFilter("bloom_match(user_id, {bf})");
```

Callers never see `MBF1`, never see a `std::vector<uint8_t>`, and never pick an encoding.

**Do keep the value reusable, though.** Building a filter over millions of members is the expensive part, and the entire point is to amortize it across many queries — so the helper must return something the caller can hold and pass repeatedly, not a hidden per-query build inside `SetFilter`. If the deep copy of a large `nlohmann::json` per query turns out to matter (a 64 MiB body is the configured maximum), an opaque handle holding a `std::shared_ptr<const std::vector<uint8_t>>` plus one overload of the template setter would avoid it — but that is an optimization to reach for only if measurements ask for it, not the starting design.

**Proto needs a bump.** `cmake/MilvusProtoGen.cmake:19` pins:

```cmake
set(PROTO_COMMIT e378dbfa23a2abc16378e8c7e525754c48d82e4b)
```

At `e378dbf` (2026-06-15) `TemplateValue` still has only fields 1–5. Bump `PROTO_COMMIT` to `7efdf25099dd41c8215f47b7a6525aa0e22eea32` or later. protoc runs at build time from the fetched sources, so a clean rebuild picks it up — there is no checked-in generated code to regenerate.

**No XXH64 available, not even transitively.** `conanfile.py:49-58` requires only `grpc/1.65.0`, `protobuf/5.27.0`, `abseil/20240116.2` (plus `gtest/1.12.1` under `with_tests=True`); grpc's transitive closure (abseil, c-ares, openssl, re2, zlib) exposes no xxhash header. There is no Arrow/parquet here, so `parquet::BlockSplitBloomFilter` — which implements exactly this algorithm — is not available to borrow.

Two options: add `xxhash/0.8.2` to `conanfile.py`, which **also requires a fallback in `cmake/ThirdPartyPackages.cmake`** because the `make build-no-conan-*` path exists; or vendor the single-header `xxhash.h` under `thirdparty/`, which matches this repo's stated practice (`thirdparty/thirdparty.md`: "sub-module projects or header-only projects") and keeps both build paths working. The vendored header is the lower-friction choice.

Note the server already has a reference C++ implementation of the same envelope and probe logic in [`internal/core/src/common/BloomFilterEnvelope.h`](https://github.com/milvus-io/milvus/blob/master/internal/core/src/common/BloomFilterEnvelope.h) — useful to read, though it is the read side rather than the builder.

**Build & test**

```bash
bash scripts/install_deps.sh
make          # build
make test     # builds and runs ./test/testing-ut and ./test/testing-it
make lint
```

Unit tests live in `test/ut/` (no server needed) → `testing-ut`; googletest 1.12.1 + gmock. `test/ut/utils/TestDqlUtils.cpp` is the natural home for the template-conversion test, and a new `test/ut/` file for the builder conformance vectors. Note there is no `make ut` and no `make code-format` target, and `scripts/build.sh` reformats sources in place unless `-f` is passed.

**DCO is enforced** via mergify (`git commit -s`). There is no `CONTRIBUTING.md` in this repo; it defers to the main milvus guide.

xiaofan-luan:
Closes #562.

Adds the C++ side of `bloom_match(field, {blob})`, the approximate-membership filter expression that shipped server-side in milvus-io/milvus#51140. A membership set of 10M int64 values is roughly 90 MB as a raw `in [...]` list and blows past the proxy's gRPC receive limit; the same set is about 14 MB as a bloom filter. The client builds the filter, ships the blob as an expression template parameter, and the proxy embeds it verbatim after validating the envelope — **the server never rebuilds it**, so the bytes have to match the other SDKs exactly.

## Usage

The MBF1 format stays invisible to callers — no byte vector, no encoding choice:

```cpp
nlohmann::json bf;
auto status = milvus::BloomFilterTemplate(user_ids, milvus::BloomFilterDefaultFPR, bf);

arguments.AddFilterTemplate("bf", bf);
arguments.SetFilter("bloom_match(user_id, {bf})");
```

For a set too large to materialise as a vector, `BloomFilterBuilder` streams:

```cpp
milvus::BloomFilterBuilder builder(10000000);
while (cursor.Next()) {
    builder.AddInt64(cursor.Id());
}
arguments.AddFilterTemplate("bf", builder.BuildTemplate());
```

The filter is a fixed-size bit array sized from `n` and the false-positive rate alone — never from the values — and insertion is order-independent and idempotent, so members can come from a cursor and be dropped as they go. `n` only drives sizing; adding more is allowed and simply raises the effective rate.

## No public API change

The blob rides as `nlohmann::json::binary`. The vendored nlohmann is 3.11.3, which has a first-class binary value type, so the `unordered_map<std::string, nlohmann::json>` that `AddFilterTemplate` already takes carries it unchanged — `ConvertFilterTemplates` gains an `is_binary()` branch mapping to `TemplateValue.bytes_val`, and nothing else moves. A JSON string would not work: the filter body is not valid UTF-8, so it would be corrupted as well as base64-inflated.

## XXH64 is implemented here, not added as a dependency

The conan closure carries no xxhash — grpc's transitive set is abseil, c-ares, openssl, re2 and zlib — and there is no Arrow here to borrow `parquet::BlockSplitBloomFilter` from. It is about a hundred lines and the reference implementations in Go and Python did the same.

## Proto pin

`cmake/MilvusProtoGen.cmake` moves from `e378dbf` to `7efdf25` for `TemplateValue.bytes_val` (milvus-io/milvus-proto#632). Same line, fast-forward, purely additive — the only deletions in `schema.proto` across the range are same-line comment edits.

## Testing, and what I could not verify locally

7 tests in `test/ut/types/TestBloomFilter.cpp`. The load-bearing one compares the **full hex** of the produced blob against the golden vectors shared with the Go SDK and the segcore C++ unit tests — three cases including a mixed-domain one, plus `cpp_generated_100_int64`, which Arrow C++ generated itself. That is the real interop check; a self-consistent round-trip would prove nothing about matching the server. Also covered: envelope header layout, domain bitmask recording, empty sets, batch-vs-single insertion equivalence, the size estimator against actual output including the 128 MiB clamp, and rejection of out-of-range rates through both the `Status` and the throwing paths.

**Those 7 tests pass**, compiled against this repo's own sources and its conan gtest. `clang-format` (17) is clean on all four files.

I could **not** get `make test` to run end to end on my machine, so the `ConvertFilterTemplates` change is compile-verified only by review, not by a build. Three separate local toolchain problems, in order:

1. `scripts/build.sh` pins `BUILD_CPPSTD=17`, but ConanCenter's current abseil publishes at cppstd 20, so conan falls back to a "compatible" binary and protobuf's cppstd-match validation rejects the combination. Forcing `--build="abseil/*" --build="protobuf/*"` gets past it. **This one looks like a repo issue rather than mine** — anyone building fresh today should hit it. Happy to send it as a separate one-line PR if you agree.
2. A stale `CPPFLAGS` on my shell pointed at a directory that does not exist on Apple Silicon, which broke openssl's build. My problem, not yours.
3. Command Line Tools' `clang`, invoked by absolute path as conan does, does not auto-detect the macOS SDK the way the `/usr/bin` shim does; `SDKROOT` fixes it. Also mine.

With all three worked around, grpc built to 100% and then the generated `.pb.h` failed with a protobuf gencode/runtime version mismatch (`ClassDataFull` is protobuf ≥5.29, the runtime is 5.27). I stopped there rather than keep digging, since CI builds on Linux and will settle it faster than I can. If CI goes red on something I should have caught, tell me and I will fix it rather than hand it back.

## Performance note

The block read-modify-write goes through `memcpy` on little-endian hosts so the compiler can vectorise it — at `-O2` the eight-word loop becomes one `ldp`, two `orr.16b` and one `stp` on arm64. My first version used the byte-wise accessors everywhere for portability and was about 4x slower on a 10M-member build (0.150s → 0.062s after the change). The byte-wise path is still there for big-endian hosts, where correctness matters more than the vectorisation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HWMAbqV46tWrVPhWRDedpG


src/include/milvus/types/BloomFilter.h
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * @brief Stamp the envelope header and return the blob.
+     */
+    std::vector<uint8_t>
+    Build() const;
+
+    /**
+     * @brief Same blob as Build(), wrapped as a JSON binary value so it can be handed to
+     * QueryArguments::AddFilterTemplate() / SearchArguments::AddFilterTemplate() directly.
```
yhmo, line 138: src/include/milvus/types/BloomFilter.h line:138
High ---- BuildTemplate() is not actually accepted by the advertised query/search APIs. QueryArguments::AddFilterTemplate() and SearchRequestBase::AddFilterTemplate() both call IsValidTemplate(), which currently accepts only boolean/number/string; V2 SearchRequest::AddFilterTemplate() delegates to the latter and ignores its error. Thus this value is rejected (or silently dropped) before ConvertFilterTemplates() can set bytes_val. Please extend the shared validation/docs for binary and add a public-path test that asserts the blob reaches TemplateValue.bytes_val.

test/ut/types/TestBloomFilter.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    for (const auto& one : fixture["cases"]) {
+        const auto name = one["name"].get<std::string>();
+        milvus::BloomFilterBuilder builder(one["int_values"].size() + one["string_values"].size(),
+                                           one["fpr"].get<double>());
+        for (const auto& value : one["int_values"]) {
+            builder.AddInt64(std::stoll(value.get<std::string>()));
+        }
+        for (const auto& value : one["string_values"]) {
+            builder.AddString(value.get<std::string>());
```
yhmo, line 72: test/ut/types/TestBloomFilter.cpp line:69
Low ---- All current string fixtures are shorter than 32 bytes (the maximum is 11), so they never exercise the custom XXH64() stripe/merge branch used for realistic long VARCHAR values. Please add at least one 32-byte-or-longer string with a cross-SDK/Arrow golden blob so this wire-contract branch is covered independently.

src/include/milvus/types/BloomFilter.h
```diff
@@ -0,0 +1,184 @@
...
+ */
+constexpr double BloomFilterDefaultFPR = 0.005;
+
+/**
+ * @brief Builds the blob a bloom_match(field, {blob}) filter expression consumes.
+ *
+ * Building the filter on the client and shipping the compact blob lets large membership sets
+ * pass the proxy gRPC receive limit, which a raw value list would exceed for the same set.
+ * The proxy embeds the blob verbatim after validating the envelope and never rebuilds the
+ * filter, so the bytes are a wire contract: they are identical to what the Go, Python, Java
```
yhmo: src/include/milvus/types/BloomFilter.h line:55
Low ---- This states Java and Node already produce compatible blobs, but their linked implementation issues (#1981 and #597) are still open. Please phrase this as the cross-SDK wire-format target (or list only shipped implementations) so users do not infer that Java/Node interoperability is currently available.

xiaofan-luan:
Thanks @yhmo — the High one was a real break and I had missed it. Pushed `3356303`.

**High — `BuildTemplate()` rejected before it reaches the converter.** Confirmed exactly as you describe: `IsValidTemplate()` returns `is_boolean() || is_number() || is_string()`, so `QueryArguments::AddFilterTemplate` and `SearchRequestBase::AddFilterTemplate` both return `INVALID_ARGUMENT` and never store the value. The conversion I added was correct and unreachable. `IsValidTemplate` now accepts `is_binary()`, and both error strings mention it.

The reason this got past me is worth stating plainly: my test called the conversion directly instead of going through `AddFilterTemplate`, so it verified the half I wrote and skipped the half that gates it. There is now a test on the public path — `AddFilterTemplate` on both `QueryArguments` and `SearchArguments`, then `ConvertFilterTemplates`, asserting the blob arrives as `bytes_val` rather than `string_val`.

On the V2 `SearchRequest::AddFilterTemplate` dropping the returned `Status`: also confirmed (`SearchRequest.cpp:93`). I left it alone — it returns `*this` for chaining, so it structurally cannot propagate a `Status`, and changing that shape is a wider API decision than this PR should make. It is now less dangerous in practice since binary is accepted, but a caller passing a genuinely invalid template still gets silence. Happy to take it as a follow-up if you want it fixed.

**Low — no string fixture reaches the stripe branch.** Correct, the shared vectors top out at 11 bytes, so XXH64's ≥32-byte accumulator and merge rounds had no coverage here at all. Added the 31/32/33-byte boundary vectors, which are the same blobs the Go and Python SDKs check themselves against — 31 stays in the tail path, 32 and 33 enter the stripe loop. They pass, so the branch is now pinned against a cross-SDK golden rather than only against itself.

**Low — the header overclaimed Java/Node.** Fixed; it now says Go and Python have shipped and the others target the same format and are in flight.

---

@Copilot's two, having checked both:

**`OptimalNumOfBytes()` truncating `m` before the power-of-two round-up** — real observation, deliberately kept. Arrow's `BlockSplitBloomFilter::OptimalNumOfBytes` truncates the same way, and so do the Go and Python SDKs. For `m` just above a power of two the filter does come out one tier smaller than the ideal size, i.e. slightly above the requested FPR. Changing it here would make this SDK size filters differently from every other implementation for the same `(n, fpr)` — and since the server validates `num_blocks` against the body it would be a visible interop divergence, not just a local difference. The `cpp_generated_100_int64` fixture, which Arrow generated, is reproduced byte-for-byte with the truncation in place.

**`binary_t::data()` returning `uint8_t*` against the protobuf setter** — not an issue. The generated setter is `set_bytes_val(const void* value, size_t size)`, which `uint8_t*` converts to implicitly; this compiles clean on Ubuntu 20.04, Fedora 39, macOS 15 and Windows in CI.

One thing I cannot verify locally and am flagging rather than claiming: the new public-path test needs the generated protobuf headers, and my machine still cannot complete the conan dependency graph (details in the PR description). The stripe-boundary test I did run locally — it passes. CI will settle the other one.

PR #571, opened 2026-08-04
Make connection setup transactional and handle server rejection

yhmo:


test/it/v1/TestConnection.cpp
```diff
@@ -67,6 +88,98 @@ TEST_F(UnconnectMilvusMockedTest, ConnectTimeout) {
...
+    EXPECT_CALL(service_, Connect(_, _, _))
+        .WillOnce([](::grpc::ServerContext*, const ConnectRequest*, ConnectResponse* response) {
+            response->mutable_status()->set_code(1);
+            response->mutable_status()->set_reason("server rejected reconnection");
+            return ::grpc::Status{};
+        });
+    status = client_->Connect(connect_param);
+    EXPECT_EQ(status.Code(), StatusCode::SERVER_FAILED);
+
+    EXPECT_CALL(service_, HasCollection(_, Property(&HasCollectionRequest::collection_name, "collection"), _))
```
yhmo, line 113: test/it/v1/TestConnection.cpp line:105
Medium ---- This assertion does not prove the existing connection was preserved: both connects use the same endpoint and metadata, so a rejected candidate that was mistakenly published would still satisfy HasCollection. Please make the attempted reconnect distinguishable—for example, connect initially with token A, retry with token B, then assert HasCollection receives token A—so this test protects the transactional guarantee.

PR #576, opened 2026-08-10
Avoid version conflict with system-installed protoc

yhmo:


cmake/MilvusProtoGen.cmake
```diff
@@ -65,7 +65,57 @@ if (NOT TARGET gRPC::grpc_cpp_plugin)
     message(FATAL_ERROR "gRPC::grpc_cpp_plugin target not found. Please provide gRPC via find_package(gRPC CONFIG REQUIRED).")
 endif()
 
-set(Protobuf_PROTOC_EXECUTABLE $<TARGET_FILE:protobuf::protoc>)
+function(_milvus_get_conan_protoc out_var)
+    set(_protobuf_include_dirs ${protobuf_INCLUDE_DIRS} ${Protobuf_INCLUDE_DIRS})
```
liliu-z: The candidate list is seeded from the loose `protobuf_INCLUDE_DIRS`/`Protobuf_INCLUDE_DIRS` variables before the include dirs of the target actually being linked, and the first directory whose parent holds a `bin/protoc` wins with no check that it belongs to the Conan package. If either variable was left pointing at a system prefix by an earlier module-mode `find_package(Protobuf)` in the dependency chain, `/usr/include` resolves to `/usr/bin/protoc`, which is then FORCE-pinned into the cache and onto `protobuf::protoc` — reintroducing exactly the system/Conan version mismatch this PR exists to prevent. Query `protobuf::libprotobuf`'s INTERFACE_INCLUDE_DIRECTORIES first, or reject candidates outside the protobuf package root.

cmake/MilvusProtoGen.cmake
```diff
@@ -65,7 +65,57 @@ if (NOT TARGET gRPC::grpc_cpp_plugin)
...
+        if (EXISTS "${_exact_protoc}")
+            set(${out_var} "${_exact_protoc}" PARENT_SCOPE)
+            return()
+        endif()
+
+        file(GLOB _versioned_protoc
+            LIST_DIRECTORIES FALSE
+            "${_protobuf_root}/bin/protoc-*${CMAKE_EXECUTABLE_SUFFIX}")
+        if (_versioned_protoc)
+            list(SORT _versioned_protoc)
```
liliu-z, line 110: The versioned-protoc fallback sorts the glob results alphabetically and takes index 0, which is not the newest binary — with `protoc-3.21.0` and `protoc-4.25.0` present it selects 3.21.0. For a change whose purpose is picking a protoc that matches the linked protobuf, silently choosing an arbitrary version defeats the intent; sort descending (or match against the protobuf version) and log which binary was chosen.

cmake/MilvusProtoGen.cmake
```diff
@@ -65,7 +65,67 @@ if (NOT TARGET gRPC::grpc_cpp_plugin)
...
+            list(GET _versioned_protoc 0 _protoc)
+            set(${out_var} "${_protoc}" PARENT_SCOPE)
+            return()
+        endif()
+    endforeach()
+endfunction()
+
+if (BUILD_FROM_CONAN)
```
yhmo: cmake/MilvusProtoGen.cmake line:118
Medium ---- This unconditional Conan override breaks cross-compilation. ConanCenter's protobuf build module deliberately uses a build-context protoc from PATH when CMAKE_CROSSCOMPILING, because the executable under the host protobuf package is built for the target platform (for example Android/ARM) and cannot run on the build machine. _milvus_get_conan_protoc instead derives the target package's bin/protoc from libprotobuf include paths and then FORCE-overwrites PROTOC_PROGRAM and protobuf::protoc, so proto generation can fail with an exec-format/not-runnable error. Please preserve the provider-selected compiler for cross builds (for example, skip this override when CMAKE_CROSSCOMPILING) and apply the version repair only to native Conan builds.

cmake/MilvusProtoGen.cmake
```diff
@@ -65,7 +65,67 @@ if (NOT TARGET gRPC::grpc_cpp_plugin)
...
+if (BUILD_FROM_CONAN AND NOT CMAKE_CROSSCOMPILING)
+    _milvus_get_conan_protoc(_MILVUS_CONAN_PROTOC)
+endif()
+
+if (_MILVUS_CONAN_PROTOC)
+    set(Protobuf_PROTOC_EXECUTABLE "${_MILVUS_CONAN_PROTOC}" CACHE FILEPATH "The protoc compiler" FORCE)
+    set(PROTOC_PROGRAM "${_MILVUS_CONAN_PROTOC}" CACHE FILEPATH "The protoc compiler" FORCE)
+    set_property(TARGET protobuf::protoc PROPERTY IMPORTED_LOCATION "${_MILVUS_CONAN_PROTOC}")
+else()
+    set(Protobuf_PROTOC_EXECUTABLE $<TARGET_FILE:protobuf::protoc>)
```
liliu-z, line 137: When BUILD_FROM_CONAN is on but _milvus_get_conan_protoc returns nothing, the else branch silently restores the pre-PR `$<TARGET_FILE:protobuf::protoc>`, which under Conan resolves to whatever `find_program(PROTOC_PROGRAM NAMES protoc)` picked up — i.e. the system protoc this PR exists to avoid. Nothing warns about that, and the `message(STATUS)` two lines below prints the unexpanded generator expression, so the configure log gives no way to tell which protoc will actually run. A `message(WARNING)` in the else branch when BUILD_FROM_CONAN is set would leave a trail; a hard FATAL_ERROR would be wrong here, since the fallback is still correct for package layouts the derivation doesn't recognize.

cmake/MilvusProtoGen.cmake
```diff
@@ -65,7 +65,67 @@ if (NOT TARGET gRPC::grpc_cpp_plugin)
...
+        endif()
+    endforeach()
+endfunction()
+
+if (BUILD_FROM_CONAN AND NOT CMAKE_CROSSCOMPILING)
+    _milvus_get_conan_protoc(_MILVUS_CONAN_PROTOC)
+endif()
+
+if (_MILVUS_CONAN_PROTOC)
+    set(Protobuf_PROTOC_EXECUTABLE "${_MILVUS_CONAN_PROTOC}" CACHE FILEPATH "The protoc compiler" FORCE)
```
liliu-z, line 123: `CACHE ... FORCE` overwrites an explicit `-DProtobuf_PROTOC_EXECUTABLE=` given on the command line. Conan's protobuf module deliberately sets that cache entry without FORCE so users can supply an external protoc, and this undoes it. The blast radius is small — line 118 already excludes cross-compiling, the main legitimate reason to point at a different protoc — so guarding this one assignment on the cache entry not already being user-set would close the gap without weakening the fix.

yhmo: no need

PR #580, opened 2026-08-12
Add tutorial

yhmo:
- Adds eight tutorials covering quickstart, collections, schemas, indexes, DML, DQL, databases,
    and RBAC.

- Provides independent Conan/CMake projects with detailed READMEs, expected output,
  troubleshooting, and instructional examples.

- Adds root-level commands to build and run tutorials, including make tutorials and make run-
  tutorial <name>.

- Integrates tutorial builds into Ubuntu CI and excludes tutorials from coverage.
- Updates root documentation and development requirements.
- Enables configurable PCH support and raises the minimum CMake version to 3.16.
- Includes related internal naming, formatting, and build-system cleanup.


tutorial/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```bash
+MILVUS_URI="https://your-milvus-endpoint" \
+MILVUS_TOKEN="your-token" \
+make -C tutorial/1_quickstart run
+```
+
+The source-linked examples used for SDK development remain under [`examples/`](../examples/README.md).
+
+The SDK package defaults to `milvus-sdk-cpp/3.0.2@milvus/dev`. Override it when needed:
```
yhmo: tutorial/README.md line:36
Medium ---- This default recipe is not available from ConanCenter; it is hosted on the Milvus `default-conan-local2` Artifactory remote. On a fresh Conan 2 installation, the documented first `make` therefore fails before configuration, and the troubleshooting advice to change only version/user/channel cannot fix it. Please document (or automate) adding the package remote before this command.

tutorial/6_dql/src/main.cpp
```diff
@@ -0,0 +1,246 @@
...
+    return {b, b + .1f, b + .2f, b + .3f};
+}
+std::map<uint32_t, float>
+Sparse(int id) {
+    return {{static_cast<uint32_t>(id % 8), 1.0f}, {static_cast<uint32_t>((id + 3) % 8), .5f}};
+}
+void
+PrintRows(const milvus::QueryResults& results) {
+    milvus::EntityRows rows;
+    if (results.OutputRows(rows).IsOk()) {
```
yhmo: tutorial/6_dql/src/main.cpp line:34
Low ---- `OutputRows` can fail while converting an otherwise successful response, but this helper silently suppresses that status (and `PrintSearch`/the search-iterator path do the same). A tutorial run can consequently report `Query succeeded` or `Search succeeded` and print no advertised rows with no diagnostic; please propagate or print the conversion error.

tutorial/1_quickstart/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.14)
```
yhmo: tutorial/1_quickstart/CMakeLists.txt line:1
Medium ---- None of the new standalone tutorial projects is configured or compiled by the repository build: root CMake only adds `examples/`, and the lint/format discovery also excludes `tutorial/`. That leaves all eight advertised copy-paste projects unprotected from package, toolchain, include, and API drift (the missing direct includes fixed in this PR are an example). Please add a CI/build target that at least configures and compiles every tutorial.

src/include/milvus/types/BloomFilter.h
```diff
@@ -88,7 +88,7 @@ class MILVUS_SDK_API BloomFilterBuilder {
      * Note: this constructor throws std::runtime_error when @a fpr is NaN or out of range.
      * Use EstimateBloomFilterSize() first if the rate comes from untrusted input.
      */
-    BloomFilterBuilder(uint64_t n, double fpr = BloomFilterDefaultFPR);
+    explicit BloomFilterBuilder(uint64_t n, double fpr = BloomFilterDefaultFPR);
```
yhmo: src/include/milvus/types/BloomFilter.h line:91
Medium ---- Adding `explicit` changes the source contract of this public constructor: code such as `BloomFilterBuilder builder = 100;`, which was valid with v3.0.2, now fails to compile even though this PR is unrelated to the bloom-filter API. Please keep the constructor implicit here (or make this compatibility break an intentional, separately documented API change).

tutorial/7_database/src/main.cpp
```diff
@@ -0,0 +1,144 @@
...
+    std::cerr << "Failed to " << op << ": " << s.Message() << std::endl;
+    return false;
+}
+void
+PrintDatabases(milvus::MilvusClientV2Ptr& client, const std::string& heading) {
+    milvus::ListDatabasesResponse response;
+    // ListDatabases returns every database name visible to the authenticated user.
+    std::cout << "Calling ListDatabases..." << std::endl;
+    auto status = client->ListDatabases(milvus::ListDatabasesRequest(), response);
+    if (!Ok(status, "list databases")) {
```
yhmo, line 27: tutorial/7_database/src/main.cpp line:27
Low ---- `PrintDatabases` reports the RPC error and returns `void`, so either the initial or final `ListDatabases` call can fail while `main` continues and ultimately exits with success. Since listing databases is an advertised tutorial step (and the final call is the cleanup verification), please return a status/bool and propagate the failure to `main`.

tutorial/7_database/src/main.cpp
```diff
@@ -0,0 +1,144 @@
...
+    if (!Ok(status, "describe database properties")) {
+        return 1;
+    }
+    std::cout << "DescribeDatabase succeeded." << std::endl;
+    std::cout << "database.replica.number: " << described.Desc().Properties().at("database.replica.number")
+              << std::endl;
+
+    // DropDatabaseProperties removes the specified key while preserving other properties.
+    std::cout << "Calling DropDatabaseProperties..." << std::endl;
+    status = client->DropDatabaseProperties(
```
yhmo, line 112: tutorial/7_database/src/main.cpp line:105
Low ---- The README says this tutorial removes `database.replica.number` and verifies the property change, but after `DropDatabaseProperties` the code never calls `DescribeDatabase` again or checks that the key is absent. Please add that readback/assertion, or narrow the README claim so the tutorial does not advertise verification it does not perform.

CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ set_option_category("Build")
 define_option(MILVUS_BUILD_TEST "Build with testing" OFF)
 define_option(MILVUS_BUILD_COVERAGE "Build with coverage" OFF)
 define_option(MILVUS_BUILD_EXAMPLES "Build examples" ON)
+define_option(MILVUS_ENABLE_PCH "Use precompiled headers when building the SDK" ON)
```
yhmo, line 65: CMakeLists.txt line:63
Medium ---- Enabling `target_precompile_headers()` by default introduces a configure-time dependency on CMake 3.16+, but this project and every new tutorial still advertise CMake 3.14 support (`cmake_minimum_required(VERSION 3.14)`). On CMake 3.14/3.15, the first SDK configure now fails with an unknown command before any target can be generated. Please either guard this option with a CMake-version check (and keep it off on older versions) or raise the documented minimum consistently.

tutorial/1_quickstart/src/main.cpp
```diff
@@ -0,0 +1,118 @@
...
+    std::cout << "Calling Connect..." << std::endl;
+    auto status = client->Connect(
+        milvus::ConnectParam{Env("MILVUS_URI", "http://localhost:19530"), Env("MILVUS_TOKEN", "root:Milvus")});
+    if (!Ok(status, "connect")) {
+        return 1;
+    }
+    std::cout << "Connect succeeded." << std::endl;
+    const std::string collection = "CPP_TUTORIAL_QUICKSTART";
+    std::cout << "Calling DropCollection for stale data..." << std::endl;
+    client->DropCollection(milvus::DropCollectionRequest().WithCollectionName(collection));
```
yhmo: tutorial/1_quickstart/src/main.cpp line:38
Low ---- This stale-data DropCollection call ignores its Status but the next line reports cleanup as completed. If the fixed collection cannot be dropped (for example because of a transient RPC or insufficient permission), the tutorial continues into CreateCollection and fails with a misleading AlreadyExists error; please check this cleanup result (and the same pre-cleanup calls in tutorials 2–8) before proceeding.

PR #582, opened 2026-08-17
Support Text datatype

yhmo:
Support `DataType::TEXT` in the C++ SDK: schema create/describe, row- and column-based insert/upsert (wire type derived from the schema), query/search decoding, default values, and the `TextFieldData` alias. Also aligned with pymilvus on array element types (TEXT arrays supported, GEOMETRY/TIMESTAMPTZ arrays rejected).

Note: the workflow edits (step `timeout-minutes`, `timeout` wrappers, and the apt `99network-timeouts` config) are unrelated to TEXT support; they stabilize CI for the new Docker-based system tests that this PR adds. The macOS `Prepare` step drops the `timeout` wrapper because GNU `timeout` is unavailable on macOS.


src/include/milvus/types/FieldData.h
```diff
@@ -371,6 +373,7 @@ using FloatFieldData = FieldData<float, DataType::FLOAT>;
 using DoubleFieldData = FieldData<double, DataType::DOUBLE>;
 using VarCharFieldData = FieldData<std::string, DataType::VARCHAR>;
 using GeometryFieldData = VarCharFieldData;     // geometry field data is passed as string
+using TextFieldData = VarCharFieldData;         // text field data is passed as string
```
yhmo, line 376: src/include/milvus/types/FieldData.h line:376
High ---- Column-based insert/upsert of TEXT fields always fails; the column type never matches a TEXT schema field.
Because TextFieldData is an alias of VarCharFieldData (Type() returns DataType::VARCHAR), CheckInsertInput (src/impl/utils/DmlUtils.cpp:140) rejects every TEXT column with DATA_UNMATCH_SCHEMA ("Field data type mismatch for field: ..."); ArrayTextFieldData fails the same element-type check for array-of-text fields (DmlUtils.cpp:142). Even if that validation were relaxed, CreateProtoFieldData (DmlUtils.cpp:778) would still serialize the proto field type as VarChar instead of Text. The IT mock builder added in this PR (test/it/mocks/Utils.cpp:149) constructs TextFieldData for TEXT schema fields, confirming the column API is meant to work, yet no test exercises it. Please derive the wire type from the collection schema in the column path (as CheckAndSetRowData does) or accept VarChar-typed string columns for TEXT fields, and add a column-based insert/upsert test.

src/include/milvus/types/FieldData.h
```diff
@@ -371,6 +373,7 @@ using FloatFieldData = FieldData<float, DataType::FLOAT>;
...
 using TimestamptzFieldData = VarCharFieldData;  // timestamptz field data is passed as string
```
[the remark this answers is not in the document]
yhmo: as designed

test/it/mocks/Utils.cpp
```diff
@@ -146,6 +146,16 @@ BuildFieldsData(const milvus::CollectionSchema& schema, std::vector<milvus::Fiel
                 fields_data.emplace_back(std::move(ptr));
                 break;
             }
+            case milvus::DataType::TEXT: {
```
yhmo, line 159: test/it/mocks/Utils.cpp line:149
Medium ---- The TEXT and array-of-TEXT branches added to BuildFieldsData are never exercised: no IT test in this PR creates a collection with a DataType::TEXT (or array-of-TEXT) field, so these branches are dead code and the TEXT datatype has no end-to-end mocked-server coverage. Unit tests cover the conversion helpers well, but the full client path (CheckInsertInput relaxation -> schema-derived wire type -> mocked server insert) is untested. Consider adding an IT test (e.g. in test/it/v2) that inserts/upserts a TEXT column and asserts the wire type is proto Text + string_data, or remove the unused mock branches until such a test exists.

src/impl/utils/DmlUtils.cpp
```diff
@@ -901,11 +941,16 @@ CreateProtoFieldData(const FieldDataSchema& data_schema, proto::schema::FieldDat
...
             break;
         case DataType::VARCHAR:
-        case DataType::GEOMETRY:
+        case DataType::TEXT:
         case DataType::TIMESTAMPTZ:
             scalar->set_allocated_string_data(
                 CreateProtoScalars<VarCharFieldData, proto::schema::StringArray>(field, field_data, nullable_default));
             break;
+        case DataType::GEOMETRY:
+            scalar->set_allocated_geometry_wkt_data(
```
yhmo, line 970: src/impl/utils/DmlUtils.cpp line:950
Low ---- This PR also changes the column-based wire format for existing GEOMETRY fields (scalar here, and per-element in CreateProtoArrayField at line 714): previously a GEOMETRY column was serialized as string_data, now as geometry_wkt_data. This matches the server expectation (checkGeometryFieldData reads GetGeometryWktData) and the row-based path, and column-based GEOMETRY insert was previously rejected by CheckInsertInput (VARCHAR-typed GeometryFieldData vs GEOMETRY schema), so this is effectively an enabling fix rather than a regression. Worth confirming this GEOMETRY behavior change is intentional and covered, since it is bundled into a "Support Text datatype" PR.

src/impl/utils/TypeUtils.cpp
```diff
@@ -600,9 +605,11 @@ CheckDefaultValue(const FieldSchema& schema) {
             break;
         case DataType::VARCHAR:
         case DataType::GEOMETRY:
+        case DataType::TEXT:
         case DataType::TIMESTAMPTZ:
             if (!val.is_string()) {
-                return {StatusCode::INVALID_ARGUMENT, "Field type is VARCHAR but default value is not string"};
+                return {StatusCode::INVALID_ARGUMENT,
+                        "Field type is " + std::to_string(type) + " but default value is not string"};
```
yhmo, line 612: src/impl/utils/TypeUtils.cpp line:612
Low ---- The default-value validation error now prints the raw numeric enum value (e.g. "Field type is 25 ...") via std::to_string(type). Neighboring messages use readable names, and this file already provides milvus::to_string(DataType) (extended with TEXT in this PR), so using it here would yield "Field type is TEXT but default value is not string" and stay consistent with the surrounding messages.

src/impl/utils/DmlUtils.cpp
```diff
@@ -774,7 +811,12 @@ CreateProtoFieldData(const FieldDataSchema& data_schema, proto::schema::FieldDat
     bool nullable_default = schema ? (schema->IsNullable() || !schema->DefaultValue().is_null()) : false;
 
     auto scalar = field_data.mutable_scalars();
-    const auto field_type = field.Type();
+    auto field_type = field.Type();
+    // derive the wire type from the schema for string-backed fields: a VARCHAR column may target
+    // a VARCHAR/GEOMETRY/TEXT/TIMESTAMPTZ schema field, so the proto type must follow the schema
+    if (schema != nullptr && field_type == DataType::VARCHAR && IsStringBackedType(schema->FieldDataType())) {
```
yhmo: src/impl/utils/DmlUtils.cpp line:817
Low ---- The wire-type derivation here (and the element-type variant in CreateProtoArrayField) only normalizes columns whose Type() is VARCHAR. The new CheckInsertInput relaxation accepts any string-backed column/schema combination, so a column built from the raw template (e.g. FieldData<std::string, DataType::GEOMETRY> for a TEXT schema field, or the reverse) passes validation but keeps its own type and can serialize the wrong scalar container (geometry_wkt_data vs string_data), which the server then rejects. Consider deriving from the schema whenever IsStringBackedType(field_type) is true instead of only for VARCHAR.

src/include/milvus/types/FieldData.h
```diff
@@ -371,6 +375,7 @@ using FloatFieldData = FieldData<float, DataType::FLOAT>;
 using DoubleFieldData = FieldData<double, DataType::DOUBLE>;
 using VarCharFieldData = FieldData<std::string, DataType::VARCHAR>;
 using GeometryFieldData = VarCharFieldData;     // geometry field data is passed as string
+using TextFieldData = VarCharFieldData;         // text field data is passed as string
```
yhmo, line 376: src/include/milvus/types/FieldData.h line:378
Low ---- Consider adding a TEXT usage example (similar to examples/src/v2/geometry_field.cpp and timestamptz_field.cpp extended in this PR). The PR's central feature has no example, and TEXT row/column-based insert plus query decoding is only discoverable from code and tests, not from the example surface.

src/impl/utils/TypeUtils.cpp
yhmo: TypeUtils.cpp defines std::to_string(milvus::DataType) inside namespace std, and its name map includes {   DataType::TEXT, "TEXT"}

cmake/MilvusProtoGen.cmake
```diff
@@ -16,7 +16,7 @@
 
 include_guard(GLOBAL)
 
-set(PROTO_COMMIT 7efdf25099dd41c8215f47b7a6525aa0e22eea32)
+set(PROTO_COMMIT c39cddab3fac51c1a1228136ce5e95b131bd2ce6)
```
yhmo, line 19: cmake/MilvusProtoGen.cmake line:19
Low ---- This proto bump is not needed for TEXT support: the previous commit 7efdf25099 already defines proto::schema::DataType::Text = 25 plus GeometryWktArray/geometry_wkt_data, so the SDK compiles against it unchanged. Bumping to c39cddab silently pulls in 17 unrelated proto commits (Decimal=30, UUID=31, TypeSchema, and new ScalarField/VectorField valid_data fields that deprecate the FieldData.valid_data the SDK still writes). If the bump is intentional (for example required by the target server version), a note in the PR description would help; otherwise consider whether it belongs in this "Support Text datatype" change.

test/st/milvus_container.py
```diff
@@ -73,6 +73,8 @@ def start_milvus_container(port: int = milvus_port, image: str = milvus_image, t
             "ETCD_USE_EMBED": "true",
             "ETCD_DATA_DIR": "/var/lib/milvus/etcd",
             "COMMON_STORAGETYPE": "local",
+            # Storage V3 is required to create collections containing TEXT fields
+            "COMMON_STORAGE_USELOONFFI": "true",
```
yhmo, line 77: test/st/milvus_container.py line:77
Low ---- Setting COMMON_STORAGE_USELOONFFI unconditionally in the shared ST harness switches the server storage engine (Storage V3) for the whole system-test suite, not just the TEXT tests. If the other ST cases were validated against the previous storage engine, please confirm the full suite still passes with LoonFFI enabled, or scope this flag to the tests that actually need TEXT.

cmake/MilvusProtoGen.cmake
yhmo: as designed

test/st/milvus_container.py
yhmo: as designed

src/impl/utils/DmlUtils.cpp
```diff
@@ -691,6 +712,22 @@ CreateProtoArrayField(const FieldDataSchema& data_schema, proto::schema::FieldDa
             }
             break;
         }
+        case DataType::GEOMETRY: {
```
yhmo: src/impl/utils/DmlUtils.cpp line:715
Low ---- The new array-of-string-backed element support (TEXT/GEOMETRY/TIMESTAMPTZ, including this geometry_wkt_data branch) is covered only by unit tests and the in-process mocked-server IT tests; there is no real-server (system-test) case for any array-of-string-backed field, and on current Milvus server master the proxy array validator (internal/proxy/validate_util.go checkArrayElement) has no branches for Text/Timestamptz/Geometry element types. The server-side contract for these array wire formats is unverified; consider one ST insert+query round-trip for an array-of-string-backed field (or confirm server support) before treating these array paths as supported.

test/st/cases/TestDml.cpp
```diff
@@ -222,6 +223,7 @@ TEST_F(MilvusServerTestDml, InsertAndQuery) {
         {"f_array", {10, 20, 30}},
         {"f_geo", "POINT (1 1)"},
         {"f_tsz", "2025-01-01T00:00:00+00:00"},
+        {"f_text", "Alice body text"},
```
yhmo: test/st/cases/TestDml.cpp line:226
Low ---- TEXT's distinguishing behavior (no max_length limit) is only proven at the unit-conversion level (the 70,000-byte string in CheckAndSetRowDataText). The system tests only round-trip short values like this one, so nothing exercises a >65535-byte TEXT value through a real server round-trip (gRPC limits, Storage V3, query decode). Consider one ST row with a long body to lock in the end-to-end contract.

src/include/milvus/types/FieldData.h
```diff
@@ -387,6 +392,9 @@ using ArrayInt64FieldData = ArrayFieldData<int64_t, DataType::INT64>;
 using ArrayFloatFieldData = ArrayFieldData<float, DataType::FLOAT>;
 using ArrayDoubleFieldData = ArrayFieldData<double, DataType::DOUBLE>;
 using ArrayVarCharFieldData = ArrayFieldData<std::string, DataType::VARCHAR>;
+using ArrayGeometryFieldData = ArrayVarCharFieldData;     // geometry array element data is passed as string
```
yhmo: src/include/milvus/types/FieldData.h line:395
Low ---- The new array aliases here (and the doc lines 214-216) advertise GEOMETRY/TIMESTAMPTZ array-element support, but the same PR removes the geometry/timestamptz array-element decode branches in DqlUtils::BuildMilvusArrayFieldData and rejects those element types in CheckInsertInput/CheckAndSetScalar (IsArrayStringBackedElementType only covers VARCHAR/TEXT). A column built from ArrayGeometryFieldData/ArrayTimestamptzFieldData can never round-trip: insert is rejected client-side, and a server response carrying such an element type now fails to decode with NOT_SUPPORTED (previously it decoded for Timestamptz via string_data and Geometry via geometry_wkt_data). Consider either keeping those decode paths and accepting these element types (consistent with the ArrayTextFieldData forward-looking path), or dropping the aliases and the doc claims so the public surface does not advertise behavior the SDK deliberately rejects.

src/impl/utils/DmlUtils.cpp
```diff
@@ -1385,8 +1432,23 @@ CheckAndSetScalar(const nlohmann::json& obj, const FieldSchema& fs, proto::schem
             scalars->Add(obj.get<double>());
             break;
         }
-        case DataType::VARCHAR:
+        case DataType::VARCHAR: {
```
yhmo: src/impl/utils/DmlUtils.cpp line:1435
Low ---- The string-backed cases in CheckAndSetScalar now duplicate the is_string + max_length validation across VARCHAR/TIMESTAMPTZ/GEOMETRY (previously VARCHAR and TIMESTAMPTZ shared a single block) and TEXT. A small shared helper (string check + optional max-length + target container) would keep these four blocks from drifting when a future type or validation rule changes.

src/impl/utils/DmlUtils.cpp
```diff
@@ -1385,8 +1432,23 @@ CheckAndSetScalar(const nlohmann::json& obj, const FieldSchema& fs, proto::schem
...
+            if (ss.size() > fs.MaxLength()) {
+                return {StatusCode::INVALID_ARGUMENT, "Exceeds max length of field: " + fs.Name()};
+            }
+            auto scalars = sf->mutable_string_data()->mutable_data();
+            scalars->Add(std::move(ss));
+            break;
+        }
         case DataType::TIMESTAMPTZ: {
+            // TIMESTAMPTZ is a scalar string-backed type but is not supported as an ARRAY element
+            if (is_array) {
```
yhmo, line 1469: src/impl/utils/DmlUtils.cpp line:1449
Low ---- These new row-based array-element guards (here TIMESTAMPTZ, and GEOMETRY at line 1464) turn input that previously passed SDK-side validation into NOT_SUPPORTED: before this PR a row-based insert of an ARRAY field with GEOMETRY/TIMESTAMPTZ elements was serialized and sent, now it is rejected locally. That is consistent with the server (proxy verifyCapacityPerRow rejects those element types) and with the column-based rejection unit-tested in CheckInsertInputRejectsUnsupportedArrayElementTypes, but this row-based rejection path itself has no direct test. Consider one row-based case asserting NOT_SUPPORTED to pin the new behavior.

.github/workflows/main.yaml
```diff
@@ -192,11 +195,12 @@ jobs:
           key: macos-15-ccache-${{ github.sha }}
           restore-keys: macos-15-ccache-
       - name: Prepare
+        timeout-minutes: 30
         run: |
           mkdir ~/.venv
           python3 -m venv ~/.venv
           source ~/.venv/bin/activate
-          sh scripts/install_deps.sh
+          timeout -k 10 1800 sh scripts/install_deps.sh
```
yhmo: .github/workflows/main.yaml line:203
Medium ---- The macOS Prepare step now wraps install_deps.sh in `timeout -k 10 1800`, but `timeout` is a GNU coreutils binary that stock macOS does not ship (coreutils provides it as `gtimeout`). Unless the macos-15 runner image has coreutils installed and linked, this step fails with "timeout: command not found" and the whole macOS job breaks. The step-level `timeout-minutes: 30` already bounds the job, so consider dropping the wrapper on macOS or guarding it with a `command -v timeout` fallback. The same pattern is applied to the macOS job in release.yaml (line:199).

.github/workflows/main.yaml
```diff
@@ -99,8 +99,9 @@ jobs:
           key: linux-${{ matrix.os.key }}-ccache-${{ github.sha }}
           restore-keys: linux-${{ matrix.os.key }}-ccache-
       - name: Prepare
+        timeout-minutes: 30
         run: |
-          sh scripts/install_deps.sh
+          timeout -k 10 1800 sh scripts/install_deps.sh
```
yhmo, line 104: .github/workflows/main.yaml line:104
Low ---- These workflow edits (timeout-minutes + `timeout` wrapper around install_deps, and the apt network-timeout config in the coverage job) are unrelated to TEXT datatype support and broaden an already XL-size PR. If they were needed to stabilize CI for the new system tests, a one-line note in the PR description would help reviewers; otherwise consider splitting them into a separate change.

PR #583, opened 2026-08-18
Support client-side roaring bitmap blobs for roaring_match

xiaofan-luan:
`roaring_match(field, {ids})` is the exact-membership sibling of `bloom_match`, added server-side in milvus-io/milvus#51968 (merged). The client compresses a membership set into a roaring bitmap and ships the blob as a template value, so a set that would blow past the proxy's gRPC receive limit as an `in [...]` list rides the wire in a fraction of the space — and, unlike a bloom filter, with no false positives, which is why the server also permits it in delete expressions.

The expression accepts **only** a `{template}` placeholder resolving to protobuf `bytes` holding a complete MRB1 blob. There is no literal-list form. So without an SDK-side builder the feature is unreachable from C++ entirely.

## What this adds

- `milvus/types/RoaringBitmap.h` / `src/impl/types/RoaringBitmap.cpp` — `RoaringBitmapBuilder` (`AddInt64`, `AddInt64s`, `Cardinality`, `Stats`, `Validate`, `Build`, `BuildTemplate`) and the convenience `RoaringBitmapTemplate(members, output)`.
- 12 unit tests plus the shared cross-SDK golden vectors.

`BuildTemplate()` returns the `nlohmann::json::binary` value the caller holds and reuses across queries — building a large bitmap is the expensive part and amortizing it is the entire point of the feature.

No new dependency, and the template path is **not** re-plumbed: `IsValidTemplate` / `ConvertFilterTemplates` already carry `is_binary()` from #564. The only edits to existing sources are three comment updates naming the new builder alongside the bloom one, plus a `MILVUS_UT_TESTDATA_DIR` compile definition for the fixture.

## Why no CRoaring dependency

Most roaring libraries' default serialization is *not* the portable format segcore decodes — that mismatch is the documented trap in the design doc. The writer is ~200 lines of well-specified layout, and the client only ever writes and never parses, so hand-writing it avoids a native dependency and a serialization dialect we don't control.

## Verification

The bytes must match what every other SDK produces, so conformance is byte-identity against the Go reference (`client/roaringfilter.Build`), not just "the server accepts it":

- **29/29 shared golden vectors byte-identical**, plus each case's cardinality, body length, high/low container counts and array/bitmap/run histogram.
- **300/300 additional randomized cases byte-identical** to the Go reference, on inputs not in the fixture set.
- The fixtures themselves were verified to pass the server's own validator and to decode under CRoaring — the library segcore probes with — with exact membership.
- 4 deliberate mutations (8224→8192 tie-break, signed sort, dropped offset table, off-by-one descriptor) each turned golden cases red.

Two subtleties worth a reviewer's attention:

1. **Sorting is unsigned.** `-1` is the *largest* key, not the smallest, and it belongs in the top high container.
2. **The run-vs-bitmap tie-break compares against 8224, not 8192.** That is the reference's in-memory `bitmapContainerSizeInBytes()` (`unsafe.Sizeof(bitmapContainer{}) + 65536/8`). Using 8192 flips the encoding for containers with 2048..2055 runs and desynchronizes this SDK's bytes from every other one. The `run_bitmap_tiebreak_run` / `run_bitmap_tiebreak_bitmap` vectors sit either side of it.

Every multi-byte field is written with explicit `static_cast<uint8_t>(v >> n)` shifts rather than a `memcpy` of a wider integer, so the blob is little-endian regardless of host byte order.

## Fixture placement differs from bloom, deliberately

Bloom inlines its vectors as a raw string literal in the test source. At 92 KB that is impossible on the Windows CI job — MSVC caps a single string literal at 16380 bytes. The file is checked in at `test/ut/testdata/` and reached through a `MILVUS_UT_TESTDATA_DIR` compile definition (absolute path, so it works from any working directory). This also keeps the file byte-identical to the copy every other Milvus SDK repo carries.

## Tests

```
make test    # testing-ut 794/794 pass (12 new), testing-it 288/288 pass
```

Baseline on clean `origin/master` in the same worktree: 782/782 and 288/288, so there are no pre-existing failures in play. clang-format 17.0.6 and clang-tidy 17 are clean on every touched file.

## Also fixes: `DeleteRequest::AddFilterTemplate` was uncallable

Found while writing the delete-path test. `DeleteRequest` declared the method twice, differing only in how the key is taken — `(std::string, json&&)` and `(const std::string&, json&&)`. Overload resolution can never prefer one of those, so **every** call was ambiguous and the method could not be called at all. Nothing in the repo called it: not the SDK internals, not the unit tests, not `test/st`, not the examples. A note from #452 claims it is "tested via integration tests", but an uncallable method cannot be called from anywhere.

Removing the const-reference overload is source-compatible by construction: no call to it can exist, because no call to it could ever have compiled.

It is fixed here rather than split out because exactness is the whole reason `roaring_match` is distinct from `bloom_match` — the server permits it in a delete expression precisely because it has no false positives, while `bloom_match` stays rejected there. Shipping `roaring_match` for C++ while the delete-side `AddFilterTemplate` is uncallable would leave the feature's headline use case reachable only through a workaround. Happy to split it out if you would rather review it separately.

Design doc: `docs/design-docs/design_docs/20260714-roaring-exact-membership-expression.md` in milvus-io/milvus.


xiaofan-luan:
Good catch — fixed in `ca58a90`.

You are right that this is the same failure mode the deleted floating-point overload exists for: `AddInt64(UINT64_MAX)` selected `AddInt64(int64_t)`, narrowed, and inserted the member `-1`, so `roaring_match` would then report an exact match on a value the caller never supplied. Silent wrong answer, no error.

So it gets the same treatment rather than a runtime check: an `= delete`d overload for unsigned types wider than 32 bits.

I deliberately scoped it to 64-bit unsigned only. `uint8_t` / `uint16_t` / `uint32_t` all widen to `int64_t` losslessly, so deleting those would cost callers a cast and buy nothing — 64-bit is the only width that can carry a value `int64_t` cannot represent. A caller who really holds `uint64_t` ids now has to say which signed value it means, which is the right place for that decision.

`static_assert`s in `RejectsFloatingPointMembers` pin both halves: narrow unsigned accepted, `uint64_t` and `unsigned long long` rejected. 794/794 unit tests still green, clang-format 17 clean.

xiaofan-luan:
Folded one more fix into this PR (`e2822eb`), flagged while writing the delete-path test here.

**`DeleteRequest::AddFilterTemplate` could not be called at all.** It was declared twice, differing only in how the key is taken:

```cpp
AddFilterTemplate(std::string key,        nlohmann::json&& t);   // DeleteRequest.h:74
AddFilterTemplate(const std::string& key, nlohmann::json&& t);   // DeleteRequest.h:92
```

Overload resolution can never prefer one of those — copying into a by-value parameter and binding to a const reference are both identity conversions from the same argument. So it is not that some calls are ambiguous; **every** call is:

```
error: call to member function 'AddFilterTemplate' is ambiguous
  note: candidate function  AddFilterTemplate(std::string key, ...)
  note: candidate function  AddFilterTemplate(const std::string& key, ...)
```

Nothing in the repo called it — not the SDK internals (the two call sites in `MilvusClientV2Impl` are `QueryRequest`, which declares one overload), not the unit tests, not `test/st`, not the examples. #452 left this note:

```cpp
// Note: DeleteRequest::AddFilterTemplate has ambiguous overloads (std::string vs const std::string&),
// skip testing it here. The method is tested via integration tests.
```

but no integration test calls it either — an uncallable method cannot be called from anywhere. That comment recorded a real defect as a testing gap, and it has sat since April.

Removing the const-reference overload is source-compatible by construction: no call to it can exist, because no call to it could ever have compiled. The by-value one is kept, matching `QueryRequest` / `SearchRequest` / `QueryArguments`, and it can move the key into the map.

Why it belongs in this PR rather than a separate one: exactness is the whole reason `roaring_match` is distinct from `bloom_match` — the server permits it in a delete expression precisely because it has no false positives, while `bloom_match` stays rejected there. So the C++ SDK shipping `roaring_match` while its delete-side `AddFilterTemplate` is uncallable would ship the feature with its headline use case reachable only through a workaround. The roaring test now uses the method directly instead of routing around it via `WithFilterTemplates`.

The regression guard is the call itself: reintroducing a second overload makes `DeleteRequestTest.GettersAndSetters` fail to compile. It exercises both an lvalue `std::string` and a string literal.

Full suite after the change: `testing-ut` 794/794, `testing-it` 288/288, clang-format 17 clean. Happy to split this out if you would rather review it separately.

src/impl/types/RoaringBitmap.cpp
```diff
@@ -0,0 +1,435 @@
...
+uint64_t
+RoaringBitmapBuilder::Cardinality() const {
+    normalize();
+    return keys_.size();
+}
+
+RoaringBitmapStats
+RoaringBitmapBuilder::Stats() const {
+    normalize();
+    return StatsOf(keys_.size(), PlanLayout(keys_));
```
yhmo, line 450: src/impl/types/RoaringBitmap.cpp line:389
Medium ---- The limits are only enforced after PlanLayout() has materialised the full layout (one ContainerPlan per 16-bit bucket, one GroupPlan per high bucket). For a sparse membership set that the server will reject -- the "shuffled full-range INT64" case the design doc calls out -- Stats()/Validate()/Build() allocate ~72 bytes per distinct container first and only then fail the high-container/decoded-size check (e.g. 10M random int64 members land in ~10M separate high containers, costing on the order of 700MB of transient layout before rejection). The Go reference (client/roaringfilter) counts distinct high>>32 and low>>16 buckets in one O(n) pass over the sorted keys and runs the body=0 estimate check before building anything. Consider deriving the two counts from the normalised keys and pre-checking the limits before PlanLayout(), so an oversized sparse set is rejected without the layout allocation.

src/include/milvus/types/RoaringBitmap.h
```diff
@@ -0,0 +1,207 @@
...
+     *
+     * Note: this throws std::runtime_error when the member set exceeds a limit. Call Validate()
+     * first, or use RoaringBitmapTemplate(), when the set comes from untrusted input.
+     */
+    std::vector<uint8_t>
+    Build() const;
+
+    /**
+     * @brief Same blob as Build(), wrapped as a JSON binary value so it can be handed to
+     * QueryArguments::AddFilterTemplate() / SearchArguments::AddFilterTemplate() directly.
```
yhmo: src/include/milvus/types/RoaringBitmap.h line:179
Low ---- BuildTemplate() doc lists only QueryArguments::AddFilterTemplate() and SearchArguments::AddFilterTemplate() as consumers. This PR also makes DeleteRequest::AddFilterTemplate() callable (its headline use case is the delete expression), and that methods own doc comment still states valid template values are "boolean, numeric, string, array" -- binary blobs (the roaring_match/bloom_match path) are omitted. A caller or future maintainer reading either comment could reasonably conclude binary templates are unsupported in delete expressions. Suggest mentioning DeleteRequest here and updating the DeleteRequest::AddFilterTemplate doc comment to include binary.

src/include/milvus/types/RoaringBitmap.h
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * Members are signed: a value is sign-extended to int64 and its two's complement bit pattern
+ * is the uint64 bitmap key, so INT8(-1), INT32(-1) and INT64(-1) are the same member, and
+ * negative values sort above positive ones inside the bitmap. That mapping is what the server
+ * probes with, so a field of any signed integer width can be matched against one blob.
+ *
+ * Duplicate and unordered input is fine: members collapse to a distinct sorted set, and the
+ * same set always yields the same bytes.
+ */
+class MILVUS_SDK_API RoaringBitmapBuilder {
```
yhmo, line 109: src/include/milvus/types/RoaringBitmap.h line:109
Low ---- RoaringBitmapBuilder is not included by the aggregate milvus/MilvusClientV2.h header, so a consumer following the documented "include only MilvusClientV2.h" contract cannot reach it without knowing to include the extra header. The repo convention is that the aggregate header covers every public type; BloomFilterBuilder has the same gap from #564, but adding `#include "types/RoaringBitmap.h"` here would establish the pattern for filter builders.

src/include/milvus/types/RoaringBitmap.h
```diff
@@ -0,0 +1,193 @@
...
+ public:
+    /**
+     * @brief Insert one member.
+     */
+    RoaringBitmapBuilder&
+    AddInt64(int64_t value);
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

xiaofan-luan:
Thanks @yhmo — all three addressed in `7785e97`.

**Limits enforced after the layout (Medium).** You are right, and the fix is the one you describe. `Validate()` and `Build()` now count distinct high-32 and low-16 buckets in one pass over the normalised keys and check the limits on that before `PlanLayout()` runs, matching what the Go reference does.

Two things I made sure of, since adding a second gate is the kind of change that can quietly alter behaviour:

- **It cannot over-reject.** The early estimate uses `body_length = 0`, which can only understate the real estimate, so it rejects a strict subset of what the exact `CheckLimits()` rejects. The accept set is unchanged.
- **Which gate fired is not observable.** It reuses `CheckLimits()`'s own messages, and the test now asserts `Build()`'s thrown message equals `Validate()`'s for an over-limit set.

The equivalence is what the tests pin, since the memory saving itself is not something a unit test can assert honestly: a set exactly at the high-container limit still validates, and the fourteen-group case in `RejectsAnOversizedDecodedBitmap` — whose lower-bound estimate fits but whose real estimate does not — still has to reach the exact check rather than being turned away early. Mutating the counter to over-count by one turns `RejectsTooManyHighContainers` red, which is the failure mode that matters here.

`Stats()` I deliberately left as it was: its per-container histogram genuinely cannot be known without laying the containers out, so making it "cheap" would mean making it lie. Its doc now says to call `Validate()` first for an untrusted set, which is what `RoaringBitmapTemplate()` already does.

**Aggregate header (Low).** Added `#include "types/RoaringBitmap.h"` to `MilvusClientV2.h`. `BloomFilterBuilder` has the same gap from #564 — I left it alone to keep this change to one type, but it is a one-line addition and I am happy to include it here if you would rather fix both at once.

**Delete-path docs (Low).** Both corrected. `BuildTemplate()` now lists `DeleteRequest::AddFilterTemplate()` as a consumer and says why delete is the interesting one — `roaring_match` is exact, so the server permits it in a delete expression and rejects `bloom_match` there. And `DeleteRequest::AddFilterTemplate`'s own comment now lists `binary` among valid template values, pointing at `RoaringBitmapBuilder::BuildTemplate()`. That comment was doubly misleading given the method was uncallable until this PR, so thank you for catching that the fix was incomplete without it.

Verification after the change: `testing-ut` 794/794, `testing-it` 288/288, clang-format 17 clean, and the 300 randomized cross-check cases against the Go reference still 300/300 byte-identical.

xiaofan-luan:
Correction to my previous comment, pushed as `e7e19c1`.

I claimed the early gate was not observable. That was wrong in one dimension, and I found it by porting the same gate to pymilvus (milvus-io/pymilvus#3764).

The gate compared a **lower-bound** decoded-size estimate, computed with `body_length = 0`. As a verdict that is sound — it can only reject sets the exact check also rejects. But the decoded-size error *reports the size*, and a caller sizing a membership set would have been handed the bound rather than the real figure to shrink against. In pymilvus the equivalent test asserts the exact number rather than just which limit fired, and it went red the moment I wired the gate in. The C++ tests did not catch it because the fourteen-group case in `RejectsAnOversizedDecodedBitmap` happens to sit under the bound, so it never took the early path.

The gate now checks **only the high-container count**, which the counting pass computes exactly, and the decoded-size limit stays in `CheckLimits()` where the body length is known.

Nothing is lost by narrowing it. The shape this gate exists for — the one you described, sparse ids landing in nearly one high container each — is precisely what the high-container limit describes, so it is still refused before `PlanLayout()` runs. A set that blows only the decoded-size limit is one whose container count is already bounded, so laying it out is proportional work rather than a blow-up. The low-container count is no longer needed and the counting pass no longer computes it.

`testing-ut` 794/794, clang-format 17 clean, and the 300 randomized cross-check cases against the Go reference still 300/300 byte-identical.

src/include/milvus/types/RoaringBitmap.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/include/milvus/types/RoaringBitmap.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/types/RoaringBitmap.cpp
yhmo: - Major problem — fixed. The scenario the comment quantified (10M random int64 → ~10M high containers → ~700MB transient layout) is now rejected by the pre-layout CountHighContainers gate in Validate()/Build().                                                                  

- Minor problem — not fixed. The decoded-size/body-size limits are still enforced only after PlanLayout(). The residual risk: a set that stays under the high-container cap (≤ 2^18 high groups) but spreads members across a huge number of low containers could still allocate a large layout before the decoded-size check rejects it.

Generally mark as resolved.

xiaofan-luan:
@yhmo — I pushed `4844382`, which closes the half of the Medium point you recorded as not fixed. Apologies for costing you the `lgtm`; I judged one more round here cheaper than a separate follow-up PR needing a fresh review cycle, but say the word if you would rather I had left it and I will split it out next time.

You were right that the residual mattered. The counting pass now takes the low-container count as well — one extra comparison per key, still no allocation — and both limits are decided from it before `PlanLayout()` runs. Your example shape is the new test: 200K groups of 4 containers is 800K container plans that never get built, because 76.8 MB of fixed overhead already exceeds the 64 MB cap.

One deliberate detail, since it is the reason I narrowed the gate to the high-container count in the first place and got it wrong. The decoded-size estimate is not exact before the layout exists — it includes the body length — so the gate compares the per-container **overhead**, which is a lower bound. Rejecting on a lower bound is sound as a verdict, but the error *reports a size*, and quoting the bound as the size understates what the caller has to shrink. So the message says:

```
Roaring bitmap estimated decoded size is at least 76800000, exceeding maximum 67108864
```

Borderline sets, where the body is what tips it over, still fall through to `CheckLimits()` and get the exact figure. Both sides are pinned: the new test asserts the "at least" wording for a set refused from the counts alone, and `RejectsAnOversizedDecodedBitmap` now also asserts its own overhead is *under* the cap, so it remains a test of the exact path rather than silently drifting onto the early one.

pymilvus landed on the same shape in milvus-io/pymilvus#3764 — @XuanYang-cn raised this independently there, which is a decent sign the split is the right one.

`testing-ut` 795/795 (12 → 13 roaring tests), clang-format 17 clean, and the 300 randomized cross-check cases against the Go reference still 300/300 byte-identical. Disabling the new check turns the new test red.

yhmo:
"make lint" failed; remove "const" from these lines:
auto high_container_status = CheckHighContainerLimit(counts.high_container_count);  // 279                             
auto high_container_status = CheckHighContainerLimit(stats.high_container_count);   // 294                             
auto status = CheckBucketLimits(CountBuckets(keys_));                                // 461

PR #584, opened 2026-08-20
Support global cluster

yhmo:


src/impl/utils/TopologyRefresher.cpp
```diff
@@ -0,0 +1,109 @@
...
+        // keep the cached topology, retry next interval
+        return;
+    }
+
+    {
+        std::lock_guard<std::mutex> lock(mtx_);
+        if (topology.Version() == current_version_) {
+            return;
+        }
+        current_version_ = topology.Version();
```
yhmo: src/impl/utils/TopologyRefresher.cpp line:102
Medium ---- current_version_ is advanced before on_change_() runs, and RefreshOnce() early-returns whenever a later fetch reports the same version, so if ReconnectToPrimary() fails to connect (e.g. the new primary is briefly UNAVAILABLE during failover), the callback is never invoked again and the client stays stuck on the dead primary until the next topology version bump. This also contradicts the "refresher will retry next interval" comment at ConnectionHandler.cpp:130. Consider committing the new version only after a successful reconnect, or tracking a pending-reconnect state and retrying the same version until it succeeds.

src/CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ file(GLOB_RECURSE impl_files
     "${IMPL_DIR}/*.cc"
 )
 
+# cpp-httplib (vendored in thirdparty) needs OpenSSL for the HTTPS topology fetch
+find_package(OpenSSL REQUIRED)
```
yhmo: src/CMakeLists.txt line:25
Medium ---- find_package(OpenSSL REQUIRED) makes OpenSSL a hard configure-time requirement for every build of the SDK, even builds that never use the global-cluster topology feature, and the OpenSSL link is PRIVATE and omitted from the MILVUS_EXPORT_STATIC_DEPS export block. Static-library consumers (BUILD_SHARED_LIBS=OFF + install/export path) will therefore get unresolved SSL/Crypto references unless they add OpenSSL manually, unlike protobuf/gRPC which are exported under MILVUS_EXPORT_STATIC_DEPS. Consider gating the requirement behind the feature (or documenting it) and adding OpenSSL::SSL OpenSSL::Crypto to the MILVUS_EXPORT_STATIC_DEPS block for static consumers.

src/impl/utils/TopologyRefresher.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+void
+TopologyRefresher::Stop() {
+    {
+        std::lock_guard<std::mutex> lock(mtx_);
+        stop_.store(true);
+    }
+    cv_.notify_all();
+    if (thread_.joinable()) {
+        thread_.join();
```
yhmo, line 57: src/impl/utils/TopologyRefresher.cpp line:57
Low ---- Stop() (reached from the destructor, which ConnectionHandler::Disconnect()/Connect() and destruction trigger while still holding mtx_) joins the refresh thread, but the thread can be blocked in FetchTopology() for up to 3*10s + backoff (~30s). A Disconnect() during an in-flight topology fetch will therefore stall for tens of seconds. Consider a bounded join timeout or a stop check inside the fetch/retry loop so shutdown is prompt.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -36,18 +62,82 @@ ConnectionHandler::Connect(const ConnectParam& connect_param) {
...
+    }
+
+    std::string new_primary_uri = BuildPrimaryUri(primary->Endpoint());
+    // a topology version bump with the same primary endpoint needs no reconnect
+    // (matching Java/pymilvus behavior)
+    if (connection_ != nullptr && connection_->GetConnectParam().Uri() == new_primary_uri) {
+        return;
+    }
+
+    ConnectParam primary_param = global_connect_param_;
```
yhmo: src/impl/utils/ConnectionHandler.cpp line:124
Medium ---- ReconnectToPrimary() rebuilds the connection from the global_connect_param_ snapshot captured at Connect() time, so post-connect configuration applied through the public API is silently lost on failover: UseDatabase(db) (MilvusConnection::UseDatabase mutates only the live connection param) and SetRpcDeadlineMs() will revert to their connect-time values, and subsequent operations may silently target the wrong database. Consider re-applying the current DbName()/RpcDeadlineMs() from the existing connection (or tracking them in ConnectionHandler) when building primary_param.

test/ut/utils/TestGlobalCluster.cpp
```diff
@@ -0,0 +1,152 @@
...
+    ASSERT_TRUE(status.IsOk()) << status.Message();
+    EXPECT_EQ(topology.Version(), 7);
+    ASSERT_EQ(topology.Clusters().size(), 1u);
+    EXPECT_EQ(topology.Clusters()[0].Endpoint(), "a:19530");
+
+    server.stop();
+    srv.join();
+}
+
+TEST(TopologyRefresherTest, CallbackOnVersionChange) {
```
yhmo, line 231: test/ut/utils/TestGlobalCluster.cpp line:119
Low ---- The new tests cover GlobalClusterUtils and TopologyRefresher in isolation, but there is no test for the ConnectionHandler global-mode integration this PR adds (~300 lines of failover/reconnect logic): global-endpoint Connect()/Disconnect(), BuildPrimaryUri() scheme selection, ReconnectToPrimary() failover including the reconnect-failure path, TriggerGlobalRefresh() on RPC UNAVAILABLE, and the CurrentEndpoint() cache-key behavior. A mocked test (local httplib server serving changing topologies) would guard the core failover behavior that the current tests leave uncovered.

test/CMakeLists.txt
```diff
@@ -48,6 +48,10 @@ file(GLOB_RECURSE ut_files
 add_executable(testing-ut ${ut_files})
 target_compile_options(testing-ut PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/bigobj>)
 target_include_directories(testing-ut PRIVATE ${PROJECT_SOURCE_DIR}/thirdparty)
+# cpp-httplib must be compiled with the same configuration as the SDK (which enables OpenSSL)
+# to avoid ODR violations; the unit tests use httplib::Server for the local topology mock.
+target_compile_definitions(testing-ut PRIVATE CPPHTTPLIB_OPENSSL_SUPPORT)
+target_link_libraries(testing-ut PRIVATE OpenSSL::SSL OpenSSL::Crypto)
```
yhmo: test/CMakeLists.txt line:54
High ---- `OpenSSL::SSL`/`OpenSSL::Crypto` are referenced here, but `find_package(OpenSSL REQUIRED)` is only called in `src/CMakeLists.txt`. `src` and `test` are sibling subdirectories, so the imported target is not visible in this scope and CMake configure fails with "Target 'testing-ut' links to OpenSSL::SSL but the target was not found" — the Windows CI job currently fails on this exact line, and the Linux/macOS builds (MILVUS_BUILD_TEST=ON) will hit the same once they finish. Add `find_package(OpenSSL REQUIRED)` to `test/CMakeLists.txt` (or at the top level before `add_subdirectory(test)`).

src/impl/utils/GlobalClusterUtils.cpp
```diff
@@ -0,0 +1,199 @@
...
+        if (code != 0) {
+            std::string message = root.value("message", "unknown error");
+            return {StatusCode::SERVER_FAILED,
+                    "Global topology API returned error code " + std::to_string(code) + ": " + message};
+        }
+        if (!root.contains("data") || !root["data"].is_object()) {
+            return {StatusCode::INVALID_ARGUMENT, "Invalid global topology response: missing 'data'"};
+        }
+        auto data = root["data"];
+        if (!data.contains("version") || !data["version"].is_number_integer() || !data.contains("clusters") ||
```
yhmo: src/impl/utils/GlobalClusterUtils.cpp line:131
Medium ---- The topology API returns `data.version` as a JSON string: the pymilvus reference implementation parses it with `int(data["version"])` and its fixtures/tests use `"version": "1"`. Requiring `is_number_integer()` here will reject real responses with "missing version or clusters" and the connect fails. Accept a numeric string too (check `is_number_integer() || is_string()` and parse with `std::stoll`).

src/CMakeLists.txt
```diff
@@ -30,6 +33,7 @@ endif()
 # add proto gens
 add_milvus_protos(milvus_sdk_objects)
 target_link_libraries(milvus_sdk_objects PRIVATE protobuf::libprotobuf gRPC::grpc++)
+target_compile_definitions(milvus_sdk_objects PRIVATE CPPHTTPLIB_OPENSSL_SUPPORT)
```
yhmo, line 47: src/CMakeLists.txt line:36
Low ---- This target-wide `CPPHTTPLIB_OPENSSL_SUPPORT` also flips the pre-existing `src/impl/BulkImport.cpp` httplib usage (which previously compiled the non-SSL `httplib::Client`) to the SSL variant, and `milvus_sdk_objects` never receives OpenSSL include dirs (`OpenSSL::SSL` is only linked to the final `milvus_sdk` target, which does not affect compilation of the object library). That compiles on Linux/macOS only because `/usr/include/openssl` is on the default include path; on Windows the headers live under the OpenSSL install dir, so once the `test/CMakeLists.txt` configure issue is fixed, `GlobalClusterUtils.cpp`/`BulkImport.cpp` will fail to find `<openssl/ssl.h>`. Consider adding `${OPENSSL_INCLUDE_DIR}` to `milvus_sdk_objects` (or linking `OpenSSL::SSL`) and scoping the SSL define to the global-cluster translation unit only.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -26,8 +29,31 @@ ConnectionHandler::Connect(const ConnectParam& connect_param) {
     // private until it succeeds, but setters must not update the current connection and then be overwritten by the
     // successful swap below.
     std::lock_guard<std::mutex> lock(mtx_);
+
+    // stop any previous global-cluster refresher
+    global_refresher_.reset();
+    global_mode_ = false;
+    global_endpoint_.clear();
+
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/GlobalClusterUtils.cpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    auto colon = authority.rfind(':');
+    if (colon == std::string::npos) {
+        host = authority;
+        port = (scheme == "https") ? 443 : 80;
+    } else {
+        host = authority.substr(0, colon);
+        port = std::stoi(authority.substr(colon + 1));
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

test/ut/utils/TestGlobalCluster.cpp
```diff
@@ -0,0 +1,152 @@
...
+    server.Get("/global-cluster/topology",
+               [body](const httplib::Request&, httplib::Response& res) { res.set_content(body, "application/json"); });
+    auto port = server.bind_to_any_port("127.0.0.1");
+    ASSERT_TRUE(server.is_valid());
+    std::thread srv([&server]() { server.listen_after_bind(); });
+
+    milvus::GlobalTopology topology;
+    auto status =
+        milvus::GlobalClusterUtils::FetchTopology("http://127.0.0.1:" + std::to_string(port), "tok", topology);
+    ASSERT_TRUE(status.IsOk()) << status.Message();
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/TopologyRefresher.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/TopologyRefresher.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/ConnectionHandler.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

test/CMakeLists.txt
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/GlobalClusterUtils.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/CMakeLists.txt
yhmo: Fixed: `find_package(OpenSSL REQUIRED)` is now declared at the top-level `CMakeLists.txt` (visible to both src and test), and `OpenSSL::SSL`/`OpenSSL::Crypto` were added to the `MILVUS_EXPORT_STATIC_DEPS` export block so static-library consumers get the link. OpenSSL remains a hard requirement by design — the global-cluster topology fetch is a core SDK feature that always links it, and it is already a transitive gRPC dependency (and installed by `install_deps.sh`).

src/CMakeLists.txt
yhmo: Addressed: `OpenSSL::SSL`/`OpenSSL::Crypto` are now linked to `milvus_sdk_objects`, so cpp-httplib's SSL headers are on the object-library include path on all platforms (fixing the Windows compile). I intentionally kept `CPPHTTPLIB_OPENSSL_SUPPORT` target-wide instead of scoping it to the global-cluster TU: `BulkImport.cpp` uses a plain `httplib::Client` (the SSL define is additive and does not change its behavior), and keeping the define consistent across TUs avoids an ODR mismatch between the SSL and non-SSL cpp-httplib instantiations in the same library.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -36,18 +68,99 @@ ConnectionHandler::Connect(const ConnectParam& connect_param) {
...
+ConnectionHandler::buildPrimaryUri(const std::string& primary_endpoint) const {
+    const ConnectParam& original = global_connect_param_;
+    return GlobalClusterUtils::BuildPrimaryUri(original.Uri(), original.TlsEnabled(), primary_endpoint);
+}
+
+bool
+ConnectionHandler::reconnectToPrimary(const GlobalTopology& topology) {
+    const ClusterInfo* primary = topology.Primary();
+    if (primary == nullptr) {
+        return true;
```
yhmo: src/impl/utils/ConnectionHandler.cpp line:126
Medium ---- reconnectToPrimary() returns true when the refreshed topology has no writable cluster (primary == nullptr), so refreshOnce() commits that topology version even though no reconnect was performed. After that, later refreshes see topology.Version() == current_version_ and skip the callback entirely; if a writable primary reappears at the same version (e.g. the failover window where the old primary is unregistered before the new one is elected), the client stays stuck on the dead connection until the next version bump. This also contradicts the documented callback contract in TopologyRefresher.h ("the new version is only committed ... when the callback succeeds"). Consider returning false when no primary is found so the refresher keeps retrying that version next interval.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -16,18 +16,50 @@
...
+    if (GlobalClusterUtils::IsGlobalEndpoint(connect_param.Uri())) {
+        auto status = GlobalClusterUtils::FetchTopology(connect_param.Uri(), connect_param.Token(), initial_topology);
+        if (!status.IsOk()) {
+            return status;
+        }
+        const ClusterInfo* primary = initial_topology.Primary();
+        if (primary == nullptr) {
+            return {StatusCode::SERVER_FAILED, "No primary (writable) cluster found in global topology"};
+        }
+        global_mode_ = true;
```
yhmo: src/impl/utils/ConnectionHandler.cpp line:55
Medium ---- The global-cluster state (global_mode_/global_endpoint_/global_connect_param_) is committed before the primary connection attempt. If connection->Connect(primary_param) fails later, Connect() returns the error with global_mode_ still true and no refresher started; when connection_ still holds a previous live connection, the client keeps routing RPCs to the old primary while CurrentEndpoint() now reports the new global endpoint, so schema/timestamp cache keys and CurrentDbName() diverge from the endpoint actually used. Consider committing the global state only after the primary connect succeeds (mirroring the "candidate connection remains private until it succeeds" invariant in this same function), or clearing global_mode_/global_endpoint_ on the failure path.

src/impl/utils/GlobalClusterUtils.cpp
```diff
@@ -0,0 +1,242 @@
...
+    return NormalizeEndpoint(endpoint) + TOPOLOGY_PATH;
+}
+
+std::string
+GlobalClusterUtils::BuildPrimaryUri(const std::string& global_uri, bool tls_enabled,
+                                    const std::string& primary_endpoint) {
+    if (primary_endpoint.find("://") != std::string::npos) {
+        return primary_endpoint;
+    }
+    if (global_uri.compare(0, 8, "https://") == 0 || tls_enabled) {
```
yhmo: src/impl/utils/GlobalClusterUtils.cpp line:128
Low ---- For a scheme-less global endpoint (e.g. "xxx.global-cluster.yyy.com") with TLS disabled, BuildTopologyUrl()/NormalizeEndpoint() default the topology fetch to https, but BuildPrimaryUri() derives "http://<primary>" for the primary gRPC endpoint, so the two endpoints disagree on the scheme and a TLS-required cluster fails at connect with a confusing error. Consider treating a scheme-less global_uri as https here too (consistent with NormalizeEndpoint), or documenting the mismatch.

src/impl/utils/GlobalClusterUtils.cpp
```diff
@@ -0,0 +1,242 @@
...
+            res = client.Get(path, headers);
+        } else {
+            httplib::Client client(host, port);
+            client.set_connection_timeout(REQUEST_TIMEOUT_SEC, 0);
+            client.set_read_timeout(REQUEST_TIMEOUT_SEC, 0);
+            res = client.Get(path, headers);
+        }
+
+        if (res && res->status == 200) {
+            return ParseTopologyResponse(res->body, topology);
```
yhmo: src/impl/utils/GlobalClusterUtils.cpp line:223
Low ---- On HTTP 200 with a malformed/truncated body, ParseTopologyResponse() is returned immediately without consuming the remaining retry attempts, even though the loop is designed to tolerate transient failures (pymilvus retries JSON decode errors). A transient truncated 200 from a proxy/LB fails the whole fetch and the refresher waits until the next 5-minute interval to retry. Consider treating a parse failure as a retryable attempt like non-200 statuses.

test/ut/utils/TestGlobalCluster.cpp
```diff
@@ -0,0 +1,183 @@
...
+    served_version.store(2);
+    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+    refresher.Stop();
+
+    EXPECT_GT(calls.load(), 0);
+    EXPECT_EQ(last_version, 2);
+
+    server.stop();
+    srv.join();
+}
```
yhmo: test/ut/utils/TestGlobalCluster.cpp line:183
Low ---- This file ends without a trailing newline (\ No newline at end of file), unlike the other files in test/ut/utils. Add the trailing newline to keep files consistent and avoid EOF warnings in some toolchains/CI.

src/include/milvus/types/GlobalCluster.h
```diff
@@ -0,0 +1,108 @@
...
+        READABLE = 0b01,                // cluster can serve read (query/search) traffic
+        WRITABLE = 0b10,                // cluster can accept write (insert/upsert/DDL) traffic
+        PRIMARY = READABLE | WRITABLE,  // read + write
+    };
+};
+
+/**
+ * @brief A single cluster within a global-cluster topology.
+ */
+class MILVUS_SDK_API ClusterInfo {
```
yhmo: src/include/milvus/types/GlobalCluster.h line:41
Low ---- ClusterInfo/GlobalTopology/ClusterCapability are exported with MILVUS_SDK_API and installed as a public header under src/include/milvus/types/, but no public API exposes them and MilvusClientV2.h does not include this header, so they are installed/exported dead surface with no consumer. If they are meant to be internal, move them under src/impl; if public, expose them through a public API and add the header to the MilvusClientV2.h aggregate include per the repo convention.

src/impl/utils/ConnectionHandler.h
```diff
@@ -23,9 +23,11 @@
 
 #include "../MilvusConnection.h"
 #include "./RpcUtils.h"
+#include "TopologyRefresher.h"
```
yhmo: src/impl/utils/ConnectionHandler.h line:26
Low ---- Including TopologyRefresher.h here pulls <thread>/<condition_variable>/<atomic> into every TU that includes ConnectionHandler.h (e.g. MilvusClientImpl), although the only use is a std::unique_ptr<TopologyRefresher> member whose destructor is defined out-of-line in the .cpp. A forward declaration (class TopologyRefresher;) would suffice and keep the compile surface of this rarely-used feature local to the implementation.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -36,18 +68,99 @@ ConnectionHandler::Connect(const ConnectParam& connect_param) {
...
+    ConnectParam primary_param = global_connect_param_;
+    primary_param.SetUri(new_primary_uri);
+    // re-apply configuration applied after Connect() so failover does not silently lose it
+    if (connection_ != nullptr) {
+        primary_param.SetDbName(connection_->GetConnectParam().DbName());
+        primary_param.SetRpcDeadlineMs(connection_->GetConnectParam().RpcDeadlineMs());
+    }
+
+    auto new_connection = std::make_shared<MilvusConnection>();
+    auto status = new_connection->Connect(primary_param);
```
yhmo, line 214: src/impl/utils/ConnectionHandler.cpp line:150
Medium ---- reconnectToPrimary() calls MilvusConnection::Connect() while holding mtx_. MilvusConnection::Connect() blocks in channel->WaitForConnected() (up to ConnectTimeout, default 10s) and then in the Connect RPC (deadline set to connect_timeout, another ~10s), so a primary failover can hold the connection mutex for up to ~20s. During that window every other SDK operation that needs the lock — the apiHandler() connection snapshot, GetConnection(), CurrentEndpoint(), UseDatabase(), SetRpcDeadlineMs(), and the TriggerGlobalRefresh() called from an UNAVAILABLE RPC — stalls. Consider building the new MilvusConnection outside the lock and only swapping it into connection_ under mtx_.

src/impl/utils/ConnectionHandler.cpp
yhmo: src/impl/utils/ConnectionHandler.cpp line:126
Medium ---- When the topology has no writable primary, reconnectToPrimary() returns true, which commits the new version in refreshOnce(). If the current primary then becomes unreachable while the topology stays at that version (no writable cluster), the UNAVAILABLE-triggered TriggerGlobalRefresh() will keep re-fetching the same version and refreshOnce() early-returns on the equal version, so the client is never reconnected even after a writable primary returns — it stays stuck on the dead connection until the version bumps again. pymilvus raises when no primary exists so the refresh keeps retrying. Consider returning false (so the refresher retries the same version) when primary == nullptr.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -16,18 +16,50 @@
...
+        global_mode_ = true;
+        global_endpoint_ = connect_param.Uri();
+        global_connect_param_ = connect_param;
+        primary_param.SetUri(buildPrimaryUri(primary->Endpoint()));
+    }
+
     auto connection = std::make_shared<MilvusConnection>();
-    auto status = connection->Connect(connect_param);
+    auto status = connection->Connect(primary_param);
     if (!status.IsOk()) {
```
yhmo: src/impl/utils/ConnectionHandler.cpp line:63
Low ---- If connection->Connect(primary_param) fails on the initial global-mode connect, Connect() returns the error but global_mode_ was already set to true (line 55), global_endpoint_ and global_connect_param_ are populated, and no TopologyRefresher is started. The handler is left in global mode: CurrentEndpoint() keeps returning the global endpoint (scoping cache keys to it) even though connection_ is null, and the caller cannot trigger a refresh. Consider resetting global_mode_/global_endpoint_ before returning on this failure path.

test/ut/utils/TestGlobalCluster.cpp
yhmo: test/ut/utils/TestGlobalCluster.cpp line:183
Low ---- The file does not end with a newline (git shows \"\\ No newline at end of file\"). The repo runs clang-format checks in CI (scripts/build.sh -l / make clang-format), and every other source file ends with a trailing newline; this one would be reformatted and fail the format check.

src/impl/utils/ConnectionHandler.cpp
yhmo: Fixed: reconnectToPrimary() now returns false when topology.Primary() is null, so refreshOnce() does not commit that version and keeps retrying the same version next interval.

src/impl/utils/ConnectionHandler.cpp
yhmo: Fixed: Connect() now commits global_mode_/global_endpoint_/global_connect_param_ (and starts the refresher) only after connection->Connect(primary_param) succeeds; on failure the handler stays in non-global mode.

src/impl/utils/GlobalClusterUtils.cpp
yhmo: Fixed: BuildPrimaryUri() now defaults a scheme-less global URI to https too, matching BuildTopologyUrl()/NormalizeEndpoint().

src/impl/utils/GlobalClusterUtils.cpp
yhmo: Fixed: FetchTopology() now treats a malformed/truncated 200 body (INVALID_ARGUMENT from ParseTopologyResponse) as a retryable attempt; an explicit server error (code != 0) still returns immediately.

test/ut/utils/TestGlobalCluster.cpp
yhmo: Fixed: added the trailing newline.

src/include/milvus/types/GlobalCluster.h
yhmo: Fixed: moved GlobalCluster.h/ClusterInfo/GlobalTopology/ClusterCapability under src/impl/types (internal, no MILVUS_SDK_API) since no public API exposes them.

src/impl/utils/ConnectionHandler.h
yhmo: Fixed: forward-declared TopologyRefresher in ConnectionHandler.h and moved the default constructor out-of-line so the unique_ptr member does not require the complete type in consumers.

src/impl/utils/ConnectionHandler.cpp
yhmo: Fixed: reconnectToPrimary() now builds and connects the new MilvusConnection outside mtx_, then swaps it in under the lock.

src/impl/utils/ConnectionHandler.cpp
yhmo: Fixed (same as the earlier thread): reconnectToPrimary() returns false when there is no writable primary, so the refresher retries that version instead of committing it.

src/impl/utils/ConnectionHandler.cpp
yhmo: Fixed (same as the earlier thread): global-mode state is committed only after the primary connect succeeds, and cleared on the failure path.

test/ut/utils/TestGlobalCluster.cpp
yhmo: Fixed: trailing newline added.

src/impl/utils/TopologyRefresher.cpp
```diff
@@ -0,0 +1,117 @@
...
+void
+TopologyRefresher::refreshLoop() {
+    while (true) {
+        {
+            std::unique_lock<std::mutex> lock(mtx_);
+            cv_.wait_for(lock, interval_, [this]() { return stop_.load() || refresh_now_.load(); });
+            if (stop_.load()) {
+                break;
+            }
+            refresh_now_.store(false);
```
yhmo, line 85: src/impl/utils/TopologyRefresher.cpp line:82
Low ---- refresh_now_ is cleared before refreshOnce() runs, so a TriggerRefresh() arriving while a fetch is in flight is not coalesced away — the loop wakes again as soon as refreshOnce() returns and immediately re-fetches. Because ConnectionHandler::TriggerGlobalRefresh() is called from apiHandler on every failed RPC, a sustained UNAVAILABLE burst turns this into a continuous topology-polling loop that hammers the topology endpoint, contradicting the "coalesced" behavior documented on TriggerRefresh(). Consider clearing/re-checking refresh_now_ after refreshOnce() (or debouncing in TriggerRefresh() with a running flag, as pymilvus does) so triggers received during an active fetch do not queue another fetch.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -36,18 +70,115 @@ ConnectionHandler::Connect(const ConnectParam& connect_param) {
...
+            primary_param.SetDbName(connection_->GetConnectParam().DbName());
+            primary_param.SetRpcDeadlineMs(connection_->GetConnectParam().RpcDeadlineMs());
+        }
+    }
+
+    // build + connect the new primary outside the lock: MilvusConnection::Connect() blocks in
+    // WaitForConnected() and the Connect RPC for up to ~2x ConnectTimeout, and holding mtx_ that
+    // long would stall every other SDK operation that snapshots the connection.
+    auto new_connection = std::make_shared<MilvusConnection>();
+    auto status = new_connection->Connect(primary_param);
```
yhmo, line 214: src/impl/utils/ConnectionHandler.cpp line:160
Low ---- The reconnect window between the config snapshot (lines 147-153) and the swap (lines 165-176) runs outside mtx_, and the candidate connection is built from that snapshot. A concurrent SetRpcDeadlineMs()/UseDatabase() — which mutate the live connection's ConnectParam under the same mtx_ and return OK — applied during that window is silently dropped: the failover installs a connection built with the pre-mutation config. This is narrower than the (now-fixed) snapshot-from-Connect-time issue this same function addressed: the fix re-applies live config at snapshot time, but not at swap time. Consider re-reading DbName()/RpcDeadlineMs() inside the swap lock and re-applying them to the candidate (or reconnecting) before installing it.

src/impl/utils/TopologyRefresher.cpp
yhmo: this issue is fixed, resolve this comment

src/impl/utils/ConnectionHandler.cpp
yhmo: this issue is fixed, resolve this comment

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -16,18 +16,52 @@
...
     std::lock_guard<std::mutex> lock(mtx_);
+
+    global_mode_ = false;
+    global_endpoint_.clear();
+
+    bool is_global = GlobalClusterUtils::IsGlobalEndpoint(connect_param.Uri());
+    ConnectParam primary_param = connect_param;
+    GlobalTopology initial_topology;
+    if (is_global) {
+        auto status = GlobalClusterUtils::FetchTopology(connect_param.Uri(), connect_param.Token(), initial_topology);
```
yhmo, line 79: src/impl/utils/ConnectionHandler.cpp line:51
Medium ---- The initial global-cluster topology fetch (GlobalClusterUtils::FetchTopology, up to MAX_RETRIES=3 with 10s connect/read timeouts plus 1s/2s/4s backoff) runs while ConnectionHandler::mtx_ is held, so a slow or unreachable topology endpoint can block every concurrent SDK operation that snapshots the connection (all apiHandler/Invoke paths, GetConnection, SetRpcDeadlineMs, CurrentEndpoint, etc.) for potentially tens of seconds. This is the same blocking-connect-under-lock pattern that reconnectToPrimary was reworked to avoid, where the new MilvusConnection is built and connected outside the mutex before the swap. Consider fetching the topology before acquiring the lock (or outside it, handing the result back under the lock) so a global-cluster Connect() cannot stall unrelated in-flight calls.

src/impl/utils/GlobalClusterUtils.cpp
```diff
@@ -0,0 +1,251 @@
...
+constexpr const char* GLOBAL_CLUSTER_MARKER = "global-cluster";
+constexpr const char* TOPOLOGY_PATH = "/global-cluster/topology";
+constexpr int MAX_RETRIES = 3;
+constexpr int64_t BASE_BACKOFF_MS = 1000;
+constexpr int64_t MAX_BACKOFF_MS = 10000;
+constexpr int REQUEST_TIMEOUT_SEC = 10;
+
+// Split a URL into scheme/host/port/path. Returns false if the URL is malformed.
+bool
+ParseUrl(const std::string& url, std::string& scheme, std::string& host, int& port, std::string& path) {
```
yhmo: src/impl/utils/GlobalClusterUtils.cpp line:41
Low ---- This new ParseUrl duplicates the existing utils/Uri.h ParseURI already used by MilvusConnection, and the two parsers already diverge: ParseUrl treats authority.rfind(:) as the host/port separator, so a bare IPv6 host (e.g. "https://[::1]") is parsed as host "[:", port 1 instead of being rejected cleanly, whereas ParseURI handles bracketed IPv6. Consider reusing ParseURI (with a small IPv6-aware extension if needed) so topology-endpoint parsing cannot drift from the gRPC channel-endpoint parsing.

test/ut/utils/TestGlobalCluster.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+    status = milvus::GlobalClusterUtils::ParseTopologyResponse(R"({"code":0,"data":{"version":1,"clusters":[{}]}})",
+                                                               topology);
+    EXPECT_EQ(status.Code(), milvus::StatusCode::INVALID_ARGUMENT);
+}
+
+TEST(GlobalClusterUtilsTest, FetchTopology) {
+    httplib::Server server;
+    const std::string body = R"({"code":0,"data":{"version":7,"clusters":[
+      {"clusterId":"a","endpoint":"a:19530","capability":3}]}})";
```
yhmo, line 128: test/ut/utils/TestGlobalCluster.cpp line:128
Low ---- FetchTopology's failure/retry paths are untested: a malformed/truncated 200 body being retried, a non-200 HTTP response being retried with backoff, and the should_stop abort are all only covered by code reasoning. A test that serves a bad 200 once and then a valid 200 (asserting the client retries and succeeds), plus one that verifies should_stop aborts promptly, would lock in the retry semantics this PR relies on for failover.

src/impl/utils/ConnectionHandler.cpp
yhmo, line 79: src/impl/utils/ConnectionHandler.cpp line:51
Medium ---- Connect() runs the global-cluster FetchTopology (up to 3 attempts with 10s connect+read timeouts plus backoff, potentially ~60s) while holding mtx_. The lock spans the whole handshake (this fetch and the following connection->Connect), so a slow or unreachable topology endpoint blocks every concurrent SDK operation that needs the mutex - the apiHandler() connection snapshot, GetConnection(), CurrentEndpoint(), UseDatabase(), SetRpcDeadlineMs(), and TriggerGlobalRefresh() - for up to a minute during Connect(). reconnectToPrimary() was already changed to build the connection outside the lock; doing the same here (fetch topology before taking mtx_, or bounding the HTTP timeouts) would keep Connect() from stalling the whole client.

src/impl/utils/TopologyRefresher.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+void
+TopologyRefresher::Stop() {
+    {
+        std::lock_guard<std::mutex> lock(mtx_);
+        stop_.store(true);
+    }
+    cv_.notify_all();
+    if (thread_.joinable()) {
+        thread_.join();
```
yhmo, line 57: src/impl/utils/TopologyRefresher.cpp line:57
Low ---- Stop() joins the refresh thread unconditionally, and the stop/abort predicate only bounds FetchTopology(). When the thread is inside the reconnect callback (reconnectToPrimary) blocked in MilvusConnection::Connect() - WaitForConnected plus the Connect RPC, up to ~2x ConnectTimeout (default 10s each) - a Disconnect() or ConnectionHandler destruction during an in-flight failover will stall for ~20s while the join waits. The should_stop fix bounds the topology fetch but not the reconnect gRPC connect; a bounded join or a stop check passed into the reconnect connect path would keep shutdown prompt.

src/impl/utils/ConnectionHandler.cpp
yhmo: Fixed: Connect() now fetches the global-cluster topology outside mtx_. this issue is fixed, resolve this comment

src/impl/utils/GlobalClusterUtils.cpp
yhmo: Fixed: removed the duplicated ParseUrl and reused utils/Uri.h ParseURI (which handles bracketed IPv6 correctly) with a small port-default fixup for the REST endpoint. this issue is fixed, resolve this comment

test/ut/utils/TestGlobalCluster.cpp
yhmo: Fixed: added FetchTopologyRetriesOnMalformedBody (bad 200 then valid 200), FetchTopologyRetriesOnNon200 (503 then 200), and FetchTopologyAbortsOnShouldStop tests. this issue is fixed, resolve this comment

src/impl/utils/ConnectionHandler.cpp
yhmo: Fixed: Connect() now fetches the topology outside mtx_ (see the other line-51 thread), so a slow topology endpoint no longer stalls concurrent operations. this issue is fixed, resolve this comment

src/impl/utils/TopologyRefresher.cpp
yhmo: Fixed: the reconnect callback now receives a should_stop predicate and checks it before starting the new gRPC connect, so shutdown aborts promptly instead of waiting on an in-flight MilvusConnection::Connect(). this issue is fixed, resolve this comment

src/impl/utils/TopologyRefresher.cpp
```diff
@@ -0,0 +1,120 @@
...
+            cv_.wait_for(lock, interval_, [this]() { return stop_.load() || refresh_now_.load(); });
+            if (stop_.load()) {
+                break;
+            }
+        }
+        refreshOnce();
+        // coalesce: clear the trigger flag only after the fetch completes, so a TriggerRefresh()
+        // arriving while the fetch was in flight does not queue an immediate re-fetch (a
+        // sustained UNAVAILABLE burst must not turn into continuous topology polling)
+        refresh_now_.store(false);
```
yhmo: src/impl/utils/TopologyRefresher.cpp line:87
Medium ---- refresh_now_ is cleared unconditionally after refreshOnce(), so a TriggerGlobalRefresh() that arrives while a fetch is in flight is silently dropped. If that in-flight fetch observed the old topology version (e.g. the failover was not yet registered at the topology API), refreshOnce() early-returns on the equal version and, with the trigger gone, no follow-up fetch is scheduled — the client stays on the dead primary until the next 5-minute interval. This is the flip side of the earlier non-coalescing fix: the flag is consumed even when it was set during the fetch. Consider consuming the flag before refreshOnce() and, after it returns, looping once more immediately if refresh_now_ was set again during the fetch, so a trigger is coalesced into one follow-up fetch instead of zero (accepting that sustained UNAVAILABLE traffic will naturally keep polling).

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -16,18 +16,55 @@
...
+}
+
 Status
 ConnectionHandler::Connect(const ConnectParam& connect_param) {
-    // Serialize the full handshake with lifecycle and configuration mutations. The candidate connection remains
-    // private until it succeeds, but setters must not update the current connection and then be overwritten by the
-    // successful swap below.
+    // stop any previous global-cluster refresher first (its thread may be inside reconnectToPrimary
+    // waiting on mtx_, so the join must happen without holding the lock to avoid a deadlock)
+    stopGlobalRefresher();
```
yhmo: src/impl/utils/ConnectionHandler.cpp line:37
Medium ---- Connect() calls stopGlobalRefresher() (which clears global_mode_) at the very start, before the new handshake succeeds. If the client is already connected in global mode and a re-Connect() fails (e.g. the topology endpoint is unreachable), the function returns the error but the previous primary connection_ is kept with its global tracking destroyed: no refresher is running, global_mode_ is false, and CurrentEndpoint() starts returning the primary URI instead of the global endpoint — silently changing cache-key scope and disabling automatic failover on an otherwise usable connection. This is the complement of the initial-connect fix (commit global state only after success): the previously-established global state was torn down before the attempt. Consider snapshotting the prior global state and restoring the refresher/global_mode_ on the failure path when an old connection still exists.

src/impl/utils/GlobalClusterUtils.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+bool
+GlobalClusterUtils::IsGlobalEndpoint(const std::string& uri) {
+    if (uri.empty()) {
+        return false;
+    }
+    std::string lowered = uri;
+    std::transform(lowered.begin(), lowered.end(), lowered.begin(),
+                   [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
+    return lowered.find(GLOBAL_CLUSTER_MARKER) != std::string::npos;
```
yhmo, line 79: src/impl/utils/GlobalClusterUtils.cpp line:79
Low ---- IsGlobalEndpoint() runs on every Connect() in the shared ConnectionHandler and treats any URI containing the substring "global-cluster" (case-insensitive, anywhere in the string) as a global-cluster endpoint. An existing user whose ordinary Milvus endpoint hostname happens to contain that substring (e.g. "https://global-cluster-prod.example.com:19530") will now be routed through the /global-cluster/topology fetch and their previously-working Connect() will fail. Should detection be restricted to the domain pattern actually used by Milvus/Zilliz global-cluster endpoints (e.g. ".global-cluster."), or is the substring match intentionally broad for parity?

src/impl/utils/GlobalClusterUtils.cpp
```diff
@@ -0,0 +1,233 @@
...
+        return {StatusCode::INVALID_ARGUMENT, "Invalid global cluster endpoint: " + endpoint};
+    }
+    // ParseURI defaults a non-https URI without an explicit port to the Milvus gRPC port (19530);
+    // the topology REST API is plain HTTP(S), so use the scheme defaults instead.
+    int port = uri.port;
+    if (uri.scheme == "https") {
+        if (port == 0) {
+            port = 443;
+        }
+    } else if (port == 0 || port == 19530) {
```
yhmo, line 172: src/impl/utils/GlobalClusterUtils.cpp line:172
Low ---- The port fixup `port == 0 || port == 19530` cannot distinguish ParseURI()'s implicit gRPC default (19530 for scheme-less/non-https URIs) from a port the caller explicitly wrote, so an explicit "http://host:19530" topology endpoint is silently rewritten to port 80. Consider only applying the default when the port actually came from ParseURI's default (or dropping the `== 19530` clause and handling the default at the URI-parse level) so an explicit port is honored as-is.

src/impl/utils/TopologyRefresher.cpp
yhmo: fixed

src/impl/utils/ConnectionHandler.cpp
yhmo: fixed

src/impl/utils/GlobalClusterUtils.cpp
yhmo: align with pymilvus, no need to fix

src/impl/utils/GlobalClusterUtils.cpp
yhmo: no need to fix

src/impl/utils/GlobalClusterUtils.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    Status last_status{StatusCode::RPC_FAILED, "Failed to fetch global topology"};
+    for (int attempt = 1; attempt <= MAX_RETRIES; attempt++) {
+        if (should_stop && should_stop()) {
+            return {StatusCode::RPC_FAILED, "Global topology fetch aborted"};
+        }
+
+        httplib::Result res;
+        if (scheme == "https") {
+            httplib::SSLClient client(host, port);
```
yhmo, line 193: src/impl/utils/GlobalClusterUtils.cpp line:193
Low ---- The topology HTTPS fetch creates httplib::SSLClient without carrying over the user's TLS configuration from ConnectParam. The gRPC channel to the primary honors EnableTls(ca_cert)/ServerName via createTlsCredentials(), but FetchTopology() verifies against the system CA bundle only, so a global-cluster deployment using a private/self-signed CA fails at Connect() during the topology fetch even though the primary gRPC connect would succeed. Consider plumbing CaCert/ServerName (or the whole ConnectParam) into FetchTopology and applying set_ca_cert_path()/enable_server_certificate_verification() so the topology fetch and the gRPC channel use the same trust configuration.

yhmo: align with pymilvus

PR #585, opened 2026-08-24
feat: add client telemetry support

xiaofan-luan:
## Summary

- implement the shared client telemetry heartbeat and command protocol in the C++ SDK
- report one metric per public logical Search, Query/Get, HybridSearch, RunAnalyzer, Insert, Delete, and Upsert call
- preserve telemetry state across reconnects, database switches, and global-cluster topology failover within the same logical endpoint/user/config scope
- make heartbeat Stop/restart and last-owner destruction safe, including destruction from the heartbeat callback thread
- keep the command heartbeat active across dynamic disable and Stop/Start reconnects so ACK and re-enable can flow, while an initial user opt-out starts no worker
- retain a bounded one-hour default history, aggregate P99 from retained samples, and skip empty high-frequency history entries
- align deterministic sampling, collection filtering, strict atomic command handling, RFC3339 validation, reply acknowledgement, unsupported backoff, request IDs, unset-database reporting, and command cursor idempotency with the Go SDK
- expose retryable CLIENT_BUSY lifecycle failures and add focused unit plus opt-in end-to-end coverage

## Related work

- Design: https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260131-client_side_telemetry.md
- Server and Go SDK correctness follow-up: https://github.com/milvus-io/milvus/pull/52597
- Go SDK conformance follow-up: https://github.com/milvus-io/milvus/pull/52804

## Verification

- full C++ unit test suite: 842/842 passed
- focused telemetry and global/lifecycle suites passed
- end-to-end test target builds successfully
- LLVM 17 clang-format dry run and git diff --check

## Compatibility note

ConnectParam gains telemetry configuration state. Consumers should rebuild against this SDK revision rather than mixing binaries compiled against the previous object layout.

### DIFF
--- a/src/impl/ClientTelemetry.cpp
+++ b/src/impl/ClientTelemetry.cpp
@@ -1,0 +1,966 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+
+#include "milvus/ClientTelemetry.h"
+
+#include <grpcpp/client_context.h>
+#include <grpcpp/channel.h>
+
+#include <algorithm>
+#include <array>
+#include <atomic>
+#include <cctype>
+#include <cmath>
+#include <condition_variable>
+#include <cstdlib>
+#include <cstring>
+#include <ctime>
+#include <deque>
+#include <iomanip>
+#include <limits>
+#include <map>
+#include <mutex>
+#include <random>
+#include <sstream>
+#include <stdexcept>
+#include <thread>
+#include <unordered_set>
+#include <utility>
+
+#include <google/protobuf/descriptor.h>
+#include <google/protobuf/message.h>
+#include <milvus/thirdparty/nlohmann/json.hpp>
+
+#include "common.pb.h"
+#include "milvus.grpc.pb.h"
+#include "milvus.pb.h"
+
+namespace milvus {
+namespace {
+
+constexpr size_t kSampleBufferSize = 1000;
+constexpr size_t kSnapshotLimit = 120;
+constexpr uint64_t kSamplingDenominator = 10000;
+constexpr size_t kMaxReplyBytes = 1024 * 1024;
+constexpr uint64_t kMaxUnsupportedBackoffMs = 30 * 60 * 1000;
+
+int64_t
+NowMillis() {
+    return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch())
+        .count();
+}
+
+std::string
+RandomUuid() {
+    std::random_device device;
+    std::mt19937_64 generator(device());
+    std::uniform_int_distribution<uint64_t> distribution;
+    auto high = distribution(generator);
+    auto low = distribution(generator);
+    std::ostringstream stream;
+    stream << std::hex << std::setfill('0') << std::setw(8) << static_cast<uint32_t>(high >> 32) << "-"
+           << std::setw(4) << static_cast<uint16_t>(high >> 16) << "-" << std::setw(4)
+           << static_cast<uint16_t>(high) << "-" << std::setw(4) << static_cast<uint16_t>(low >> 48) << "-"
+           << std::setw(12) << (low & 0x0000FFFFFFFFFFFFULL);
+    return stream.str();
+}
+
+std::string
+LocalTimeString() {
+    auto now = std::chrono::system_clock::now();
+    auto value = std::chrono::system_clock::to_time_t(now);
+    std::tm time{};
+#ifdef _WIN32
+    gmtime_s(&time, &value);
+#else
+    gmtime_r(&value, &time);
+#endif
+    std::ostringstream stream;
+    stream << std::put_time(&time, "%Y-%m-%dT%H:%M:%SZ");
+    return stream.str();
+}
+
+std::string
+HostName() {
+    const char* value = std::getenv("HOSTNAME");
+    return value == nullptr ? "Unknown" : value;
+}
+
+std::string
+CollectionName(const google::protobuf::Message& request) {
+    const auto* field = request.GetDescriptor()->FindFieldByName("collection_name");
+    if (field == nullptr || field->cpp_type() != google::protobuf::FieldDescriptor::CPPTYPE_STRING) {
+        return "";
+    }
+    return request.GetReflection()->GetString(request, field);
+}
+
+uint32_t
+RotateRight(uint32_t value, uint32_t count) {
+    return (value >> count) | (value << (32 - count));
+}
+
+// Small self-contained SHA-256 implementation keeps the SDK independent from a specific TLS provider.
+class Sha256 {
+ public:
+    Sha256()
+        : state_{0x6a09e667, 0xbb67ae85, 0x3c6ef372, 0xa54ff53a,
+                 0x510e527f, 0x9b05688c, 0x1f83d9ab, 0x5be0cd19} {
+    }
+
+    void
+    Update(const std::string& value) {
+        Update(reinterpret_cast<const uint8_t*>(value.data()), value.size());
+    }
+
+    std::string
+    Finish() {
+        uint64_t bit_length = total_size_ * 8;
+        buffer_[buffer_size_++] = 0x80;
+        if (buffer_size_ > 56) {
+            while (buffer_size_ < 64) {
+                buffer_[buffer_size_++] = 0;
+            }
+            Transform(buffer_.data());
+            buffer_size_ = 0;
+        }
+        while (buffer_size_ < 56) {
+            buffer_[buffer_size_++] = 0;
+        }
+        for (int index = 7; index >= 0; --index) {
+            buffer_[buffer_size_++] = static_cast<uint8_t>(bit_length >> (index * 8));
+        }
+        Transform(buffer_.data());
+
+        std::ostringstream stream;
+        stream << std::hex << std::setfill('0');
+        for (auto value : state_) {
+            stream << std::setw(8) << value;
+        }
+        return stream.str();
+    }
+
+ private:
+    void
+    Update(const uint8_t* data, size_t size) {
+        total_size_ += size;
+        for (size_t index = 0; index < size; ++index) {
+            buffer_[buffer_size_++] = data[index];
+            if (buffer_size_ == 64) {
+                Transform(buffer_.data());
+                buffer_size_ = 0;
+            }
+        }
+    }
+
+    void
+    Transform(const uint8_t* block) {
+        static const uint32_t constants[64] = {
+            0x428a2f98, 0x71374491, 0xb5c0fbcf, 0xe9b5dba5, 0x3956c25b, 0x59f111f1, 0x923f82a4,
+            0xab1c5ed5, 0xd807aa98, 0x12835b01, 0x243185be, 0x550c7dc3, 0x72be5d74, 0x80deb1fe,
+            0x9bdc06a7, 0xc19bf174, 0xe49b69c1, 0xefbe4786, 0x0fc19dc6, 0x240ca1cc, 0x2de92c6f,
+            0x4a7484aa, 0x5cb0a9dc, 0x76f988da, 0x983e5152, 0xa831c66d, 0xb00327c8, 0xbf597fc7,
+            0xc6e00bf3, 0xd5a79147, 0x06ca6351, 0x14292967, 0x27b70a85, 0x2e1b2138, 0x4d2c6dfc,
+            0x53380d13, 0x650a7354, 0x766a0abb, 0x81c2c92e, 0x92722c85, 0xa2bfe8a1, 0xa81a664b,
+            0xc24b8b70, 0xc76c51a3, 0xd192e819, 0xd6990624, 0xf40e3585, 0x106aa070, 0x19a4c116,
+            0x1e376c08, 0x2748774c, 0x34b0bcb5, 0x391c0cb3, 0x4ed8aa4a, 0x5b9cca4f, 0x682e6ff3,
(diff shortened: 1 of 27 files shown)


src/include/milvus/MilvusClientV2.h
```diff
@@ -18,6 +18,7 @@
 
 #include <functional>
 
+#include "ClientTelemetry.h"
```
yhmo, line 22: src/include/milvus/MilvusClientV2.h line:21
Low ---- The repository's aggregate-header convention (see technical-details.md and the "one header" rule) is that a consumer can include only `milvus/MilvusClientV2.h` and still reach every public type. `ClientTelemetry.h` is added here, but the new `ClientRequestContext` / `ScopedClientRequestId` API -- which this PR wires into `grpcCall` as the `client_request_id` metadata and demonstrates in the E2E -- is not reachable from the aggregate: `ClientTelemetry.h`, `ConnectParam.h` and `MilvusClientV2.h` never include `ClientRequestContext.h`, and the same gap exists in the V1 `MilvusClient.h`. A consumer following the one-header pattern who wants to set a per-request ID must discover and add a second include or the code will not compile. Suggest adding `#include "ClientRequestContext.h"` to the aggregate headers.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -22,30 +22,65 @@ namespace milvus {
...
 }
 
 Status
 ConnectionHandler::Disconnect() {
-    std::lock_guard<std::mutex> lock(mtx_);
-    if (connection_ != nullptr) {
-        return connection_->Disconnect();
+    std::unique_lock<std::mutex> lifecycle_lock(lifecycle_mtx_, std::try_to_lock);
+    if (!lifecycle_lock.owns_lock()) {
+        return {StatusCode::UNKNOWN_ERROR, "Connection lifecycle change is already in progress"};
```
yhmo: src/impl/utils/ConnectionHandler.cpp line:69
Low ---- Question: this PR changes the concurrency semantics of four public lifecycle APIs -- `Disconnect`, `UseDatabase`, `SetRpcDeadlineMs`, `SetRetryParam` -- from blocking on the connection lock to failing fast with `UNKNOWN_ERROR: Connection lifecycle change is already in progress` whenever another thread holds the lifecycle lock. The try-lock itself is a reasonable way to avoid the new deadlock window (a custom command handler running on the heartbeat thread calling back into the client), but the caller-visible behavior change is not documented anywhere on the public API. A caller that previously serialized concurrent disconnect / use-database calls now receives a transient error it cannot distinguish from a real failure. Could you confirm this is intended, and consider returning a more specific/retryable status (or documenting the fail-fast contract) so concurrent callers can distinguish "retry" from "failed"?

src/include/milvus/ClientTelemetry.h
```diff
@@ -0,0 +1,168 @@
...
+    ClientTelemetryManager&
+    operator=(const ClientTelemetryManager&) = delete;
+
+    void
+    AttachChannel(const std::shared_ptr<grpc::Channel>& channel, const std::string& username,
+                  const std::string& database, const std::string& uri, const std::string& sdk_version,
+                  const std::string& connection_scope = "");
+
+    void
+    UpdateDatabase(const std::string& database);
```
yhmo: src/include/milvus/ClientTelemetry.h line:100
Low ---- `UpdateDatabase` is a new public method with no callers anywhere on this branch; the reported database name is only refreshed through `AttachChannel` on the reconnect/`UseDatabase` path, never through this method. If it is meant to support live DB switches it is currently dead code (and drift-prone, since a future caller may assume it updates what the heartbeat reports); otherwise consider removing it or documenting it as reserved for future use so the public surface does not imply behavior that is not wired up.

test/CMakeLists.txt
```diff
@@ -65,6 +65,11 @@ add_executable(testing-it ${it_files})
 target_compile_options(testing-it PRIVATE $<$<CXX_COMPILER_ID:MSVC>:/bigobj>)
 link_milvus_test(testing-it)
 
+if (CMAKE_SYSTEM_NAME MATCHES "(Linux|Darwin)")
+    add_executable(testing-telemetry-e2e e2e/ClientTelemetryE2E.cpp)
```
yhmo, line 80: test/CMakeLists.txt line:69
Low ---- `testing-telemetry-e2e` is built on Linux/macOS but no CI workflow runs it (`main.yaml` only runs `make test`), so the client<->server wire contract -- heartbeat registration, command round-trips, the `applied`/`ignored` config reply, and the on-wire sampling distribution -- is only validated when someone runs the binary manually against a live cluster. This is exactly the gap milvus#52597 closed for the Go SDK by replacing rotted telemetry examples with a CI-run integration suite, because unit tests cannot reach the wire. Non-blocking, but wiring this into CI (or a documented CI step) would keep the cross-SDK conformance contract continuously validated.

src/impl/ClientTelemetry.cpp
```diff
@@ -0,0 +1,1185 @@
...
+        std::lock_guard<std::mutex> lock(mutex);
+        if (ready) {
+            return;
+        }
+        ready = true;
+        stopped = false;
+        if (!config.enabled) {
+            return;
+        }
+        worker = std::thread([this]() { HeartbeatLoop(); });
```
yhmo: src/impl/ClientTelemetry.cpp line:499
High ---- Narrower remaining case after the Start/Stop rework: a command handler that destroys the client itself (dropping the last manager reference on the heartbeat worker thread) still terminates. `Stop()` recognizes `called_from_worker`, sets `stopped` and returns without joining, so `~Impl` proceeds to destroy the `worker` std::thread member while it is still joinable and represents the current thread -- `std::thread::~thread()` then calls `std::terminate()` (and the thread that returns into the freed handler frame is use-after-free). The new tests cover self-restart via `UseDatabase` and the external-stop race, but not handler-triggered client destruction. Guarding the destructor (e.g. detaching/marking the worker so a non-worker owner joins it, or refusing self-destruction until the handler unwinds) would close the last path.

src/impl/ClientTelemetry.cpp
```diff
@@ -0,0 +1,1168 @@
...
+        {
+            std::lock_guard<std::mutex> lock(mutex);
+            if (!ready && stopped) {
+                return;
+            }
+            stopped = true;
+            condition.notify_all();
+        }
+        if (worker.joinable() && worker.get_id() != std::this_thread::get_id()) {
+            worker.join();
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/ClientTelemetry.cpp
```diff
@@ -0,0 +1,1168 @@
...
+#include "common.pb.h"
+#include "milvus.grpc.pb.h"
+#include "milvus.pb.h"
+#include "milvus/ClientRequestContext.h"
+
+namespace milvus {
+namespace {
+
+constexpr size_t kSampleBufferSize = 1000;
+constexpr size_t kSnapshotLimit = 120;
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/ClientTelemetry.cpp
```diff
@@ -0,0 +1,1168 @@
...
+                };
+                std::map<std::string, Total> totals;
+                for (const auto& snapshot : selected) {
+                    for (const auto& operation : snapshot.metrics) {
+                        auto& total = totals[operation.operation];
+                        total.requests += operation.global.request_count;
+                        total.successes += operation.global.success_count;
+                        total.errors += operation.global.error_count;
+                        total.average += operation.global.avg_latency_ms * operation.global.request_count;
+                        total.p99 += operation.global.p99_latency_ms * operation.global.request_count;
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/include/milvus/MilvusClientV2.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/ConnectionHandler.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/include/milvus/ClientTelemetry.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/ClientTelemetry.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/ConnectionHandler.cpp
```diff
@@ -32,19 +32,34 @@ ConnectionHandler::~ConnectionHandler() {
 
 Status
 ConnectionHandler::Connect(const ConnectParam& connect_param) {
+    // Keep the candidate private until its handshake succeeds, but do not hold mtx_ while
+    // AttachChannel waits for an in-flight command handler. The lifecycle lock also fences
+    // concurrent explicit connects, disconnects, database switches, and global failovers.
+    std::lock_guard<std::mutex> lifecycle_lock(lifecycle_mtx_);
```
yhmo: src/impl/utils/ConnectionHandler.cpp line:38
Low ---- Question: the other lifecycle methods (`Disconnect`, `UseDatabase`, `SetRpcDeadlineMs`, `SetRetryParam`) now take `lifecycle_mtx_` with `try_to_lock` and fail fast with CLIENT_BUSY, but `Connect()` still takes it with a blocking `lock_guard` and then calls `MilvusConnection::Connect()` -> `telemetry->AttachChannel()`, which blocks on `command_mutex` (held across `ProcessCommands` in `ClientTelemetry.cpp`). If a custom command handler running on the heartbeat worker calls `client->Connect()` while another thread concurrently calls `Connect()`, that is a deadlock: the external thread holds `lifecycle_mtx_` and waits for `command_mutex`, while the worker holds `command_mutex` and waits for `lifecycle_mtx_`. Since this PR already tests `UseDatabase` from a handler, handler-initiated reconnects are an anticipated flow; consider giving `Connect()` the same `try_to_lock` treatment (returning CLIENT_BUSY on contention) or documenting that command handlers must not call `Connect()` directly.

src/impl/MilvusConnection.h
```diff
@@ -539,6 +555,11 @@ class MilvusConnection {
         }
 
         ::grpc::ClientContext context;
+        const std::string& contextual_request_id = ClientRequestContext::Get();
+        const std::string request_id = options.request_id.empty() ? contextual_request_id : options.request_id;
+        if (ClientRequestContext::IsValid(request_id)) {
```
yhmo, line 560: src/impl/MilvusConnection.h line:560
Low ---- The `client_request_id` metadata injection here is the on-wire contract that makes per-request correlation work, but no test asserts a server actually receives it. The E2E only proves the request ID reaches `show_errors` (via `ClientRequestContext::Get()` recorded into telemetry errors), which is independent of this gRPC metadata path, and the mocked heartbeat tests never inspect `ServerContext::client_metadata`. Consider adding a mocked/UT assertion that a request carrying a valid `ScopedClientRequestId`/`ClientRequestContext` value is visible in the server-side metadata (and that invalid/empty IDs are not), so this path cannot silently regress.

src/impl/ClientTelemetry.cpp
```diff
@@ -0,0 +1,1315 @@
...
+}
+
+std::string
+RandomUuid() {
+    std::random_device device;
+    std::mt19937_64 generator(device());
+    std::uniform_int_distribution<uint64_t> distribution;
+    auto high = distribution(generator);
+    auto low = distribution(generator);
+    std::ostringstream stream;
```
yhmo, line 104: src/impl/ClientTelemetry.cpp line:102
Low ---- This file uses std::ostringstream / std::put_time / std::setw / std::setfill / std::hex (RandomUuid at line 102, LocalTimeString at line 119, and Sha256::Finish at line 183) but does not directly include <sstream> or <iomanip>; it compiles only through transitive includes pulled in by the gRPC/protobuf and nlohmann/json headers. Adding `#include <sstream>` and `#include <iomanip>` to the include block (lines 16-32) would make the include set self-contained and avoid breakage if a transitive header stops providing them.

src/impl/MilvusClientImpl.cpp
```diff
@@ -39,7 +40,27 @@ namespace milvus {
 
 std::shared_ptr<MilvusClient>
 MilvusClient::Create() {
-    return std::make_shared<MilvusClientImpl>();
+    return std::shared_ptr<MilvusClient>(new MilvusClientImpl(), [](MilvusClientImpl* client) noexcept {
```
yhmo: src/impl/MilvusClientImpl.cpp line:43
Low ---- The custom deleter in `MilvusClient::Create()` (lines 43-64) is byte-for-byte identical to the one added in `MilvusClientV2::Create()` (MilvusClientV2Impl.cpp lines 46-67) except for the concrete impl type and comment. Both defer deletion to a detached cleanup thread when the telemetry worker is the current thread, so the destructor never joins a global topology refresher from inside a command handler. This subtle threading/ownership logic is now duplicated and drift-prone: a future fix to one copy (e.g. the fallback when thread creation fails) could be missed in the other. Consider extracting a shared helper, e.g. `DeferTelemetryClientDeletion(const ClientTelemetryManagerPtr&, T* client)`, and calling it from both factories.

src/impl/utils/ConnectionHandler.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusConnection.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/ClientTelemetry.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientImpl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

test/CMakeLists.txt
yhmo: ignore

PR #587, opened 2026-08-25
fix: treat ttl_seconds as an ignored push_config field

xiaofan-luan:
## Summary
- `ttl_seconds` is a field of the server-side push API (`PushClientCommandRequest` / `POST /_telemetry/commands`), not of `ClientCommand`, so it never reaches clients by design (see the TTL section of the design doc).
- Stop type-validating it in the `push_config` command handler and report it as ignored like any other unknown field, so a stray value can no longer fail the whole command.
- Update `RejectsWrongCommandPayloadTypes` accordingly: an invalid `ttl_seconds` now succeeds with `ignored: ["ttl_seconds"]` and leaves the applied config untouched.

This is a follow-up to #585 (the change was pushed to the head branch shortly after merge and did not land).

pr: #585

## Related work
- Design: https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260131-client_side_telemetry.md

## Verification
- focused unit tests: `ClientTelemetryTest.*` 26/26 passed
- full unit suite rebuilt against latest master

### DIFF
--- a/test/ut/TestClientTelemetry.cpp
+++ b/test/ut/TestClientTelemetry.cpp
@@ -1019,13 +1019,16 @@ TEST(ClientTelemetryTest, RejectsWrongCommandPayloadTypes) {
     manager.ProcessCommands(
         {{"push", "push_config", R"({"enabled":"false"})", 1, false, ""},
          {"collection", "collection_metrics", R"({"enabled":false,"collections":"books"})", 2, false, ""},
-         {"ttl", "push_config", R"({"enabled":true,"ttl_seconds":"bad"})", 3, false, ""}});
+         {"ttl", "push_config", R"({"ttl_seconds":"bad"})", 3, true, ""}});
 
     auto replies = manager.PendingCommandReplies();
     ASSERT_EQ(replies.size(), 3U);
     EXPECT_FALSE(replies[0].success);
     EXPECT_FALSE(replies[1].success);
-    EXPECT_FALSE(replies[2].success);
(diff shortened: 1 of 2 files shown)


test/ut/TestClientTelemetry.cpp
```diff
@@ -1019,13 +1019,16 @@ TEST(ClientTelemetryTest, RejectsWrongCommandPayloadTypes) {
     manager.ProcessCommands(
         {{"push", "push_config", R"({"enabled":"false"})", 1, false, ""},
          {"collection", "collection_metrics", R"({"enabled":false,"collections":"books"})", 2, false, ""},
-         {"ttl", "push_config", R"({"enabled":true,"ttl_seconds":"bad"})", 3, false, ""}});
+         {"ttl", "push_config", R"({"ttl_seconds":"bad"})", 3, true, ""}});
```
yhmo, line 1022: test/ut/TestClientTelemetry.cpp line:1022
Low ---- The `persistent` flag on this command changed from `false` to `true` (with `enabled` removed from the payload) without a comment. Since `ProcessCommands` computes `CalculateConfigHash` for persistent commands, this hash now incorporates the `{"ttl_seconds":"bad"}` payload. The test does not assert the config hash, so this is harmless here, but could you confirm whether the flag change is intentional (e.g. to mirror a real persistent `push_config`) and, if so, worth a brief note in the test?

yhmo:
this pr is merged into https://github.com/milvus-io/milvus-sdk-cpp/pull/601

PR #588, opened 2026-08-25
fix: harden client telemetry control flow

xiaofan-luan:
## Summary\n\n- make very large heartbeat intervals safe and cancellable across platforms\n- fence command batches across telemetry transport rebinding and canonicalize ACK IDs\n- wake unsupported backoff on channel replacement and bound show_errors truncation, including UTF-8 payloads\n\nFollow-up hardening for #585.\n\n## Validation\n\n- 854/854 unit tests passed\n- 291/291 mocked integration tests passed\n- final telemetry-focused suite: 33/33 passed\n- clang-format and git diff --check passed\n

### DIFF
--- a/src/impl/ClientTelemetry.cpp
+++ b/src/impl/ClientTelemetry.cpp
@@ -68,12 +68,18 @@ constexpr size_t kSnapshotLimit = 4096;
 constexpr uint64_t kSamplingScale = 1000000000ULL;
 constexpr size_t kMaxReplyBytes = 1024 * 1024;
 constexpr uint64_t kMaxUnsupportedBackoffMs = 30 * 60 * 1000;
+constexpr uint64_t kMaxHeartbeatIntervalMs = static_cast<uint64_t>(std::numeric_limits<int64_t>::max());
+// condition_variable implementations commonly form an absolute deadline internally.
+// Keep each addition comfortably inside every supported platform clock and represent
+// longer intervals as stop/rebind-aware chunks.
+constexpr uint64_t kMaxWaitChunkMs = 24ULL * 60 * 60 * 1000;
 
 TelemetryConfig
 NormalizedTelemetryConfig(TelemetryConfig config) {
     if (config.heartbeat_interval_ms == 0) {
         config.heartbeat_interval_ms = 10000;
     }
+    config.heartbeat_interval_ms = std::min(config.heartbeat_interval_ms, kMaxHeartbeatIntervalMs);
     config.sampling_rate = std::max(0.0, std::min(1.0, config.sampling_rate));
     if (config.error_max_count == 0) {
         config.error_max_count = 100;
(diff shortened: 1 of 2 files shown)


src/impl/ClientTelemetry.cpp
```diff
@@ -612,7 +656,16 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
                 }
                 delay = std::max(delay, backed_off);
             }
-            condition.wait_for(lock, std::chrono::milliseconds(delay), [this]() { return stopped; });
+            const auto observed_revision = wait_revision;
+            while (!stopped && observed_revision == wait_revision && delay > 0) {
+                const auto chunk = std::min(delay, kMaxWaitChunkMs);
+                if (condition.wait_for(
+                        lock, std::chrono::milliseconds(static_cast<int64_t>(chunk)),
+                        [this, observed_revision]() { return stopped || wait_revision != observed_revision; })) {
```
yhmo, line 664: src/impl/ClientTelemetry.cpp line:664
Medium ---- A server-pushed heartbeat_interval_ms change cannot wake the new chunked wait: the wait predicate only reacts to `stopped` or `wait_revision`, but the push_config handler calls condition.notify_all() without bumping wait_revision. With this PR's INT64_MAX interval support chunked into 24h waits, a corrective push (e.g. back to 10s) is deferred until the current chunk elapses. Consider bumping wait_revision in the push_config handler so an interval re-push is applied on the next loop iteration.

src/impl/ClientTelemetry.cpp
```diff
@@ -94,6 +100,27 @@ NowMillis() {
         .count();
 }
 
+int64_t
+SaturatingWindowStart(int64_t end_time, uint64_t interval_ms) {
+    const auto signed_interval = static_cast<int64_t>(std::min(interval_ms, kMaxHeartbeatIntervalMs));
+    if (end_time < std::numeric_limits<int64_t>::min() + signed_interval) {
+        return std::numeric_limits<int64_t>::min();
```
yhmo, line 107: src/impl/ClientTelemetry.cpp line:107
Low ---- SaturatingWindowStart saturates to INT64_MIN at the maximum heartbeat interval, and that value is later serialized verbatim as the snapshot "timestamp" in show_latency_history replies. A consumer expecting an epoch-millis timestamp receives -9223372036854775808. Consider capping the window start at the retained-history floor (e.g. end_time - kMaxHistoryRangeMs) so the reported timestamp stays interpretable.

test/ut/TestClientTelemetry.cpp
```diff
@@ -1011,6 +1306,45 @@ TEST(ClientTelemetryTest, PushConfigIsAtomicAndReportsAppliedAndIgnoredKeys) {
     EXPECT_EQ(payload["ignored"], nlohmann::json({"ttl_seconds", "unknown_a", "unknown_b"}));
 }
 
+TEST(ClientTelemetryTest, ShowErrorsBoundsOversizedNonMessageFields) {
```
yhmo, line 1309: test/ut/TestClientTelemetry.cpp line:1309
Low ---- The show_errors truncation tests cover a single oversized field, but the selection loop that alternates across the four truncatable fields and the "still exceeds 1MB after all fields are emptied" throw path are only reachable when one retained error carries multiple oversized fields (e.g. both a huge collection and a huge error_msg). A test with two oversized fields would exercise the next-longest-field selection and the monotonic-progress fallback.

yhmo:
this pr is merged into https://github.com/milvus-io/milvus-sdk-cpp/pull/601

PR #589, opened 2026-08-25
fix: treat ttl_seconds as an ignored push_config field

xiaofan-luan:
## What changed

- Treat ttl_seconds as an ignored client command field instead of validating its type.
- Keep the existing ignored-key response semantics for stray payload fields.
- Update the focused telemetry test.

## Why

ttl_seconds belongs to the server-side PushClientCommandRequest API and is not part of the ClientCommand contract delivered to SDKs. A stray field should therefore be reported as ignored rather than causing the full push_config command to fail.

Follow-up to #585.

## Validation

- cmake --build build --target testing-ut -j 8
- build/test/testing-ut --gtest_filter=ClientTelemetryTest.RejectsWrongCommandPayloadTypes --gtest_color=no
- git diff --check upstream/master..HEAD

### DIFF
--- a/src/impl/ClientTelemetry.cpp
+++ b/src/impl/ClientTelemetry.cpp
@@ -884,12 +884,6 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
                 sampling_rate = std::max(0.0, std::min(1.0, sampling_rate));
                 applied.emplace_back("sampling_rate");
             }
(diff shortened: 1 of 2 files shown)


xiaofan-luan:
Closing as an exact duplicate of #587. Both PRs have the same base commit and identical resulting tree; continuing with #587, which is already approved.

PR #590, opened 2026-08-27
Speed up builds with unity build and line-tables debug info

yhmo:
## What this PR does

  This PR adds opt-in build-time optimizations and speeds up the coverage pipeline, with zero impact
  on the SDK's runtime behavior, public API, or default build configuration.

  ### 1. Unity build + line-tables debug info switches

  Two environment switches, combinable with any make target that builds the SDK (make, make test,
  make coverage, make package), both default off:

  - UNITY=ON — enables CMake unity builds for the SDK object library. The 220+ small request/response
    translation units pay their shared header closure once per batch instead of once per file.
    Generated protobuf sources are always excluded (the skip is set at the generation site in
    MilvusProtoGen.cmake, so future protos inherit it automatically).

  - LINE=ON — maps to -gline-tables-only (Clang) / -g1 (GCC): keeps stack traces and file:line
    debug info while skipping variable-level DWARF. No effect on code generation; Release builds are
    unaffected (they carry no debug info anyway).

  Unity-safety source cleanups shipped along: deduplicated anonymous-namespace constants and
  little-endian writers (now shared via src/impl/utils/LittleEndian.h), and removed a duplicate
  explicit template instantiation. Validated by building with MILVUS_UNITY_BATCH_SIZE=300 —
  all 220 sources in a single batch — which compiles cleanly, so every accepted batch size works.

  ### 2. Coverage pipeline acceleration

  - Examples are no longer built for coverage runs (they are filtered out of the report anyway)
  - COVERAGE_HTML=OFF skips the genhtml step; codecov only consumes lcov_output.info
  - Parallel lcov capture (COVERAGE_JOBS): lcov runs gcov sequentially per object; data files
    are now balanced across workers at file granularity and merged afterwards

  - LINE=ON is safe for coverage builds — validated by byte-identical filtered lcov output
    between full and line-tables debug info (gcov reads .gcno/.gcda notes, not DWARF)

  - Lint builds force UNITY=OFF in build.sh (mirroring the existing PCH guard) so clang-tidy
    keeps per-file compile commands

  ### 3. CI wiring

  - The Linux matrix (Ubuntu 20.04 / Fedora 39) builds **Release** with `UNITY=ON` so lint, tests,
  and the tutorials share one Release Conan dependency stack (`LINE` only affects Debug/
  RelWithDebInfo, so it is unset there). Windows tests the same Release configuration with unity.
- macOS keeps the Debug build with `UNITY=ON LINE=ON`.
- The coverage job keeps `UNITY=OFF` (per-file gcov/lcov attribution) with `LINE=ON`, an 8G ccache
  budget, and skips genhtml; it also exercises the full UT/IT/ST suite in Debug on Linux.
- Release packaging (release.yaml) deliberately stays with the default `UNITY=OFF` so official
  artifacts match the default compilation mode; it only skips examples.

  All new switches are validated at startup with actionable errors.

  ## Measured results

  Local cold builds (JOBS=4, ccache disabled, Debug, SDK + examples):

   Configuration    Wall time    vs baseline
  ━━━━━━━━━━━━━━━  ━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   Baseline         597 s        —
  ───────────────  ───────────  ────────────────────────────────
   UNITY=ON         256 s        −57%
  ───────────────  ───────────  ────────────────────────────────
   LINE=ON          385 s        −36%, peak RSS −28%, disk −47%
  ───────────────  ───────────  ────────────────────────────────
   Both             227 s        −62% (2.6x)

  Local coverage report step: lcov pipeline 5m11s → 1m08s (4.6x), output byte-identical.

  CI (vs pre-optimization baselines from earlier this month):

   Metric                  Before           After      Change
  ━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━  ━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━
   Total (5 build jobs)    ~17.7–18.0k s    13.5k s    −22…−25%
  ──────────────────────  ───────────────  ─────────  ──────────────────────────
   Coverage job            4.1–4.2k s       1.2k s     −71% (testing step −78%)
  ──────────────────────  ───────────────  ─────────  ──────────────────────────
   Ubuntu lint step        4.1–4.5k s       2.9k s     −30…−35%
  ──────────────────────  ───────────────  ─────────  ──────────────────────────
   macOS job               2.9–3.4k s       2.0k s     −32…−42%

  All gains were measured despite a month of codebase growth in between, and all test suites
  (865 unit + 291 mocked-integration tests) pass on every platform.

  ## Safety

  - No public headers, API, or ABI changes; the installed library is byte-for-byte identical by
    default

  - Defaults keep current behavior: UNITY=OFF, LINE=full, unchanged flags
  - Line-tables only changes DWARF emission; unity only groups TUs (no cross-TU optimization)
  - Coverage output validated identical; unity+coverage deliberately avoided (attribution risk) with
    a
    build-time warning


src/CMakeLists.txt
```diff
@@ -32,6 +32,25 @@ endif()
 
 # add proto gens
 add_milvus_protos(milvus_sdk_objects)
+
+if (MILVUS_ENABLE_UNITY)
+    # Generated protobuf sources are huge standalone TUs (milvus.grpc.pb.cc is the slowest
+    # file in the build); batching them with hand-written code would amplify memory pressure
+    # and can make the build slower. They also rely on standalone-TU semantics, so never
+    # include them in unity batches.
+    set_source_files_properties(
```
yhmo: src/CMakeLists.txt line:41
Medium ---- Generated protobuf sources are excluded from unity builds by repeating the full generated-file inventory here, while `add_milvus_protos(...)` in `cmake/MilvusProtoGen.cmake` already owns that exact list. A future proto can therefore be generated but not marked as skipped unless both sites are updated, letting it enter a unity batch contrary to the stated invariant. Please set `SKIP_UNITY_BUILD_INCLUSION` where each generated output is added, or share one helper/inventory between generation and exclusion.

CMakeLists.txt
```diff
@@ -63,6 +63,15 @@ define_option(MILVUS_BUILD_TEST "Build with testing" OFF)
 define_option(MILVUS_BUILD_COVERAGE "Build with coverage" OFF)
 define_option(MILVUS_BUILD_EXAMPLES "Build examples" ON)
 define_option(MILVUS_ENABLE_PCH "Use precompiled headers when building the SDK" ON)
+define_option(MILVUS_ENABLE_UNITY "Use CMake unity builds for the SDK object library" OFF)
+define_option_string(MILVUS_UNITY_BATCH_SIZE "CMake unity build batch size" "8")
+define_option_string(MILVUS_DEBUG_INFO "Debug information level" "full" "full" "line-tables")
+
+if (MILVUS_ENABLE_UNITY AND CMAKE_VERSION VERSION_LESS 3.16.1)
```
yhmo, line 70: CMakeLists.txt line:70
Low ---- CMake documents `SKIP_UNITY_BUILD_INCLUSION` as added in 3.16, so requiring a nonempty `VERSION_GREATER_EQUAL "3.16"` via `NOT VERSION_LESS "3.16"` would be more precise.

CMakeLists.txt
```diff
@@ -63,6 +63,15 @@ define_option(MILVUS_BUILD_TEST "Build with testing" OFF)
 define_option(MILVUS_BUILD_COVERAGE "Build with coverage" OFF)
 define_option(MILVUS_BUILD_EXAMPLES "Build examples" ON)
 define_option(MILVUS_ENABLE_PCH "Use precompiled headers when building the SDK" ON)
+define_option(MILVUS_ENABLE_UNITY "Use CMake unity builds for the SDK object library" OFF)
+define_option_string(MILVUS_UNITY_BATCH_SIZE "CMake unity build batch size" "8")
+define_option_string(MILVUS_DEBUG_INFO "Debug information level" "full" "full" "line-tables")
```
yhmo, line 68: CMakeLists.txt line:68
Low ---- `define_option_string` records cache choices for the UI but does not reject invalid command-line values. A typo such as `MILVUS_DEBUG_INFO=line_tables` silently takes the full-debug path, while an unexpected `MILVUS_UNITY_BATCH_SIZE` can surface only later as a CMake property error. Please validate that the debug value is exactly one of the supported modes and that the batch size is a positive integer during configure.

.github/workflows/main.yaml
```diff
@@ -87,6 +87,8 @@ jobs:
       CCACHE_COMPRESSLEVEL: 5
       CCACHE_MAXSIZE: 2G
       BUILD_SHARED_LIBS: ${{ matrix.os.build_shared_libs }}
+      UNITY: ON
```
yhmo, line 93: .github/workflows/main.yaml line:90
Medium ---- The job-level `UNITY: ON` / `LINE: ON` env vars also apply to the `Lint` step, which runs `make lint` -> `scripts/build.sh -l`. That build configures with `MILVUS_ENABLE_UNITY=ON`, so `compile_commands.json` then contains entries for the generated unity batch files rather than the individual sources. `run_clang_tidy.py` passes each individual source file to clang-tidy with `-p compile_commands.json`; with no per-file compile command available, clang-tidy either errors or silently checks nothing. The coverage job already forces `UNITY: OFF` / `LINE: OFF` for per-file attribution reasons; the lint step needs the same treatment (or UNITY/LINE should be scoped to the Unit Testing step only).

scripts/build.sh
```diff
@@ -29,12 +29,34 @@ MILVUS_SDK_VERSION=${MILVUS_SDK_VERSION:-}
 DO_INSTALL="OFF"
 CMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX:-/usr/local}
 BUILD_SHARED_LIBS=${BUILD_SHARED_LIBS:-ON}
+UNITY=${UNITY:-OFF}
+LINE=${LINE:-OFF}
+MILVUS_UNITY_BATCH_SIZE=${MILVUS_UNITY_BATCH_SIZE:-8}
```
yhmo, line 34: scripts/build.sh line:34
Low ---- `UNITY` and `LINE` are validated above, but `MILVUS_UNITY_BATCH_SIZE` is passed straight to CMake. A non-numeric value (e.g. `MILVUS_UNITY_BATCH_SIZE=abc`) fails at `set_target_properties(... UNITY_BUILD_BATCH_SIZE "abc")` with an opaque configure error. Consider validating it as a positive integer alongside the ON/OFF checks.

CMakeLists.txt
```diff
@@ -144,6 +153,21 @@ endif()
...
 include(MilvusProtoGen)
 
+# Line-tables-only debug info keeps stack traces and file:line while skipping
+# variable-level DWARF, which cuts cold-build time and object size. gcov line
+# and function coverage data comes from the .gcno/.gcda notes files (generated
+# by -fprofile-arcs -ftest-coverage), not from DWARF, so this is safe for
+# coverage builds too (validated by identical lcov output).
+if (MILVUS_DEBUG_INFO STREQUAL "line-tables")
+    if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+        add_compile_options($<$<CONFIG:Debug,RelWithDebInfo>:-gline-tables-only>)
```
yhmo: CMakeLists.txt line:163
Medium ---- CMake 3.16 (the minimum declared by `cmake_minimum_required` and README) documents only `$<CONFIG:cfg>`; comma-separated config matching begins in 3.19. Consequently, valid `LINE=ON` Debug/RelWithDebInfo builds on CMake 3.16-3.18 silently evaluate this condition false and retain default full debug info, contradicting the new option. Please use a 3.16-compatible expression such as `$<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:...>` in both compiler branches, or explicitly raise the required CMake version for this feature.

DEVELOPMENT.md
```diff
@@ -86,6 +87,29 @@ $ bash scripts/build.sh -v v3.0.2 -t Release
 ```
 `MILVUS_SDK_VERSION` is embedded into the SDK version string generated from `src/impl/version.h.in`. If you do not pass `MILVUS_SDK_VERSION` or `scripts/build.sh -v`, CMake derives the version from the latest reachable git tag. For release builds, pass the intended version explicitly or build from the release tag.
 
+## Speed up local builds
+
+Two make-level switches reduce cold-build wall time without changing runtime behavior. They can
+be combined with any make target:
```
yhmo: DEVELOPMENT.md line:93
Low ---- "any make target" is broader than the implementation: `doc`, `clean`, `tutorials`, and the `run` wrappers do not invoke `scripts/build.sh`, so `UNITY` and `LINE` have no effect there. Please narrow this to build targets that configure the SDK (or list the relevant targets) so users do not expect every Makefile target to honor these switches.

scripts/coverage.sh
```diff
@@ -18,6 +18,10 @@
 
 BUILD_OUTPUT_DIR="cmake_build"
 COVERAGE_OUTPUT_DIR="code_coverage"
+# Set COVERAGE_HTML=OFF to skip the genhtml step (lcov_output.info is still
+# generated). CI uses this: codecov only consumes the .info tracefile and the
+# HTML tree dies with the ephemeral runner anyway. Local builds keep HTML.
+COVERAGE_HTML=${COVERAGE_HTML:-ON}
```
yhmo, line 24: scripts/coverage.sh line:24
Low ---- This new switch recognizes only literal `ON`; malformed or common alternate values such as `YES`, `TRUE`, or `0` fall through the later `else` and silently skip the HTML report. Since build.sh validates its ON/OFF switches at startup, please use the same `case` validation here so typos fail immediately with an actionable message.

src/CMakeLists.txt
yhmo: Fixed. `SKIP_UNITY_BUILD_INCLUSION` is now set in `cmake/MilvusProtoGen.cmake` right where each generated file is registered (`add_proto_source`/`add_proto_service`), so the generation site owns the exclusion and a future proto can never leak into a unity batch. The duplicated inventory in `src/CMakeLists.txt` is removed.

CMakeLists.txt
yhmo: Acknowledged. I kept the conservative `3.16.1` guard deliberately: I could not confirm whether `SKIP_UNITY_BUILD_INCLUSION` exists in the 3.16.0 patch release, and rejecting a slightly older patch version is harmless, while silently ignoring the property would let generated sources enter unity batches. Happy to relax the check to plain `3.16` (or drop the guard) if maintainers confirm 3.16.0 carries the property.

CMakeLists.txt
yhmo: Fixed. Configure now fails fast with `FATAL_ERROR` unless `MILVUS_DEBUG_INFO` is exactly `full` or `line-tables` and `MILVUS_UNITY_BATCH_SIZE` is a positive integer, so typos surface immediately instead of silently taking the full-debug path.

.github/workflows/main.yaml
yhmo: Fixed. The Lint step now sets a step-level `UNITY: OFF` override, so the lint tree is configured one file per TU and `compile_commands.json` keeps per-source entries for `run_clang_tidy.py`. `LINE` stays enabled for lint (it only affects DWARF emission and speeds up the lint build), and Unit Testing still inherits the job-level `UNITY`/`LINE`. This should also remove the lint-step regression observed on the Ubuntu job.

scripts/build.sh
yhmo: Fixed. `build.sh` validates `MILVUS_UNITY_BATCH_SIZE` as a positive integer alongside the existing `UNITY`/`LINE` checks, failing with an actionable message before CMake runs.

CMakeLists.txt
yhmo: Fixed. Both compiler branches now use `$<$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>:...>`, which CMake 3.16 supports, so `LINE=ON` no longer silently no-ops on CMake 3.16-3.18.

DEVELOPMENT.md
yhmo: Fixed. The docs now list the targets that configure the SDK through `scripts/build.sh` (`make`, `make test`, `make coverage`, `make package`) and explicitly note that `doc`, `clean`, `tutorials`, and the `run` wrappers are unaffected.

scripts/coverage.sh
yhmo: Fixed. `coverage.sh` validates `COVERAGE_HTML` with the same `ON|OFF` case pattern used by `build.sh`, so values such as `YES`, `TRUE`, or `0` fail immediately with a clear message.

scripts/coverage.sh
```diff
@@ -18,6 +18,23 @@
...
+# lcov 1.x runs gcov sequentially per object file, which makes capture the
+# slowest part of the report pipeline. Capture chunks are balanced at data-file
+# granularity (a single large object cannot pin the critical path) and merged
+# afterwards; the filtered output is byte-identical to the sequential run.
+COVERAGE_JOBS=${COVERAGE_JOBS:-$(nproc 2>/dev/null || sysctl -n hw.logicalcpu 2>/dev/null || echo 4)}
```
yhmo, line 29: scripts/coverage.sh line:29
Low ---- COVERAGE_JOBS is a newly documented switch (DEVELOPMENT.md) but, unlike COVERAGE_HTML on the lines above, it is never validated. A value such as COVERAGE_JOBS=abc makes the `[[ ... -le 1 ]]` / `[[ ${#files[@]} -le COVERAGE_JOBS ]]` comparisons emit "integer expression expected" and `$((COVERAGE_JOBS - 1))` silently evaluate to -1, quietly degrading to a single worker while printing confusing stderr noise; COVERAGE_JOBS=0 silently forces the sequential path. Consider validating it as a positive integer with the same pattern used for COVERAGE_HTML here and for MILVUS_UNITY_BATCH_SIZE in scripts/build.sh.

scripts/build.sh
```diff
@@ -246,6 +281,9 @@ CMAKE_CMD="cmake \
 ${CMAKE_VERSION_ARG} \
 -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX} \
 -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS} \
+-DMILVUS_ENABLE_UNITY=${UNITY} \
```
yhmo, line 294: scripts/build.sh line:284
Low ---- DEVELOPMENT.md states that UNITY applies to `make coverage`, but the new workflow comment in .github/workflows/main.yaml says unity batches can distort gcov/lcov per-file line attribution and therefore keeps UNITY=OFF on the coverage job. There is no guard or doc warning for a local user who follows the docs and runs `UNITY=ON make coverage`, which would silently produce a distorted/misattributed coverage report (lcov attributing lines to the generated unity batch files instead of the real sources). Consider failing fast (or at least warning) when MILVUS_ENABLE_UNITY=ON and MILVUS_BUILD_COVERAGE=ON, and note the interaction in the coverage section of DEVELOPMENT.md.

src/CMakeLists.txt
```diff
@@ -32,6 +32,14 @@ endif()
 
 # add proto gens
 add_milvus_protos(milvus_sdk_objects)
+
+if (MILVUS_ENABLE_UNITY)
+    # Generated protobuf sources stay out of unity batches via SKIP_UNITY_BUILD_INCLUSION
+    # set next to each generated file in cmake/MilvusProtoGen.cmake.
+    set_target_properties(milvus_sdk_objects PROPERTIES
+        UNITY_BUILD ON
+        UNITY_BUILD_BATCH_SIZE "${MILVUS_UNITY_BATCH_SIZE}")
```
yhmo, line 41: src/CMakeLists.txt line:41
Medium ---- The batch-size option accepts any positive integer, but BloomFilter.cpp and RoaringBitmap.cpp still define the same anonymous-namespace helpers (WriteU16LE/WriteU32LE/WriteU64LE). A valid batch large enough to put those files together (the target currently has 220 eligible implementation sources, so 220 does so) makes them share one unnamed namespace and fails compilation with redefinition errors. Please rename the helpers per format (matching the header constants) or consolidate them in one internal utility so every accepted batch size remains buildable.

scripts/coverage.sh
yhmo: Fixed. `coverage.sh` now validates `COVERAGE_JOBS` as a positive integer with the same pattern as `COVERAGE_HTML` and `MILVUS_UNITY_BATCH_SIZE`, so values like `abc` or `0` fail immediately with a clear message instead of degrading to a single worker with arithmetic noise.

scripts/build.sh
yhmo: Fixed. `build.sh` now prints an explicit warning when `UNITY=ON` is combined with a coverage build, and the UNITY bullet in DEVELOPMENT.md documents the interaction and recommends keeping it off for coverage runs. We chose a warning rather than failing hard: the attribution-distortion claim is not empirically validated for this toolchain yet, and CI keeps the coverage job pinned to `UNITY: OFF` regardless.

src/CMakeLists.txt
yhmo: Fixed. The three helpers were byte-for-byte identical, so they now live in one shared internal header, `src/impl/utils/LittleEndian.h` (inline functions in `namespace milvus`), and both files dropped their local copies (net -27 lines). `ReadU32LE`/`ReadU64LE` stay in `BloomFilter.cpp` because only its XXH64 implementation uses them. Verified with `MILVUS_UNITY_BATCH_SIZE=300`, which forces all 220 eligible sources into a single unity batch and builds cleanly, plus the BloomFilter/RoaringBitmap unit tests (22/22 pass) — every accepted batch size now builds.

scripts/build.sh
```diff
@@ -29,12 +29,38 @@ MILVUS_SDK_VERSION=${MILVUS_SDK_VERSION:-}
...
 BUILD_FROM_CONAN="ON"
 CPPSTD=${CPPSTD:-14}
 if [[ ! "${CPPSTD}" =~ ^[0-9]+$ ]]; then
   echo "ERROR! CPPSTD must be a numeric C++ standard value, such as 14, 17, or 20"
   exit 1
 fi
+case "${UNITY}" in
+  ON|OFF) ;;
+  *)
+    echo "ERROR! UNITY must be ON or OFF"
```
yhmo, line 44: scripts/build.sh line:44
Low ---- UNITY is validated here but never disabled for lint runs, so `UNITY=ON make lint` still configures a unity tree and compile_commands.json then holds entries for the generated unity batch files instead of the individual sources; run_clang_tidy.py's per-source lookup falls back to slow/incorrect commands. The CI step already works around this with a step-level `UNITY: OFF` override in main.yaml, but local users following the new DEVELOPMENT.md switches can still hit it. Since build.sh already forces MILVUS_ENABLE_PCH=OFF when RUN_CPPLINT=ON for the same clang-tidy reason, please mirror that here (force UNITY=OFF for lint runs) so the guard lives in the script rather than being re-discovered in every workflow.

.github/workflows/main.yaml
```diff
@@ -131,8 +138,18 @@ jobs:
...
+      # Coverage objects carry full DWARF + gcov sections (50-70 MB each); 2G only
+      # held ~30 of 200+ entries, crippling follow-up-commit hit rates. The disk
+      # cleanup step frees ~20G, so 8G is safe here.
+      CCACHE_MAXSIZE: 8G
       BUILD_SHARED_LIBS: OFF
+      # Keep unity off here: unity batches can distort gcov/lcov per-file line
+      # attribution. line-tables is safe: gcov reads .gcno/.gcda notes, not DWARF
+      # (validated: identical hit/miss coverage output with full debug info).
+      UNITY: OFF
+      LINE: ON
```
yhmo, line 184: .github/workflows/main.yaml line:150
Low ---- This `LINE: ON` maps to `MILVUS_DEBUG_INFO=line-tables` (`-g1` on GCC), which deliberately skips variable-level DWARF, yet the comment added a few lines above says coverage objects "carry full DWARF + gcov sections (50-70 MB each)" and sizes `CCACHE_MAXSIZE` on that basis. The two statements are inconsistent within the same job, so a future maintainer re-sizing the ccache budget (or toggling `LINE`) can be misled. Please align the comment with the actual object layout (line-tables + gcov sections) or state explicitly that the 8G budget is deliberately conservative.

scripts/build.sh
yhmo: Fixed. `build.sh` now mirrors the existing PCH guard: when `RUN_CPPLINT=ON` and `UNITY=ON`, it prints a note and forces `UNITY=OFF` so clang-tidy keeps per-file compile commands. The step-level `UNITY: OFF` override in main.yaml is removed so the guard lives in one place.

.github/workflows/main.yaml
yhmo: Fixed. The comment now describes the actual object layout (gcov sections + line-tables debug info, or full DWARF when LINE is toggled off) and states explicitly that 8G is a deliberately conservative budget chosen to also cover full-DWARF runs.

.github/workflows/main.yaml
```diff
@@ -87,11 +90,25 @@ jobs:
       CCACHE_COMPRESSLEVEL: 5
       CCACHE_MAXSIZE: 2G
       BUILD_SHARED_LIBS: ${{ matrix.os.build_shared_libs }}
+      UNITY: ON
```
yhmo, line 93: .github/workflows/main.yaml line:93
Low ---- The PR description says "Linux/macOS build jobs run with UNITY=ON LINE=ON", but this Linux matrix job sets only UNITY: ON in its env -- LINE is not set, so build.sh passes MILVUS_DEBUG_INFO=full and this job builds with full DWARF instead of the advertised line-tables speedup. The macOS job (env around line 321) and the coverage job (line 211) both set LINE: ON. Is the omission intentional, or an oversight? If intended, the description should be updated; if not, adding LINE: ON here would apply the line-tables gain to the largest CI job.

.github/workflows/main.yaml
```diff
@@ -105,10 +122,12 @@ jobs:
...
         if: ${{ matrix.os.distro == 'Ubuntu' }}
         run: |
-          make lint
+          # Release matches the dependency stack the tutorials step consumes below,
+          # so one Conan cache serves the whole job.
+          make lint-release
       - name: Unit Testing
         run: |
-          make test
+          make test-release
```
yhmo, line 132: .github/workflows/main.yaml line:130
Low ---- This step switches the Linux job from `make test` (Debug) to `make test-release` (Release), and the Windows job now also configures -DCMAKE_BUILD_TYPE=Release, while the macOS job (line 344) still runs `make test` (Debug). The Release choice is explained by the shared-Conan-cache rationale, and the practical loss is small since the unit tests use gtest/EXPECT rather than assert() and src/impl has no NDEBUG-gated code -- but Debug is no longer exercised by unit tests on Linux or Windows. Please confirm this drop in Debug test coverage is intentional and acceptable, with macOS remaining the only Debug-tested platform.

.github/workflows/main.yaml
```diff
@@ -117,6 +136,52 @@ jobs:
...
+          export PATH="$HOME/.venv/bin:$PATH"
+          conan cache clean "*" -s -b -d || true
+      - name: Prune old ccache generations
+        if: always()
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PRUNE_PREFIXES: "linux-${{ matrix.os.key }}-ccache- linux-${{ matrix.os.key }}-conan-"
+          KEEP_KEYS: "linux-${{ matrix.os.key }}-ccache-${{ github.sha }} linux-${{ matrix.os.key }}-conan-${{ github.base_ref || github.ref_name }}-${{ hashFiles('conanfile.*', 'CMakeLists.txt', 'cmake/**', 'scripts/install_deps.sh') }}"
+        run: |
+          python3 <<'PY'
```
yhmo: .github/workflows/main.yaml line:155
Low ---- This ~50-line "Prune old ccache generations" Python block is duplicated verbatim in four jobs (Linux line 155, coverage line 272, macOS line 357, Windows line 476). Any future fix to the prune logic (e.g., handling the 100-entry per_page cap, rate-limit backoff, or the always()-after-a-failed-build case) must be replicated in all four copies and is easy to miss. Consider extracting it to a checked-in script (e.g., scripts/prune_caches.py) that each job invokes with the same env vars, so the logic has a single owner.

.github/workflows/main.yaml
```diff
@@ -117,6 +136,52 @@ jobs:
...
+          prefixes = os.environ["PRUNE_PREFIXES"].split()
+          keeps = set(os.environ["KEEP_KEYS"].split())
+          def api(path, method="GET"):
+              req = urllib.request.Request(
+                  "https://api.github.com" + path,
+                  headers={"Authorization": "Bearer " + token, "Accept": "application/vnd.github+json"},
+                  method=method)
+              with urllib.request.urlopen(req) as response:
+                  return json.load(response) if response.status != 204 else None
+          caches = api("/repos/%s/actions/caches?ref=%s&per_page=100" % (repo, ref))
```
yhmo: .github/workflows/main.yaml line:169
Medium ---- This cache-list GET is not wrapped in try/except, while the DELETE in the same loop is. The step comment promises pruning "must never fail the build", but a transient cache-API error on this GET (rate limit/429, 5xx) raises an uncaught urllib.error.HTTPError and fails the job in this if: always() step even after a fully green build. Wrap the GET (or the whole best-effort section) with the same try/except used for the DELETE so any cache-API failure only prints a warning, as the comment intends.

scripts/coverage.sh
```diff
@@ -64,21 +86,92 @@ FILE_INFO_UT="${DIR_LCOV_OUTPUT}/lcov_ut.info"
...
+            cat "${chunk_root}/part_${j}.log"
+            failed=1
+        fi
+    done
+    if [[ ${failed} -ne 0 ]]; then
+        rm -rf "${chunk_root}"
+        return 1
+    fi
+
+    ${LCOV_CMD} --gcov-tool "${GCOV_TOOL}" "${merge_args[@]}" -o "${output}"
```
yhmo, line 170: scripts/coverage.sh line:154
Low ---- lcov -a aborts (verified locally: exit 255, "no valid records found in tracefile") if ANY input part contains zero records. The parallel capture applies --exclude "*.pb.cc"/"*.grpc.pb.cc" per chunk, and the proto object files are captured then dropped; a chunk whose files are all excluded (possible for large COVERAGE_JOBS or unusual file layouts) produces an empty part, and the merge then fails the whole capture. A cheap guard that skips part files with no records before merging would make the parallel path robust.

.github/workflows/release.yaml
```diff
@@ -138,6 +138,10 @@ jobs:
     timeout-minutes: 100
     env:
       BUILD_SHARED_LIBS: ON
+      UNITY: ON
```
yhmo: .github/workflows/release.yaml line:141
Medium ---- The official release packages are now built with unity builds (Linux/macOS `UNITY: ON` here and at line 193, Windows `-DMILVUS_ENABLE_UNITY=ON` at line 272), but the PR description's safety claim -- "the installed library is byte-for-byte identical by default" -- covers only the default UNITY=OFF configuration. Unity merges translation units, so the compiler can see across the former TU boundaries and the published artifacts are produced by a different compilation mode than that baseline; the claim does not extend to them, and the release-workflow change is not mentioned in the PR description. Since release binaries are the most conservative surface, consider keeping UNITY=OFF for release packaging (the speedup mainly benefits cold CI builds) or explicitly documenting that released artifacts are unity-built and validating artifact equivalence against a non-unity build.

DEVELOPMENT.md
```diff
@@ -86,6 +87,33 @@ $ bash scripts/build.sh -v v3.0.2 -t Release
...
+  and distort the lcov report (`build.sh` prints a warning for this combination).
+- `LINE=ON` maps to CMake `MILVUS_DEBUG_INFO=line-tables`, which keeps stack traces and
+  `file:line` debug information but skips variable-level DWARF (`-gline-tables-only` on Clang,
+  `-g1` on GCC). It applies to Debug and RelWithDebInfo builds. Switch back to `LINE=OFF` when
+  you need to inspect variables in a debugger. It is also safe for coverage builds: gcov
+  coverage data comes from the `.gcno`/`.gcda` notes files, not DWARF (validated by identical
+  lcov output between full and line-tables debug info).
+
+Both default to `OFF`, so regular builds are unchanged. Measured cold-build effect for
+`UNITY=ON LINE=ON` versus defaults: about 2.3x faster with roughly half the disk usage. Use
```
yhmo: DEVELOPMENT.md line:112
Low ---- The doc says `UNITY=ON LINE=ON` gives "about 2.3x faster with roughly half the disk usage", but the PR body's measured table reports 2.6x (597s -> 227s) for the two flags combined and attributes the -47% disk figure to LINE=ON alone. Since both documents describe the same switches, a reader comparing them sees contradictory numbers; please align the doc with the PR body's figures (2.6x combined, and restate the disk figure since -47% was measured for LINE=ON only unless the combined number is re-measured).

scripts/coverage.sh
yhmo: Fixed. Before merging, the parallel capture now filters each chunk part and keeps only parts that contain records (verified by the presence of SF: entries); empty parts produced by all-excluded chunks (e.g. generated protobuf sources) are skipped, so lcov -a can no longer abort with exit 255 on a zero-record tracefile. The guard was validated locally: merging an empty part reproduces the original lcov ERROR/exit 255, while skipping it merges successfully.

.github/workflows/main.yaml
yhmo: Fixed. The omission was deliberate but under-documented: this job now builds **Release** (to share one Conan Release stack across lint, tests, and tutorials), and `MILVUS_DEBUG_INFO` only applies to Debug/RelWithDebInfo, so `LINE` is a no-op there. The workflow now carries a comment explaining why `LINE` is unset, and the PR description's CI-wiring section has been rewritten to match the actual per-job configuration.

.github/workflows/main.yaml
yhmo: The premise needs one correction: Debug-on-Linux test coverage is **not** dropped. The coverage job runs `make coverage`, which executes the full UT + IT + ST suite in Debug (with coverage instrumentation) on Linux for every PR — that path is unchanged and stays green. Combined with macOS still running Debug `make test`, both configurations remain continuously tested; the Release switch only moved which *job* exercises Debug on Linux. That said, the point about documenting the intent was fair, and the updated PR description now spells out the per-job build-type matrix.

.github/workflows/main.yaml
yhmo: Fixed. The prune logic now lives in `scripts/prune_caches.py`; all four jobs invoke it with the same `PRUNE_PREFIXES`/`KEEP_KEYS` env contract, so future changes have a single owner.

.github/workflows/main.yaml
yhmo: Fixed. The list call (and any other API failure) is now caught in the extracted `scripts/prune_caches.py`: a failed GET prints `WARN: could not list caches, skipping prune` and exits 0, matching the "never fail the build" contract the same way the DELETE wrapper did.

.github/workflows/release.yaml
yhmo: Fixed by reverting the conservative direction: release packaging (Linux/macOS env and the Windows cmake invocation) is back to the default `UNITY=OFF`, so official artifacts are produced by the default compilation mode and the PR description's safety claim applies to them unchanged. Release builds keep only the `CMAKE_BUILD_EXAMPLES=OFF` optimization, which does not affect artifact contents. The description now states this explicitly.

DEVELOPMENT.md
yhmo: Fixed. The doc now says "about 2.6x faster", matching the PR body table (597 s → 227 s). The disk claim ("roughly half") is consistent: the combined measurement was 3.0 GB → 1.6 GB (−47%).

.github/workflows/main.yaml
```diff
@@ -176,6 +229,18 @@ jobs:
...
+        # This job also builds the gRPC stack from source; strip the leftover build
+        # trees and sources before the post-job cache save (see the matrix job).
+        if: always()
+        run: conan cache clean "*" -s -b -d || true
+      - name: Prune old cache generations
+        if: always()
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PRUNE_PREFIXES: "ubuntu-2204-coverage-ccache- ubuntu-2204-conan-"
+          KEEP_KEYS: "ubuntu-2204-conan-${{ github.base_ref || github.ref_name }}-${{ hashFiles('conanfile.*', 'CMakeLists.txt', 'cmake/**', 'scripts/install_deps.sh') }}"
```
yhmo: .github/workflows/main.yaml line:242
Medium ---- KEEP_KEYS here keeps only the conan key, but PRUNE_PREFIXES also matches the ccache prefix, so every ubuntu-2204-coverage-ccache-* entry is deleted on each run -- including the current commit's key that this job just restored. The restore falls back through `restore-keys: ubuntu-2204-coverage-ccache-`, so with all previous generations deleted a later commit can never match one: coverage ccache always starts cold and the 8G budget justified in the comment above ("follow-up-commit hit rates") is effectively wasted. The other three jobs all keep the current sha's ccache key in KEEP_KEYS; adding `ubuntu-2204-coverage-ccache-${{ github.sha }}` here would match that behaviour and preserve the intended cross-commit hit rate.

scripts/prune_caches.py
```diff
@@ -0,0 +1,75 @@
...
+        "https://api.github.com" + path,
+        headers={"Authorization": "Bearer " + token, "Accept": "application/vnd.github+json"},
+        method=method,
+    )
+    with urllib.request.urlopen(request) as response:
+        return json.load(response) if response.status != 204 else None
+
+
+def main():
+    token = os.environ["GH_TOKEN"]
```
yhmo: scripts/prune_caches.py line:48
Low ---- The module docstring promises pruning is "deliberately infallible: any API error ... only prints a warning and never fails the build", but os.environ["GH_TOKEN"] here (and GITHUB_REPOSITORY / GITHUB_REF on the next lines) are read before the try/except. A missing or differently-named variable raises KeyError and fails the step; every call site runs the step with `if: always()`, so that would turn a green build red. Reading them via os.environ.get(...) with a warning + early return would keep the script consistent with its documented contract.

scripts/prune_caches.py
```diff
@@ -0,0 +1,75 @@
...
+    repo = os.environ["GITHUB_REPOSITORY"]
+    ref = os.environ["GITHUB_REF"]
+    prefixes = os.environ.get("PRUNE_PREFIXES", "").split()
+    keeps = set(os.environ.get("KEEP_KEYS", "").split())
+
+    try:
+        caches = call_api("/repos/%s/actions/caches?ref=%s&per_page=100" % (repo, ref), token)
```
yhmo: scripts/prune_caches.py line:55
Low ---- The list call caps at per_page=100 and never follows pages. Once a branch accumulates more than 100 cache entries (each commit adds its own per-sha ccache key), the older generations are not returned and therefore never pruned, which works against the script's stated goal of staying under the 10 GB quota. Iterating until total_count entries are consumed (or following the Link header) would close the gap.

.github/workflows/main.yaml
yhmo: Fixed. `KEEP_KEYS` now also includes `ubuntu-2204-coverage-ccache-${{ github.sha }}`, so the generation this run restored (and will re-save) is never deleted, matching the behavior of the other three jobs. One nit on the mechanism as described: the post-job save still recreates the current key after pruning, so follow-up commits could still hit the prefix in normal operation — but the keep-key form is strictly more robust (no cold-cache window if a save ever fails) and consistent, so it is the right shape.

scripts/prune_caches.py
yhmo: Fixed. `GH_TOKEN`, `GITHUB_REPOSITORY`, and `GITHUB_REF` are now read via `os.environ.get(...)`; if any is missing the script prints a warning and returns 0, keeping the documented never-fails-the-build contract (verified locally with an unset environment).

scripts/prune_caches.py
yhmo: Fixed. The list call now follows pagination (`page` parameter) until `total_count` entries are consumed, with a hard cap of 100 pages as a safety bound, so older generations beyond the first 100 entries are reachable and prunable.

src/impl/types/BloomFilter.cpp
```diff
@@ -40,11 +40,13 @@
 #include <stdexcept>
 #include <string>
 
+#include "../utils/LittleEndian.h"
```
yhmo, line 43: src/impl/types/BloomFilter.cpp line:43
Low ---- LittleEndian.h extracts only the Write*LE helpers, but ReadU32LE and ReadU64LE (defined just below this include in the anonymous namespace) carry the same unity-batch collision risk that motivated this header: if any sibling TU sharing a unity batch ever adds the same names, the batch fails to compile. Since the CI builds now default to UNITY=ON on the Linux/macOS jobs, consider moving the Read helpers into LittleEndian.h too so the whole little-endian surface stays centralized.

.github/workflows/main.yaml
```diff
@@ -272,7 +352,9 @@ jobs:
           rem guard every compile would try to spawn a non-existent ccache.exe
           rem and the build would die with "CreateProcess failed".
           where ccache >nul 2>&1 && (set LAUNCHER=-DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache) || (set LAUNCHER=)
-          cmake -S . -B build -DMILVUS_BUILD_TEST=YES -DBUILD_SHARED_LIBS=ON -G Ninja %LAUNCHER%
+          rem MSVC has no line-tables-only mode, so only unity build applies here.
+          rem Test the optimized build that matches what release packaging consumes.
```
yhmo: .github/workflows/main.yaml line:356
Low ---- This comment claims the Windows unity Release build "matches what release packaging consumes", but the release.yaml Windows job builds with the default UNITY=OFF (only -DMILVUS_BUILD_EXAMPLES=OFF is added there). CI therefore validates a unity configuration that differs from the shipped artifact, and a maintainer trusting this comment could believe the release packaging path is exercised here. Consider rewording to "tests the optimized build configuration" or explicitly noting the unity divergence.

scripts/prune_caches.py
```diff
@@ -0,0 +1,90 @@
...
+        if page > 100:  # hard cap; 10k entries is far beyond any realistic repository
+            break
+
+    removed = 0
+    failed = 0
+    for cache in entries:
+        key = cache["key"]
+        if any(key.startswith(prefix) for prefix in prefixes) and key not in keeps:
+            try:
+                call_api("/repos/%s/actions/caches/%d" % (repo, cache["id"]), token, "DELETE")
```
yhmo, line 92: scripts/prune_caches.py line:81
Low ---- The prune runs with `if: always()` and deletes every non-kept generation during the run, before the post-job cache save of the current run. With `concurrency.cancel-in-progress: true` in main.yaml, a job cancelled by a newer push can skip its save after the old generation was already deleted, losing the last per-sha ccache generation. The expensive conan keys survive (they are in KEEP_KEYS), so this only costs a cold ccache rebuild, but consider keeping the previous generation until the new key is actually saved.

src/impl/types/BloomFilter.cpp
yhmo: Fixed. `ReadU32LE`/`ReadU64LE` moved into `LittleEndian.h` alongside the Write helpers, so the whole little-endian surface is centralized and no sibling TU can re-introduce the collision. BloomFilter unit tests pass (9/9).

.github/workflows/main.yaml
yhmo: Fixed. The comment now reads "Test the optimized build configuration" and explicitly notes that release packaging stays with the default `UNITY=OFF`, so the two configurations differ there.

scripts/prune_caches.py
yhmo: Fixed. `prune_caches.py` now retains the newest non-kept entry per prefix as a fallback, so a run cancelled between the prune and its post-job save cannot take a chain from warm to empty. Steady state stays bounded at two generations per prefix.

Makefile
```diff
@@ -45,6 +45,9 @@ all: all-debug
 lint:
 	@(env bash ${PWD}/scripts/build.sh -l)
 
+lint-release:
```
yhmo, line 48: Makefile line:48
Low ---- The new `lint-release` target is not declared `.PHONY` (the pre-existing `lint` target has the same gap), while every other command target in this file (`test`, `tutorials`, `run-tutorial`, `clean`, `doc`, `package`, `run`) is declared `.PHONY` in the list at the bottom. Since `make lint-release` is now a CI-critical step in `.github/workflows/main.yaml`, a stray file or directory named `lint-release` at the repo root would make make treat the target as up-to-date and the CI lint step would silently become a no-op. Consider adding `lint` and `lint-release` to the `.PHONY` list.

yhmo: Fixed. Both `lint` and `lint-release` are now declared `.PHONY` alongside the other command targets, so a stray file or directory with either name can no longer make the CI lint step silently skip.

PR #591, opened 2026-08-31
Align RBAC/GetServerVersion with pymilvus

yhmo:


src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3659,11 +3696,23 @@ MilvusClientV2Impl::UpdatePassword(const UpdatePasswordRequest& request) {
...
-    return connection_.Invoke<proto::milvus::UpdateCredentialRequest, proto::common::Status>(
+    auto status = connection_.Invoke<proto::milvus::UpdateCredentialRequest, proto::common::Status>(
         pre, &MilvusConnection::UpdateCredential, nullptr);
+    if (!status.IsOk()) {
+        return status;
+    }
+
+    if (request.ResetConnection()) {
+        // reconnect with the new credentials so subsequent RPCs keep the updated identity
+        return connection_.ResetUserCredentials(request.UserName(), request.NewPassword());
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:3713
Medium ---- Partial-mutation on ResetConnection: if the UpdateCredential RPC succeeds but the reconnect issued by ResetUserCredentials fails, the server password has already been changed yet this method returns the reconnect error. The existing connection still holds the old credentials, so subsequent RPCs on this client fail authentication, and a caller retrying with the old password also fails. Consider distinguishing "password updated, reconnect failed" from a total failure in the returned Status (or documenting that the caller must reconnect manually with the new password) so the partially-mutated state is not silent.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -2656,8 +2682,19 @@ MilvusClientV2Impl::compact(const std::string& database_name, const CompactReque
...
-            rpc_request.set_target_size(request.TargetSize());
+        rpc_request.set_l0compaction(request.IsL0());
+        if (request.TargetSize() != 0) {
+            if (request.TargetSize() < 0) {
+                return Status{StatusCode::INVALID_ARGUMENT, "target_size must be a positive integer"};
+            }
+            int64_t target_size_mb = 0;
+            std::string normalized;
+            const std::string target_size_text = std::to_string(request.TargetSize()) + request.TargetSizeUnit();
+            auto status = ParseTargetSizeMB(target_size_text, target_size_mb, normalized);
```
yhmo, line 2710: src/impl/MilvusClientV2Impl.cpp line:2693
Low ---- Compact now routes TargetSize through ParseTargetSizeMB, whose error text is worded for the Optimize API ("Invalid optimize target size ...", "Optimize target size must be at least 1MB"). A Compact caller that passes an invalid unit (e.g. "xq") or a sub-1MB value will see "optimize" wording, which is misleading in this context. Consider generalizing the helper message or wrapping it here so the error matches the Compact operation.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -104,6 +104,32 @@ MilvusClientV2Impl::GetServerVersion(std::string& version) {
...
+        response.SetVersion(info.build_tags());
+        response.SetBuildTime(info.build_time());
+        response.SetGitCommit(info.git_commit());
+        response.SetGoVersion(info.go_version());
+        response.SetDeployMode(info.deploy_mode());
+        return Status::OK();
+    };
+
+    return connection_.Invoke<proto::milvus::ConnectRequest, proto::milvus::ConnectResponse>(
+        nullptr, &MilvusConnection::ConnectRpc, post);
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:130
Low ---- The Detail() path issues a Connect RPC with a default-constructed ConnectRequest (no client_info), whereas pymilvus populates client_info via register_request and this SDK's own MilvusConnection::Connect handshake fills sdk_type/sdk_version/user. Server-side Connect logging/telemetry will not see the SDK identity for this call. Consider populating client_info here for parity with the normal connect flow.

src/impl/MilvusClientV2Impl.cpp
yhmo, line 2710: src/impl/MilvusClientV2Impl.cpp line:2693
Low ---- This compact path reuses ParseTargetSizeMB, whose error strings still say "optimize" (e.g. "Invalid optimize target size unit: 1xq" and "Optimize target size must be at least 1MB"). A Compact caller who passes an invalid unit or a sub-1MB target size will see an error that references "optimize", which is misleading. Consider making the helper's messages generic (e.g. "target size") or wrapping the status text here so Compact users get compact-specific wording.

src/impl/MilvusClientV2Impl.cpp
yhmo: src/impl/MilvusClientV2Impl.cpp line:3713
Low ---- Partial-state concern: UpdateCredential has already succeeded server-side before ResetUserCredentials() runs. If the reconnect fails (network error, or the new credentials rejected during the Connect handshake), UpdatePassword returns the reconnect error even though the password was actually changed, so a caller could retry with the old password or conclude the update failed. Consider documenting this semantics, or returning the RPC status as primary when the reset fails. Note also the IT test UpdatePasswordResetConnection only asserts the UpdateCredential request and OK status; it does not verify the follow-up Connect actually used the new credentials, so the reset path is only partially exercised.

src/impl/MilvusClientV2Impl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientV2Impl.cpp

src/impl/MilvusClientV2Impl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientV2Impl.cpp

test/st/MilvusContainerEnv.h
```diff
@@ -96,15 +96,9 @@ class MilvusContainerEnv : public ::testing::Environment {
...
-            "docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' " + container_id_ + " 2>&1";
-        std::string container_ip = execCommand(inspect_cmd);
-        container_ip.erase(container_ip.find_last_not_of(" \n\r\t") + 1);
-        container_ip.erase(0, container_ip.find_first_not_of(" \n\r\t"));
-        std::cout << "Milvus container IP: " << container_ip << std::endl;
-
-        setenv("MILVUS_HOST", container_ip.c_str(), 1);
+        // The tests connect to the host-published gRPC port (29730), which is only
+        // reachable through the host loopback, not the container's internal IP.
+        setenv("MILVUS_HOST", "127.0.0.1", 1);
```
yhmo, line 101: test/st/MilvusContainerEnv.h line:101
Medium ---- This change is unrelated to the stated RBAC/GetServerVersion alignment and bundles ST-harness churn into the feature PR: Milvus gRPC port 19530->29730, health 9091->19291, MinIO 9000->19200 (milvus_container.py, MilvusServerTest.h and several Test*.cpp), and MILVUS_HOST is now hardcoded to 127.0.0.1. The previous code deliberately resolved the container IP because "localhost may not work in CI"; hardcoding 127.0.0.1 only works when the test process and the Docker daemon share the same loopback (typical local Docker) and can regress CI setups that run the tests in a separate container / docker-in-docker, where the host-published port is not reachable via 127.0.0.1. If these renumbering and host-resolution changes are required to unblock CI (e.g. avoiding port conflicts on a shared runner), please split them into a separate test-infra PR or document why loopback is safe, so the feature change stays reviewable and the harness change is explicit.

yhmo: Thanks for the review. The ST-harness renumbering is intentional: the host ports were moved (gRPC 29730, health 19291, MinIO 19200) to avoid collisions on the shared CI runner. The ST tests run inside the `coverage-ubuntu` job directly on the GitHub runner, which shares the loopback with the Docker daemon, so `127.0.0.1:29730` reaches the host-published port. The previous container-IP resolution broke after this renumbering because the Milvus container only listens on 19530 internally, and no docker-in-docker / nested-container topology is used in this pipeline, so hardcoding the loopback is safe here. No change needed.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3659,11 +3713,29 @@ MilvusClientV2Impl::UpdatePassword(const UpdatePasswordRequest& request) {
...
 
-    return connection_.Invoke<proto::milvus::UpdateCredentialRequest, proto::common::Status>(
+    auto status = connection_.Invoke<proto::milvus::UpdateCredentialRequest, proto::common::Status>(
         pre, &MilvusConnection::UpdateCredential, nullptr);
+    if (!status.IsOk()) {
+        return status;
+    }
+
+    if (request.ResetConnection()) {
+        auto reset_status = connection_.ResetUserCredentials(request.UserName(), request.NewPassword());
```
yhmo, line 3729: src/impl/MilvusClientV2Impl.cpp line:3729
Low ---- The ResetConnection failure branch here (password updated, reconnect with new credentials failed) is documented and carefully worded, but no test exercises it. TestPrivilege.cpp::UpdatePasswordResetConnection only covers the success path (the second Connect RPC succeeds). Consider an IT case where the reconnect Connect fails, asserting the partial-state Status message so a future refactor cannot silently change the documented contract.

src/include/milvus/request/rbac/UpdatePasswordRequest.h
```diff
@@ -86,10 +86,58 @@ class MILVUS_SDK_API UpdatePasswordRequest {
...
+
+    /**
+     * @brief Whether to reset the current connection with the new credentials after the password is updated.
+     */
+    bool
+    ResetConnection() const;
+
+    /**
+     * @brief Set whether to reset the current connection with the new credentials after the password is updated.
+     * When true, the client reconnects using the new password so subsequent RPCs keep working.
```
yhmo, line 115: src/include/milvus/request/rbac/UpdatePasswordRequest.h line:115
Low ---- ResetConnection() reconnects with the *target* user's credentials (ResetUserCredentials(UserName(), NewPassword())). When the caller is connected as a different user (e.g. an admin resets another user's password) or connected via token, this silently switches the connection identity to the updated user and clears any existing token. Consider documenting this identity-switch behavior on the option so callers do not rely on "subsequent RPCs keep working" in the admin/other-user case.

yhmo: as designed

PR #595, opened 2026-09-02
Expose DML cost and validate search/hybrid_search params

yhmo:
- DmlResults gains a Cost() accessor; insert/upsert/delete now fill it from the server Status extra_info 'report_value', matching SearchResponse and PyMilvus.
- SearchRequestBase/SearchRequest/HybridSearchArguments/HybridSearchRequest Validate() now reject limit <= 0 and round_decimal outside (-2, 7), matching PyMilvus check_pass_param.

### DIFF
--- a/src/impl/MilvusClientV2Impl.cpp
+++ b/src/impl/MilvusClientV2Impl.cpp
@@ -1855,6 +1855,7 @@ MilvusClientV2Impl::insert(const InsertRequest& request, InsertResponse& respons
         results.SetIdArray(std::move(id_array));
         results.SetTimestamp(rpc_response.timestamp());
         results.SetInsertCount(static_cast<uint64_t>(rpc_response.insert_cnt()));
+        FillDmlCost(rpc_response.status(), results);
         response.SetResults(std::move(results));
 
         // special for dml api: if the api failed, remove the schema cache of this collection
@@ -1980,6 +1981,7 @@ MilvusClientV2Impl::upsert(const UpsertRequest& request, UpsertResponse& respons
         results.SetIdArray(std::move(id_array));
         results.SetTimestamp(rpc_response.timestamp());
         results.SetUpsertCount(static_cast<uint64_t>(rpc_response.upsert_cnt()));
+        FillDmlCost(rpc_response.status(), results);
         response.SetResults(std::move(results));
 
         // special for dml api: if the api failed, remove the schema cache of this collection
@@ -2067,6 +2069,7 @@ MilvusClientV2Impl::Delete(const DeleteRequest& request, DeleteResponse& respons
         results.SetIdArray(std::move(id_array));
         results.SetTimestamp(rpc_response.timestamp());
         results.SetDeleteCount(static_cast<uint64_t>(rpc_response.delete_cnt()));
+        FillDmlCost(rpc_response.status(), results);
         response.SetResults(std::move(results));
 
         if (!IsRealFailure(rpc_response.status())) {
@@ -2244,6 +2247,8 @@ MilvusClientV2Impl::hybridSearch(const HybridSearchRequest& request, HybridSearc
                                  const std::string& cluster_id) {
     const auto endpoint = connection_.CurrentEndpoint();
     const auto database_name = connection_.CurrentDbName(request.DatabaseName());
+    auto validate = [&request]() { return request.Validate(); };
+
     auto pre = [&endpoint, &database_name, &request, &cluster_id](proto::milvus::HybridSearchRequest& rpc_request) {
         return ConvertHybridSearchRequest<HybridSearchRequest>(request, database_name, rpc_request, cluster_id,
                                                                endpoint);
@@ -2272,7 +2277,7 @@ MilvusClientV2Impl::hybridSearch(const HybridSearchRequest& request, HybridSearc
 
     return InvokeWithTelemetry(connection_, "HybridSearch", request.CollectionName(), [&]() {
         return connection_.Invoke<proto::milvus::HybridSearchRequest, proto::milvus::SearchResults>(
-            pre, &MilvusConnection::HybridSearch, post);
+            validate, pre, &MilvusConnection::HybridSearch, post);
     });
 }
 
--- a/src/impl/request/dql/HybridSearchRequest.cpp
+++ b/src/impl/request/dql/HybridSearchRequest.cpp
@@ -185,4 +185,35 @@ HybridSearchRequest::WithStrictGroupSize(bool strict_group_size) {
     return *this;
 }
 
+Status
+HybridSearchRequest::Validate() const {
+    auto status = ValidateLimit(limit_);
+    if (!status.IsOk()) {
+        return status;
+    }
+
+    status = ValidateRoundDecimal(extra_params_);
+    if (!status.IsOk()) {
+        return status;
+    }
+
+    for (const auto& sub_request : sub_requests_) {
+        if (sub_request == nullptr) {
+            return {StatusCode::INVALID_ARGUMENT, "Sub request can not be null!"};
+        }
+        auto status = sub_request->Validate();
+        if (!status.IsOk()) {
+            return status;
+        }
+    }
+    if (function_ == nullptr) {
+        return {StatusCode::INVALID_ARGUMENT, "Rerank function is undefined!"};
+    }
+    if (function_->GetFunctionType() != FunctionType::RERANK) {
+        return {StatusCode::INVALID_ARGUMENT, "Hybrid search only accepts RERANK function!"};
+    }
+
+    return Status::OK();
+}
+
 }  // namespace milvus
--- a/src/impl/request/dql/SearchRequest.cpp
+++ b/src/impl/request/dql/SearchRequest.cpp
@@ -308,6 +308,16 @@ SearchRequest::AddOrderByField(OrderByField order_by_field) {
 
 Status
 SearchRequest::Validate() const {
+    auto status = ValidateLimit(Limit());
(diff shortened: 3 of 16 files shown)


src/include/milvus/types/DmlResults.h
```diff
@@ -95,12 +95,25 @@ class MILVUS_SDK_API DmlResults {
     void
     SetUpsertCount(uint64_t count);
 
+    /**
+     * @brief The operation cost in VCU reported by the server, 0 when unavailable.
```
yhmo: src/include/milvus/types/DmlResults.h line:99
Medium ---- DmlResults::Cost() uses 0 as the "unavailable" sentinel, but the sibling SearchResponse::Cost() in this same codebase reports -1 when the server does not include "report_value" (DqlUtils::FillSearchResponseExtraInfo initializes SetCost(-1)). A caller who uses the common idiom resp.Cost() < 0 to detect "not reported" will silently get different behavior for insert/upsert/delete results, and 0 is also a conceivable real cost value, which makes the sentinel ambiguous. Consider aligning DmlResults to the same -1 sentinel (or explicitly calling out the divergence) so both result types behave consistently.

test/ut/types/TestDmlResults.cpp
```diff
@@ -50,3 +50,10 @@ TEST_F(DmlResultsTest, UpsertCount) {
     dml_results.SetUpsertCount(75);
     EXPECT_EQ(dml_results.UpsertCount(), 75);
 }
+
+TEST_F(DmlResultsTest, Cost) {
```
yhmo, line 54: test/ut/types/TestDmlResults.cpp line:54
Medium ---- This test only exercises the Cost()/SetCost() accessors. The new parsing logic introduced by this PR (DmlUtils::FillDmlCost, which reads status.extra_info["report_value"] and is now called from every insert/upsert/delete path in both V1 and V2) has no direct test: no case for a present report_value, a missing key, or a non-numeric value (the std::stoll exception path). Given the cost value is the headline feature of this PR and the parser swallows parse failures silently (leaving cost 0), consider a focused unit test that constructs a proto::common::Status with/without extra_info and asserts DmlResults::Cost(). The new Validate() rejection tests are also one-sided (only limit=0 and round_decimal=7 are rejected; -2 and valid boundary values like -1 and 6 are not covered).

src/impl/request/dql/HybridSearchRequest.cpp
```diff
@@ -185,4 +185,34 @@ HybridSearchRequest::WithStrictGroupSize(bool strict_group_size) {
...
+    for (const auto& sub_request : sub_requests_) {
+        if (sub_request == nullptr) {
+            return {StatusCode::INVALID_ARGUMENT, "Sub request can not be null!"};
+        }
+        auto status = sub_request->Validate();
+        if (!status.IsOk()) {
+            return status;
+        }
+    }
+    if (function_ == nullptr) {
```
yhmo, line 209: src/impl/request/dql/HybridSearchRequest.cpp line:208
Medium ---- This new Validate() is now enforced on the V2 HybridSearch RPC path (the newly added `validate` lambda in MilvusClientV2Impl::hybridSearch), so the null-function branch here, the null-sub-request branch above it, and the `GetFunctionType() != RERANK` branch below are live production code. The unit tests added in this PR only exercise the `limit<=0` and `round_decimal=7` branches, so a request with a null/incorrect rerank function or a null sub-request could regress without any test catching it. Consider adding focused tests for these branches, plus one positive test that a fully valid HybridSearchRequest still passes.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -2244,6 +2247,8 @@ MilvusClientV2Impl::hybridSearch(const HybridSearchRequest& request, HybridSearc
                                  const std::string& cluster_id) {
     const auto endpoint = connection_.CurrentEndpoint();
     const auto database_name = connection_.CurrentDbName(request.DatabaseName());
+    auto validate = [&request]() { return request.Validate(); };
```
yhmo, line 2250: src/impl/MilvusClientV2Impl.cpp line:2250
Low ---- Before this PR, V2 HybridSearch performed no client-side validation (the Invoke call had no validate step), so invalid requests reached the server. Adding `validate` here changes existing V2 callers' behavior: limit<=0, round_decimal outside (-2, 7), a missing/non-RERANK rerank function, or a null sub-request now fail client-side with INVALID_ARGUMENT instead of a server error. This matches PyMilvus and is intended per the PR; it also prevents a latent null-pointer dereference in ConvertHybridSearchRequest (request.Rerank()->Params()) when no rerank function is set. Worth confirming the intended compatibility impact for existing callers.

src/impl/types/SearchRequestBase.cpp
```diff
@@ -259,6 +259,15 @@ SearchRequestBase::SetTimezone(const std::string& timezone) {
 
 Status
 SearchRequestBase::Validate() const {
+    if (limit_ <= 0) {
```
yhmo: src/impl/types/SearchRequestBase.cpp line:262
Low ---- The new limit>0 and round_decimal range checks are now duplicated in four Validate() implementations (SearchRequestBase, SearchRequest, HybridSearchArguments, HybridSearchRequest) with identical error strings. On the vector-search path SearchRequest::Validate() runs these checks and then delegates to SearchRequestBase::Validate(), which validates the same two conditions again. If the server's accepted range ever changes, all four copies must be updated consistently. Consider extracting a shared helper (e.g., ValidateLimit/ValidateRoundDecimal) so the policy has a single source of truth.

src/impl/request/dql/HybridSearchRequest.cpp
```diff
@@ -185,4 +185,34 @@ HybridSearchRequest::WithStrictGroupSize(bool strict_group_size) {
...
+Status
+HybridSearchRequest::Validate() const {
+    if (limit_ <= 0) {
+        return {StatusCode::INVALID_ARGUMENT, "limit must be greater than 0"};
+    }
+
+    const auto round_decimal = GetRoundDecimal();
+    if (round_decimal <= -2 || round_decimal >= 7) {
+        return {StatusCode::INVALID_ARGUMENT, "round_decimal must be in range (-2, 7)"};
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/request/dql/SearchRequest.cpp
```diff
@@ -308,6 +308,15 @@ SearchRequest::AddOrderByField(OrderByField order_by_field) {
...
 Status
 SearchRequest::Validate() const {
+    if (Limit() <= 0) {
+        return {StatusCode::INVALID_ARGUMENT, "limit must be greater than 0"};
+    }
+
+    const auto round_decimal = RoundDecimal();
+    if (round_decimal <= -2 || round_decimal >= 7) {
+        return {StatusCode::INVALID_ARGUMENT, "round_decimal must be in range (-2, 7)"};
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/types/HybridSearchArguments.cpp
```diff
@@ -204,6 +204,15 @@ HybridSearchArguments::ExtraParams() const {
...
 Status
 HybridSearchArguments::Validate() const {
+    if (limit_ <= 0) {
+        return {StatusCode::INVALID_ARGUMENT, "limit must be greater than 0"};
+    }
+
+    const auto round_decimal = RoundDecimal();
+    if (round_decimal <= -2 || round_decimal >= 7) {
+        return {StatusCode::INVALID_ARGUMENT, "round_decimal must be in range (-2, 7)"};
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/types/SearchRequestBase.cpp
```diff
@@ -259,6 +259,15 @@ SearchRequestBase::SetTimezone(const std::string& timezone) {
...
+        return Status{StatusCode::INVALID_ARGUMENT, "limit must be greater than 0"};
+    }
+
+    const auto round_decimal = GetExtraInt64(extra_params_, "round_decimal", -1);
+    if (round_decimal <= -2 || round_decimal >= 7) {
+        return Status{StatusCode::INVALID_ARGUMENT, "round_decimal must be in range (-2, 7)"};
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/include/milvus/types/DmlResults.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

test/ut/types/TestDmlResults.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/request/dql/HybridSearchRequest.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientV2Impl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/types/SearchRequestBase.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #596, opened 2026-09-03
Match pymilvus search iterator filter and DML/get short-circuits

yhmo:
- Add external_filter_func to iterator arguments; SearchIteratorV2 applies it per fetched page via SingleResult::FilterRows, matching pymilvus page filter.
- Short-circuit insert/upsert with empty data and get with empty ids (no RPC), returning empty results like pymilvus.

### DIFF
--- a/scripts/build_tutorials.sh
+++ b/scripts/build_tutorials.sh
@@ -19,17 +19,26 @@ if ! conan remote list | grep -q "^${REMOTE_NAME}:"; then
     conan remote add "${REMOTE_NAME}" "${REMOTE_URL}" --allowed-packages "milvus-sdk-cpp/*"
 fi
 
-echo "Creating milvus-sdk-cpp/${SDK_VERSION}@${SDK_USER}/${SDK_CHANNEL} from the current checkout"
-conan create "${ROOT_DIR}" \
+echo "Exporting milvus-sdk-cpp/${SDK_VERSION}@${SDK_USER}/${SDK_CHANNEL} from the current checkout"
+conan export "${ROOT_DIR}" \
     --version="${SDK_VERSION}" \
     --user="${SDK_USER}" \
-    --channel="${SDK_CHANNEL}" \
+    --channel="${SDK_CHANNEL}"
+
+# Build the SDK package only when the current checkout is not already cached.
+# conan create always forces a rebuild; export + install --build=missing reuses
+# the cached package whenever the recipe revision is unchanged.
+SDK_REF="milvus-sdk-cpp/${SDK_VERSION}@${SDK_USER}/${SDK_CHANNEL}"
+SDK_BUILD_DIR="$(mktemp -d)"
+trap 'rm -rf "${SDK_BUILD_DIR}"' EXIT
+conan install --requires="${SDK_REF}" \
+    --build=missing \
     -s build_type=Release \
     -s compiler.cppstd=14 \
     -s:b build_type=Release \
     -s:b compiler.cppstd=17 \
     -c tools.build:jobs="${JOBS}" \
-    --build=missing
+    --output-folder="${SDK_BUILD_DIR}"
 
 export MILVUS_SDK_VERSION="${SDK_VERSION}"
 export MILVUS_SDK_USER="${SDK_USER}"
--- a/src/impl/MilvusClientV2Impl.cpp
+++ b/src/impl/MilvusClientV2Impl.cpp
@@ -1772,6 +1772,24 @@ Status
 MilvusClientV2Impl::insert(const InsertRequest& request, InsertResponse& response, bool allow_retry) {
     const auto endpoint = connection_.CurrentEndpoint();
     const auto database_name = connection_.CurrentDbName(request.DatabaseName());
+
+    // Short-circuit empty data like pymilvus: no rows to insert, do not issue the RPC.
+    // Check the row count rather than container emptiness so a request whose
+    // columns hold zero rows (FieldData::Count() == 0) is still short-circuited.
+    const auto& columns = request.ColumnsData();
+    uint64_t column_count = 0;
+    if (!columns.empty() && columns.front() != nullptr) {
+        column_count = columns.front()->Count();
+    }
+    if (column_count == 0 && request.RowsData().empty()) {
+        if (request.CollectionName().empty()) {
+            return {StatusCode::INVALID_ARGUMENT, "Collection name cannot be empty!"};
+        }
+        DmlResults results;
+        response.SetResults(std::move(results));
+        return Status::OK();
+    }
+
     CollectionDescPtr collection_desc;
     std::vector<proto::schema::FieldData> rpc_fields;
     auto validate = [this, &endpoint, &database_name, &request, &collection_desc, &rpc_fields]() {
@@ -1892,6 +1910,25 @@ Status
 MilvusClientV2Impl::upsert(const UpsertRequest& request, UpsertResponse& response, bool allow_retry) {
     const auto endpoint = connection_.CurrentEndpoint();
     const auto database_name = connection_.CurrentDbName(request.DatabaseName());
+
+    // Short-circuit empty data like pymilvus: no rows to upsert, do not issue the RPC.
+    // Keep partial-update field operations flowing to the server for validation even
+    // when the request carries no rows (field_ops is a C++-specific input). The row
+    // count, not container emptiness, drives the short-circuit (see insert()).
+    const auto& upsert_columns = request.ColumnsData();
+    uint64_t upsert_column_count = 0;
+    if (!upsert_columns.empty() && upsert_columns.front() != nullptr) {
+        upsert_column_count = upsert_columns.front()->Count();
+    }
+    if (upsert_column_count == 0 && request.RowsData().empty() && request.FieldOps().empty()) {
+        if (request.CollectionName().empty()) {
+            return {StatusCode::INVALID_ARGUMENT, "Collection name cannot be empty!"};
+        }
+        DmlResults results;
+        response.SetResults(std::move(results));
+        return Status::OK();
+    }
+
     std::vector<proto::schema::FieldData> rpc_fields;
     CollectionDescPtr collection_desc;
     auto validate = [this, &endpoint, &database_name, &request, &collection_desc, &rpc_fields]() {
@@ -2354,6 +2391,16 @@ MilvusClientV2Impl::get(const GetRequest& request, GetResponse& response, const 
 Status
 MilvusClientV2Impl::getWithoutTelemetry(const GetRequest& request, GetResponse& response,
                                         const std::string& cluster_id) {
+    // Short-circuit empty ids like pymilvus: no rows to fetch, do not issue the RPC.
+    if (request.IDs().GetRowCount() == 0) {
+        if (request.CollectionName().empty()) {
+            return {StatusCode::INVALID_ARGUMENT, "Collection name cannot be empty!"};
+        }
+        response.SetResults(QueryResults{});
+        response.SetSessionTs(0);
+        return Status::OK();
+    }
+
     const auto endpoint = connection_.CurrentEndpoint();
     const auto database_name = connection_.CurrentDbName(request.DatabaseName());
     CollectionDescPtr collection_desc;
--- a/src/impl/types/SearchIteratorImpl.cpp
+++ b/src/impl/types/SearchIteratorImpl.cpp
@@ -58,24 +58,63 @@ SearchIteratorImpl<T>::Next(SingleResult& results) {
         output_count = std::min(output_count, left_count);
     }
 
-    if (SearchIteratorImpl::CachedCount(cache_) < output_count) {
-        // if cache is not sufficient, try to fill the result by probing with constant width
-        // until finish filling or exceeding max trial time: 10
-        auto status = trySearchFill(output_count);
+    // Apply the client-side page filter (pymilvus external_filter_func) if set.
+    // A filtered-out page does not grow the cache, so keep filling and fetching
+    // until a non-empty page is produced or the server has no more results.
+    const bool has_filter = static_cast<bool>(args_.ExternalFilterFunc());
+    while (true) {
+        if (SearchIteratorImpl::CachedCount(cache_) < output_count) {
+            // if cache is not sufficient, try to fill the result by probing with constant width
+            // until finish filling or exceeding max trial time: 10
+            auto status = trySearchFill(output_count);
+            if (!status.IsOk()) {
+                return status;
+            }
+        }
+
+        // return batch from the cache if cache is big enough
+        results.Clear();
+        auto status = SearchIteratorImpl::FetchPageFromCache(cache_, args_.OutputFields(), output_count, results);
         if (!status.IsOk()) {
             return status;
         }
+
+        if (results.GetRowCount() == 0) {
+            // no more rows from the server
+            break;
+        }
+
+        // keep width_ derived from the unfiltered page: the external filter below usually
+        // prunes rows below BatchSize, so the old post-loop width update would never fire
+        // and width_ (which drives radius expansion) would stay frozen at the init value.
+        if (results.GetRowCount() == args_.BatchSize()) {
+            updateWidth(results);
+        }
+
+        if (!has_filter) {
+            break;
+        }
+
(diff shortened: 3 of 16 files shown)


src/impl/types/SearchIteratorV2Impl.cpp
```diff
@@ -67,6 +67,19 @@ SearchIteratorV2Impl<T>::Next(SingleResult& results) {
             break;
         }
 
+        // Apply the client-side page filter (pymilvus external_filter_func) if set.
+        // A filtered-out page does not grow the cache, so keep pulling until the
+        // target length is reached or the server has no more results.
+        if (args_.ExternalFilterFunc()) {
```
yhmo, line 73: src/impl/types/SearchIteratorV2Impl.cpp line:73
Medium ---- ExternalFilterFunc() is only consulted in the SearchIteratorV2 paging loop. When the server does not support iterator v2, MilvusClientV2Impl::searchIterator (and the legacy MilvusClientImpl::SearchIterator) silently fall back to SearchIteratorImpl, whose Next() never applies the filter, so a caller's external_filter_func is silently dropped. pymilvus instead raises NOT_SUPPORTED in that case. Consider honoring the filter on the V1 fallback path too, or documenting/warning on the fallback so it is not silently ignored.

src/impl/types/SearchIteratorV2Impl.cpp
```diff
@@ -67,6 +67,19 @@ SearchIteratorV2Impl<T>::Next(SingleResult& results) {
...
+            status = args_.ExternalFilterFunc()(*single_result);
+            if (!status.IsOk()) {
+                return status;
+            }
+            if (single_result->GetRowCount() == 0) {
```
yhmo, line 78: src/impl/types/SearchIteratorV2Impl.cpp line:78
Medium ---- No test exercises the new per-page filter behavior in the iterator loop: the added tests cover only SingleResult::FilterRows and the getter/setter, not that Next() applies the filter on each page, drops a fully-filtered page via this continue, or accumulates filtered rows across pages up to target_len. pymilvus has iterator-level tests for external_filter_func (tests/unit/test_search_iterator.py). Consider a mocked SearchIteratorV2 test covering a fully-filtered page and a partially-filtered page.

src/impl/types/SearchResults.cpp
```diff
@@ -167,6 +168,60 @@ SingleResult::WithHighlightResults(std::vector<HighlightResults>&& highlight_res
...
+            continue;
+        }
+
+        FieldDataPtr target;
+        for (uint64_t idx : keep_indices) {
+            if (idx >= field->Count()) {
+                return {StatusCode::INVALID_ARGUMENT, "row index out of bound"};
+            }
+            FieldDataPtr piece;
+            auto status = CopyFieldData(field, idx, idx + 1, piece);
```
yhmo: src/impl/types/SearchResults.cpp line:194
Low ---- FilterRows copies the page field-by-field and row-by-row: for each output field and each kept index it builds a single-row FieldData (CopyFieldData) and then merges it (AppendFieldData), so filtering a page costs O(fields x kept_rows) allocations plus per-row vector copies on every Next() page. pymilvus filters via list slicing. For large batch sizes or wide schemas this is noticeably more work than needed; consider copying directly into a pre-reserved vector per field instead of per-row piece + append.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1772,6 +1772,14 @@ Status
 MilvusClientV2Impl::insert(const InsertRequest& request, InsertResponse& response, bool allow_retry) {
     const auto endpoint = connection_.CurrentEndpoint();
     const auto database_name = connection_.CurrentDbName(request.DatabaseName());
+
+    // Short-circuit empty data like pymilvus: no rows to insert, do not issue the RPC.
+    if (request.ColumnsData().empty() && request.RowsData().empty()) {
+        DmlResults results;
+        response.SetResults(std::move(results));
+        return Status::OK();
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -2354,6 +2372,11 @@ MilvusClientV2Impl::get(const GetRequest& request, GetResponse& response, const
 Status
 MilvusClientV2Impl::getWithoutTelemetry(const GetRequest& request, GetResponse& response,
                                         const std::string& cluster_id) {
+    // Short-circuit empty ids like pymilvus: no rows to fetch, do not issue the RPC.
+    if (request.IDs().GetRowCount() == 0) {
+        return Status::OK();
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/types/SearchResults.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include "milvus/types/SearchResults.h"
 
+#include <algorithm>
 #include <stdexcept>
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/include/milvus/types/IteratorArguments.h
```diff
@@ -16,9 +16,12 @@
 
 #pragma once
 
+#include <functional>
+
 #include "FieldSchema.h"
 #include "QueryArguments.h"
 #include "SearchArguments.h"
+#include "SearchResults.h"
 #include "milvus/Export.h"
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/types/SearchIteratorV2Impl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/types/SearchIteratorV2Impl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/types/SearchResults.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

test/it/v1/TestSearchIterator.cpp
```diff
@@ -302,3 +302,149 @@ TEST_F(MilvusMockedTest, SearchIteratorV2PinsProbeTimestampForSessionConsistency
...
+            milvus::ConvertCollectionSchema(collection_schema, *proto_schema);
+            return ::grpc::Status{};
+        });
+
+    const uint64_t batch_size = 100;
+    uint64_t current_poz = 0;
+    bool probe_compability = true;
+    EXPECT_CALL(service_, Search(_, _, _))
+        .WillRepeatedly([&](::grpc::ServerContext*, const SearchRequest* request, SearchResults* response) {
+            response->mutable_results()->mutable_search_iterator_v2_results()->set_token("dummy");
```
yhmo: test/it/v1/TestSearchIterator.cpp line:343
Medium ---- This new test only exercises the SearchIteratorV2 path: the mocked Search always returns search_iterator_v2_results.token="dummy", so searchIterator() keeps the V2 iterator and the V1 fallback (SearchIteratorImpl::Next) is never reached. Since this PR also rewrote the V1 Next() with the client-side filter loop (per-page filter, fully-filtered-page continue, and the extra trySearchFill refill), that fallback path now has no direct test coverage — the V1 filter behavior is only guaranteed by code reading. Consider an IT variant with an empty token so SearchIteratorImpl::Next runs the filter loop against the mock.

src/impl/types/SearchIteratorImpl.cpp
```diff
@@ -58,19 +58,54 @@ SearchIteratorImpl<T>::Next(SingleResult& results) {
...
+        if (results.GetRowCount() == 0) {
+            // no more rows from the server
+            break;
+        }
+
+        if (!has_filter) {
+            break;
+        }
+
+        status = args_.ExternalFilterFunc()(results);
```
yhmo, line 98: src/impl/types/SearchIteratorImpl.cpp line:91
Low ---- With an external filter set, the filtered page usually has fewer rows than BatchSize, so the post-loop `if (results.GetRowCount() == args_.BatchSize()) updateWidth(results)` almost never fires and width_ stays frozen at the value from the init page / first unfiltered page. Since width_ drives the radius expansion in nextParams/trySearchFill, a stale width can make the V1 fallback probe the wrong range as iteration progresses (under- or over-extending radius and skipping rows). Consider updating width_ from the unfiltered page before FilterRows is applied, or recomputing it when the filter is active.

test/it/v1/TestSearchIterator.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/types/SearchIteratorImpl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/include/milvus/types/SearchResults.h
```diff
@@ -147,6 +147,15 @@ struct MILVUS_SDK_API SingleResult {
     SingleResult&
     WithHighlightResults(std::vector<HighlightResults>&& highlight_results);
 
+    /**
+     * @brief Keep only the rows whose indices are in `keep_indices`, dropping the rest.
+     * This is used by the client-side page filter of the search iterator to prune
+     * hits fetched from the server. The order of the kept rows follows `keep_indices`.
+     * @return Status::OK on success, otherwise an INVALID_ARGUMENT status.
```
yhmo: src/include/milvus/types/SearchResults.h line:154
Low ---- The FilterRows doc claims failures return "an INVALID_ARGUMENT status", but CopyFieldDataByIndices (DqlUtils.cpp:2208) also returns NOT_SUPPORTED for unsupported field/element types and INVALID_ARGUMENT for out-of-range indices. Consider wording this as "an error status" or enumerating the two codes so callers do not rely on a single error code.

scripts/build_tutorials.sh
```diff
@@ -19,17 +19,26 @@ if ! conan remote list | grep -q "^${REMOTE_NAME}:"; then
     conan remote add "${REMOTE_NAME}" "${REMOTE_URL}" --allowed-packages "milvus-sdk-cpp/*"
 fi
 
-echo "Creating milvus-sdk-cpp/${SDK_VERSION}@${SDK_USER}/${SDK_CHANNEL} from the current checkout"
-conan create "${ROOT_DIR}" \
+echo "Exporting milvus-sdk-cpp/${SDK_VERSION}@${SDK_USER}/${SDK_CHANNEL} from the current checkout"
+conan export "${ROOT_DIR}" \
```
yhmo, line 23: scripts/build_tutorials.sh line:23
Low ---- This conan create -> export + install refactor is unrelated to the PR's stated purpose (pymilvus iterator filter parity and DML/get short-circuits). Bundling an independent CI/build-script change into a feature PR makes the change harder to review and to revert independently. Consider splitting it into its own PR or at least calling it out in the PR description.

src/include/milvus/types/SearchResults.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

scripts/build_tutorials.sh
yhmo: This change is intentional and directly supports the CI work in this PR: build_tutorials.sh now exports + installs --build=missing so the cached SDK package is reused when src is unchanged, and the conan cache key now hashes src/** so changed sources get saved under a new key. Keeping it with the PR lets the tutorials step benefit from the same caching; splitting it out would leave that optimization orphaned. The conanfile/src hashing change is part of the same CI optimization and is covered by the PR description.

src/impl/types/SearchIteratorImpl.cpp
```diff
@@ -58,24 +58,63 @@ SearchIteratorImpl<T>::Next(SingleResult& results) {
...
+            updateWidth(results);
+        }
 
-    if (results.GetRowCount() == args_.BatchSize()) {
-        updateWidth(results);
+        if (!has_filter) {
+            break;
+        }
+
+        status = args_.ExternalFilterFunc()(results);
```
yhmo, line 98: src/impl/types/SearchIteratorImpl.cpp line:98
Low ---- The V1 fallback applies the client-side page filter to the page assembled by FetchPageFromCache, which can combine rows from several server pages plus leftover cache rows from a previous Next() call (e.g. when the cache front page is a split remainder and the next page fills output_count). On the V2 path (and in pymilvus) the callback always receives exactly one server page. For stateless row filters the results are identical, but for page-dependent filters (per-page dedup/sampling, keep-first-N-of-each-page) the same caller code behaves differently depending on which iterator implementation the SDK selects (older servers force the silent V1 fallback). Consider applying the filter per server page on the V1 path too, or document the grouping difference so callers do not rely on the "each page of hits fetched by the search iterator" doc claim.

src/impl/types/SearchResults.cpp
```diff
@@ -167,6 +168,46 @@ SingleResult::WithHighlightResults(std::vector<HighlightResults>&& highlight_res
...
+            return status;
+        }
+        filtered_fields.emplace_back(std::move(target));
+    }
+    output_fields_ = std::move(filtered_fields);
+
+    if (!highlight_results_.empty()) {
+        std::vector<HighlightResults> filtered_highlights;
+        filtered_highlights.reserve(keep_indices.size());
+        for (uint64_t idx : keep_indices) {
```
yhmo, line 200: src/impl/types/SearchResults.cpp line:200
Low ---- FilterRows is inconsistent between row data and highlight data for out-of-range indices: CopyFieldDataByIndices returns INVALID_ARGUMENT when an index is >= the field row count, but highlight indices >= highlight_results_.size() are silently skipped here. A caller passing the same keep_indices hits an explicit error for fields and a silent partial result for highlights. Consider propagating an out-of-range index as an error here too, or at least keeping the two paths' error semantics consistent.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -1772,6 +1772,17 @@ Status
 MilvusClientV2Impl::insert(const InsertRequest& request, InsertResponse& response, bool allow_retry) {
     const auto endpoint = connection_.CurrentEndpoint();
     const auto database_name = connection_.CurrentDbName(request.DatabaseName());
+
+    // Short-circuit empty data like pymilvus: no rows to insert, do not issue the RPC.
+    if (request.ColumnsData().empty() && request.RowsData().empty()) {
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:1777
Low ---- The empty-data short-circuit checks container emptiness rather than row count: an InsertRequest whose ColumnsData() holds a non-empty vector with a single FieldData whose Count() is 0 bypasses the short-circuit and runs the full schema-loading/RPC path with num_rows=0, while pymilvus short-circuits on the row count (len(data)==0) and returns {"insert_count":0,"ids":[]} without an RPC. Consider deriving the condition from the actual row count (rows.size() or fields.front()->Count()) so a 0-row insert/upsert never reaches the wire regardless of how the empty data is constructed.

src/impl/types/SearchResults.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientV2Impl.cpp

src/impl/types/SearchIteratorImpl.cpp
yhmo: This grouping difference is a known consequence of the V1 fallback: the legacy iterator fills the cache from multiple server pages (plus cached leftovers) before FetchPageFromCache assembles a page, so the filter callback sees an assembled page rather than one server page. The V2 path and pymilvus both filter a single server page per callback. For stateless filters the results are identical; only page-dependent filters (per-page dedup/sampling) observe the difference, and those are inherently tied to the iterator implementation the server selects. Rebuilding the V1 fill path to filter per server page would be a larger refactor of a fallback used only with older servers, so I'm keeping the current behavior. The grouping semantics are documented in the V1 Next() comment.

PR #598, opened 2026-09-07
Align MilvusClientV2 load/describe/compaction/stats APIs with pymilvus

yhmo:
## Summary

Align `MilvusClientV2` request/response surfaces with pymilvus `MilvusClient`.

### Request
- `LoadCollectionRequest` / `LoadPartitionsRequest`: add `LoadPriority()` (`"low"` selects low priority; any other value, including `"high"`, defaults to high priority), wired to proto `load_params["load_priority"]`.

### Response
- `describe_collection` (`CollectionDesc`): expose `GetConsistencyLevel()` and `NumPartitions()`; populate `update_timestamp` in the proto conversion.
- `get_collection_stats` (`CollectionStat`): expose the raw `Statistics()` key/value map.
- `get_compaction_plans` (`GetCompactionPlansResponse`): return `CompactionID()` and `State()`.
- `get_refresh_external_collection_progress` / `list_refresh_external_collection_jobs`: surface `external_spec` per job (parsed from the proto string).

### Tests / misc
- Unit tests for all new accessors and proto conversions (`TestTypeUtils`, `TestCollectionDesc`, `TestUtilityResponses`, etc.).
- Mocked integration test for V2 `GetCompactionPlans` in `TestCompact.cpp`.
- Example formatting cleanups.

### DIFF
--- a/src/impl/MilvusClientImpl.cpp
+++ b/src/impl/MilvusClientImpl.cpp
@@ -226,17 +226,23 @@ MilvusClientImpl::DescribeCollection(const std::string& collection_name, Collect
 
     auto post = [&collection_desc](const proto::milvus::DescribeCollectionResponse& response) {
         CollectionSchema schema;
-        ConvertCollectionSchema(response.schema(), schema);
+        auto status = ConvertCollectionSchema(response.schema(), schema);
+        if (!status.IsOk()) {
+            return status;
+        }
         schema.SetShardsNum(response.shards_num());
         collection_desc.SetSchema(std::move(schema));
         collection_desc.SetID(response.collectionid());
+        collection_desc.SetCreatedTime(response.created_timestamp());
+        collection_desc.SetUpdateTime(response.update_timestamp());
+        collection_desc.SetConsistencyLevel(ConsistencyLevelCast(response.consistency_level()));
+        collection_desc.SetNumPartitions(response.num_partitions());
 
         std::vector<std::string> aliases;
         aliases.reserve(response.aliases_size());
         aliases.insert(aliases.end(), response.aliases().begin(), response.aliases().end());
 
         collection_desc.SetAlias(std::move(aliases));
-        collection_desc.SetCreatedTime(response.created_timestamp());
         return Status::OK();
     };
 
@@ -2220,11 +2226,17 @@ MilvusClientImpl::getCollectionDesc(const std::string& endpoint, const std::stri
             CollectionDesc desc;
             auto post = [&desc](const proto::milvus::DescribeCollectionResponse& response) {
                 CollectionSchema schema;
-                ConvertCollectionSchema(response.schema(), schema);
+                auto status = ConvertCollectionSchema(response.schema(), schema);
+                if (!status.IsOk()) {
+                    return status;
+                }
                 schema.SetShardsNum(response.shards_num());
                 desc.SetSchema(std::move(schema));
                 desc.SetID(response.collectionid());
                 desc.SetCreatedTime(response.created_timestamp());
+                desc.SetUpdateTime(response.update_timestamp());
+                desc.SetConsistencyLevel(ConsistencyLevelCast(response.consistency_level()));
+                desc.SetNumPartitions(response.num_partitions());
                 return Status::OK();
             };
             auto status =
--- a/src/impl/MilvusClientV2Impl.cpp
+++ b/src/impl/MilvusClientV2Impl.cpp
@@ -360,6 +360,9 @@ MilvusClientV2Impl::LoadCollection(const LoadCollectionRequest& request) {
         for (const auto& rg : request.TargetResourceGroups()) {
             rpc_request.add_resource_groups(rg);
         }
+        if (!request.LoadPriority().empty()) {
+            (*rpc_request.mutable_load_params())["load_priority"] = request.LoadPriority();
+        }
         return Status::OK();
     };
 
@@ -1223,6 +1226,9 @@ MilvusClientV2Impl::LoadPartitions(const LoadPartitionsRequest& request) {
         }
         for (const auto& rg : request.TargetResourceGroups()) {
             rpc_request.add_resource_groups(rg);
+        }
+        if (!request.LoadPriority().empty()) {
+            (*rpc_request.mutable_load_params())["load_priority"] = request.LoadPriority();
         }
         return Status::OK();
     };
@@ -3138,7 +3144,19 @@ MilvusClientV2Impl::GetCompactionPlans(const GetCompactionPlansRequest& request,
         return Status::OK();
     };
 
-    auto post = [&response](const proto::milvus::GetCompactionPlansResponse& rpc_response) {
+    auto post = [&response, &request](const proto::milvus::GetCompactionPlansResponse& rpc_response) {
+        response.SetCompactionID(request.CompactionID());
+        switch (rpc_response.state()) {
+            case proto::common::CompactionState::Completed:
+                response.SetState(CompactionStateCode::COMPLETED);
+                break;
+            case proto::common::CompactionState::Executing:
+                response.SetState(CompactionStateCode::EXECUTING);
+                break;
+            default:
+                response.SetState(CompactionStateCode::UNKNOWN);
+                break;
+        }
         CompactionPlans plans;
         plans.reserve(rpc_response.mergeinfos_size());
         for (int i = 0; i < rpc_response.mergeinfos_size(); ++i) {
@@ -3444,7 +3462,12 @@ MilvusClientV2Impl::GetRefreshExternalCollectionProgress(const GetRefreshExterna
     };
 
     auto post = [&response](const proto::milvus::GetRefreshExternalCollectionProgressResponse& rpc_response) {
-        response.SetJobInfo(ConvertRefreshExternalCollectionJobInfo(rpc_response.job_info()));
+        RefreshExternalCollectionJobInfo info;
+        auto status = ConvertRefreshExternalCollectionJobInfo(rpc_response.job_info(), info);
+        if (!status.IsOk()) {
+            return status;
+        }
+        response.SetJobInfo(std::move(info));
         return Status::OK();
     };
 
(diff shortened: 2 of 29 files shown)


src/impl/utils/TypeUtils.cpp
```diff
@@ -558,6 +563,10 @@ ConvertDescribeCollectionResponse(const proto::milvus::DescribeCollectionRespons
     collection_desc.SetSchema(std::move(schema));
     collection_desc.SetID(rpc_response.collectionid());
     collection_desc.SetCreatedTime(rpc_response.created_timestamp());
+    collection_desc.SetUpdateTime(rpc_response.update_timestamp());
+    collection_desc.SetConsistencyLevel(ConsistencyLevelCast(rpc_response.consistency_level()));
```
yhmo, line 579: src/impl/utils/TypeUtils.cpp line:567
Low ---- CollectionDesc::GetConsistencyLevel() now surfaces a collection's consistency level, but a collection created with consistency level "Customized" (proto common.ConsistencyLevel.Customized = 4, a value the server accepts for collection create) will be silently reported as BOUNDED: ConsistencyLevelCast() maps every unmapped proto value to Bounded and the C++ ConsistencyLevel enum has no Customized member, whereas pymilvus returns the raw proto value from DescribeCollection. Since this PR's goal is pymilvus parity, consider representing Customized in the C++ enum (and in both ConsistencyLevelCast directions), or at least documenting the fallback so callers can tell it apart from Bounded.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -360,6 +360,9 @@ MilvusClientV2Impl::LoadCollection(const LoadCollectionRequest& request) {
         for (const auto& rg : request.TargetResourceGroups()) {
             rpc_request.add_resource_groups(rg);
         }
+        if (!request.LoadPriority().empty()) {
+            (*rpc_request.mutable_load_params())["load_priority"] = request.LoadPriority();
+        }
```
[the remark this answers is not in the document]
yhmo: Thanks for the suggestion. This is intentional: pymilvus also forwards the load priority value to `load_params` without a client-side whitelist, and the server parses and rejects invalid values. Hard-coding a fixed set here would couple the SDK to a fixed enum that may grow, so we keep the behavior in parity with pymilvus and leave validation to the server.

src/impl/utils/TypeUtils.cpp
yhmo: Agreed that Customized needs attention, but we deliberately do not expose `ConsistencyLevel::CUSTOMIZED` in the public enum: it is only meaningful together with a user-supplied `guarantee_timestamp`, which the SDK request surface does not support yet. Unmapped proto levels therefore fall back to `BOUNDED`. We will revisit Customized support together with the guarantee-timestamp feature.

src/include/milvus/request/partition/LoadPartitionsRequest.h
```diff
@@ -252,6 +252,27 @@ class MILVUS_SDK_API LoadPartitionsRequest {
...
+     * @brief Set load priority.
+     * The load priority of the collection. Possible values are "Low", "Medium" and "High".
+     */
+    void
+    SetLoadPriority(const std::string& load_priority);
+
+    /**
+     * @brief Set load priority.
+     * The load priority of the collection. Possible values are "Low", "Medium" and "High".
+     */
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientV2Impl.cpp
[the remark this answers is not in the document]
yhmo: Thanks for the clarification. I agree that no change is needed here. Resolving this conversation.

src/impl/utils/TypeUtils.cpp
yhmo: Thanks for the clarification. I agree that no change is needed here. Resolving this conversation.

src/impl/utils/TypeUtils.cpp
```diff
@@ -378,6 +378,11 @@ ConvertRefreshExternalCollectionJobInfo(const proto::milvus::RefreshExternalColl
     info.SetExternalSource(rpc_info.external_source());
     info.SetStartTime(static_cast<uint64_t>(rpc_info.start_time()));
     info.SetEndTime(static_cast<uint64_t>(rpc_info.end_time()));
+    if (rpc_info.external_spec().empty()) {
+        info.SetExternalSpec(nullptr);
+    } else {
+        info.SetExternalSpec(nlohmann::json::parse(rpc_info.external_spec()));
```
yhmo, line 386: src/impl/utils/TypeUtils.cpp line:384
Low ---- ConvertRefreshExternalCollectionJobInfo now parses external_spec with nlohmann::json::parse, which throws parse_error on malformed server data. That exception propagates as an uncaught C++ exception out of the public GetRefreshExternalCollectionProgress and ListRefreshExternalCollectionJobs entry points (in the list case, a single malformed job aborts the whole response). This matches the pre-existing DescribeCollection convention (TypeUtils.cpp also throws on an invalid schema external_spec), so the behavior is consistent with the repo, but it is a newly exposed throw path on two public V2 APIs. Consider catching nlohmann::json::parse_error and returning a Status instead of throwing, or explicitly documenting the throw contract on these APIs.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -360,6 +360,9 @@ MilvusClientV2Impl::LoadCollection(const LoadCollectionRequest& request) {
         for (const auto& rg : request.TargetResourceGroups()) {
             rpc_request.add_resource_groups(rg);
         }
+        if (!request.LoadPriority().empty()) {
+            (*rpc_request.mutable_load_params())["load_priority"] = request.LoadPriority();
```
yhmo, line 364: src/impl/MilvusClientV2Impl.cpp line:364
Low ---- The load_priority wiring has no RPC-level test: the new tests in TestCollectionRequests.cpp / TestPartitionRequests.cpp only exercise request getter/setter round-trips, so nothing proves load_params["load_priority"] is actually sent on the wire (and omitted when unset). The new GetCompactionPlansV2 mocked test in this PR is a good example of the repo style for asserting request fields; a similar mocked test asserting load_params["load_priority"] == "low" for both LoadCollection and LoadPartitions would lock in this parity feature.

src/impl/utils/TypeUtils.cpp
yhmo: This matches the existing repo convention: ConvertDescribeCollectionResponse already parses external_spec with nlohmann::json::parse and throws parse_error on malformed data, and that throw is covered by an explicit EXPECT_THROW test. The server is the only producer of this JSON, so malformed input indicates a server bug or wire corruption rather than a user-controllable path. Catching and returning a Status here would diverge from the DescribeCollection behavior, so we keep the consistent throw contract.

src/impl/MilvusClientV2Impl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/TypeUtils.cpp
yhmo: Thanks for the clarification. I agree that no change is needed here. Resolving this conversation.

src/impl/MilvusClientImpl.cpp
```diff
@@ -226,7 +226,10 @@ MilvusClientImpl::DescribeCollection(const std::string& collection_name, Collect
 
     auto post = [&collection_desc](const proto::milvus::DescribeCollectionResponse& response) {
         CollectionSchema schema;
-        ConvertCollectionSchema(response.schema(), schema);
+        auto status = ConvertCollectionSchema(response.schema(), schema);
```
yhmo, line 229: src/impl/MilvusClientImpl.cpp line:229
Low ---- These V1 describe paths (this post handler and the getCollectionDesc SchemaCache loader at ~line 2226) now propagate the new Status from ConvertCollectionSchema but still do not populate the fields this PR adds to the shared CollectionDesc type. ConvertDescribeCollectionResponse (the V2 path used by MilvusClientV2Impl::describeCollection) now sets consistency_level, num_partitions and update_timestamp, so the same public type returns the defaults (BOUNDED / 0) when obtained through MilvusClient::DescribeCollection even though the server reports real values. Consider setting the same three fields here too (consistency_level(), num_partitions(), update_timestamp() are all available on the proto response) so the V1 and V2 describe paths stay consistent.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -3138,7 +3144,19 @@ MilvusClientV2Impl::GetCompactionPlans(const GetCompactionPlansRequest& request,
...
     };
 
-    auto post = [&response](const proto::milvus::GetCompactionPlansResponse& rpc_response) {
+    auto post = [&response, &request](const proto::milvus::GetCompactionPlansResponse& rpc_response) {
+        response.SetCompactionID(request.CompactionID());
+        switch (rpc_response.state()) {
+            case proto::common::CompactionState::Completed:
+                response.SetState(CompactionStateCode::COMPLETED);
+                break;
+            case proto::common::CompactionState::Executing:
```
yhmo, line 3153: src/impl/MilvusClientV2Impl.cpp line:3153
Low ---- The new GetCompactionPlansV2 mocked test only exercises the Completed branch of this switch; the Executing and default (UNKNOWN) mappings are untested, so a mis-mapping such as Executing -> COMPLETED would pass CI undetected. Consider extending the mock to also return proto::common::CompactionState::Executing (and an unset state) and assert the corresponding CompactionStateCode values.

src/impl/MilvusClientImpl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientV2Impl.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #599, opened 2026-09-08
Add exception barrier to public SDK calls

yhmo:
## Summary

Ensure public `MilvusClientV2`/`MilvusClient` calls never throw: exceptions escaping the RPC pipeline, the gRPC layer, or direct stub callers are converted into an `UNKNOWN_ERROR` `Status` at the boundary.

### Changes
- `ConnectionHandler::apiHandler`: wraps the validate/pre/rpc/wait/post pipeline in a `try/catch`; body moves to a private `invokeInternal()`. Covers `std::exception` and `catch(...)`.
- `MilvusConnection::grpcCall`: same barrier for RPCs that go through `grpcCall` (e.g. CheckHealth, GetLoadingProgress); body moves to a private `grpcCallInternal()`.
- `MilvusConnection::Connect` and `MilvusConnection::DumpMessages`: these call the gRPC stub directly (handshake and streaming respectively), so they now carry the same barrier around their handshake/stream bodies.
- Shared `StatusFromException` / `StatusFromUnknownException` helpers give the conversion a single definition across apiHandler, grpcCall, Connect, DumpMessages and SchemaCache, with dedicated unit tests. The helpers document that `UNKNOWN_ERROR` from the barrier does not imply the server-side operation was not applied (a response-conversion failure after a successful RPC).
- `examples/src/v2/rbac.cpp`: use `object_name "*"` for the Global `CreateDatabase` grant/revoke — the empty object name is rejected by the current server.
- `PrivilegeRequest` docs and the `RevokePrivilege` mocked test now encode `"*"` for the Global scope.
- New mocked tests `ExceptionBarrierConvertsStdExceptionToStatus` / `ExceptionBarrierConvertsNonStdExceptionToStatus` exercise both catch branches via throwing validate lambdas.

### Tests
- `testing-ut`: 897 passed.
- `testing-it`: 52 passed (Connect/DumpMessages/Privilege/ExceptionBarrier/load/describe/session).

### DIFF
--- a/examples/src/v2/rbac.cpp
+++ b/examples/src/v2/rbac.cpp
@@ -133,7 +133,7 @@ main(int argc, char* argv[]) {
     status = client->GrantPrivilege(milvus::GrantPrivilegeRequest()
                                         .WithRoleName(role_name)
                                         .WithObjectType("Global")
-                                        .WithObjectName("")
+                                        .WithObjectName("*")
                                         .WithPrivilege("CreateDatabase"));
     util::CheckStatus("grant global privilege to role: " + role_name, status);
 
@@ -218,7 +218,7 @@ main(int argc, char* argv[]) {
     status = client->RevokePrivilege(milvus::RevokePrivilegeRequest()
                                          .WithRoleName(role_name)
                                          .WithObjectType("Global")
-                                         .WithObjectName("")
+                                         .WithObjectName("*")
                                          .WithPrivilege("CreateDatabase"));
     util::CheckStatus("revoke global privilege from role: " + role_name, status);
 
--- a/src/impl/MilvusConnection.h
+++ b/src/impl/MilvusConnection.h
@@ -35,6 +35,7 @@
 #include "milvus/ClientRequestContext.h"
 #include "milvus/ClientTelemetry.h"
 #include "milvus/Status.h"
+#include "utils/RpcUtils.h"
 #include "milvus/types/ConnectParam.h"
 #include "schema.pb.h"
 
@@ -560,6 +561,24 @@ class MilvusConnection {
     grpcCall(const char* name,
              grpc::Status (proto::milvus::MilvusService::Stub::*func)(grpc::ClientContext*, const Request&, Response*),
              const Request& request, Response& response, const GrpcContextOptions& options) {
+        // Defense-in-depth: most callers already run inside ConnectionHandler::apiHandler's
+        // barrier, but lifecycle APIs such as ConnectionHandler::GetLoadingProgress call
+        // grpcCall directly and rely on these branches.
+        try {
+            return grpcCallInternal(name, func, request, response, options);
+        } catch (const std::exception& e) {
+            return StatusFromException(e);
+        } catch (...) {
+            return StatusFromUnknownException();
+        }
+    }
+
+    template <typename Request, typename Response>
+    Status
+    grpcCallInternal(const char* name,
+                     grpc::Status (proto::milvus::MilvusService::Stub::*func)(grpc::ClientContext*, const Request&,
+                                                                             Response*),
+                     const Request& request, Response& response, const GrpcContextOptions& options) {
         (void)name;
         std::shared_ptr<proto::milvus::MilvusService::Stub> stub;
         {
--- a/src/impl/utils/ConnectionHandler.h
+++ b/src/impl/utils/ConnectionHandler.h
@@ -212,6 +212,23 @@ class ConnectionHandler {
                Status (MilvusConnection::*rpc)(const Request&, Response&, const GrpcOpts&),
                std::function<Status(const Response&)> wait_for_status, std::function<Status(const Response&)> post,
                uint64_t rpc_timeout_ms = 0) {
+        try {
+            return invokeInternal(validate, pre, rpc, wait_for_status, post, rpc_timeout_ms);
+        } catch (const std::exception& e) {
+            return StatusFromException(e);
+        } catch (...) {
+            return StatusFromUnknownException();
+        }
+    }
+
+    // Runs the shared validate/pre/rpc/wait/post pipeline. apiHandler wraps this in an
+    // exception barrier so public SDK calls always return a Status instead of throwing.
+    template <typename Request, typename Response>
+    Status
+    invokeInternal(const std::function<Status(void)>& validate, std::function<Status(Request&)> pre,
+                   Status (MilvusConnection::*rpc)(const Request&, Response&, const GrpcOpts&),
+                   std::function<Status(const Response&)> wait_for_status, std::function<Status(const Response&)> post,
+                   uint64_t rpc_timeout_ms) {
         MilvusConnectionPtr connection;
         RetryParam retry_param;
         uint64_t timeout = 0;
--- a/codecov.yml
+++ b/codecov.yml
@@ -37,6 +37,7 @@ ignore:
   - src/include/milvus/thirdparty
   - test
   - thirdparty
+  - tutorial
   - "**/*.txt"
   - "**/*.md"
   - "**/*.yml"
--- a/src/impl/MilvusConnection.cpp
+++ b/src/impl/MilvusConnection.cpp
@@ -177,60 +177,59 @@ MilvusConnection::Connect(const ConnectParam& param, const std::string& runtime_
 
     // grpc channel has been create, now we call the proto::milvus::MilvusClient::Connect() interface
     // to send some basic information of client to the server, including the sdk type, version, etc.
-    proto::milvus::ConnectRequest rpc_request;
-    auto client_info = rpc_request.mutable_client_info();
-    client_info->set_sdk_type("CPP");
-    client_info->set_user(param.Username());
-    client_info->set_sdk_version(GetBuildVersion());
-    client_info->set_host(param.Host());
-
-    auto now = std::chrono::system_clock::now();
-    std::time_t now_time = std::chrono::system_clock::to_time_t(now);
-    std::tm* local_time = std::localtime(&now_time);
-    std::stringstream ss;
-    ss << std::put_time(local_time, "%Y-%m-%d %H:%M:%S");
-    client_info->set_local_time(ss.str());
-
-    // the default value of ConnectTimeout is 10 seconds, means if the server could not return response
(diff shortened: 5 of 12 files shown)


examples/src/v2/rbac.cpp
```diff
@@ -133,7 +133,7 @@ main(int argc, char* argv[]) {
     status = client->GrantPrivilege(milvus::GrantPrivilegeRequest()
                                         .WithRoleName(role_name)
                                         .WithObjectType("Global")
-                                        .WithObjectName("")
+                                        .WithObjectName("*")
```
yhmo, line 136: examples/src/v2/rbac.cpp line:136
Medium ---- The public API docs still contradict this change: src/include/milvus/request/rbac/PrivilegeRequest.h (lines 27-31 and line 77) states the object name is "empty for the Global scope", and the mocked test test/it/v2/TestPrivilege.cpp:79 still asserts object_name() == "" for a Global grant. The current server rejects an empty object name (internal/proxy/util.go ValidateObjectName -> validateName returns "object name should be not empty"), which is exactly why this example must use "*" - and this matches pymilvus examples/rbac.py and the Java SDK RBACExample.java, which both use object_name "*" for Global-scoped grants. Please update the PrivilegeRequest.h doc comment (and the mocked test if it encodes the rejected "" form) so future callers following the docs do not hit the same server rejection this PR fixes.

src/impl/MilvusConnection.h
```diff
@@ -560,6 +560,22 @@ class MilvusConnection {
     grpcCall(const char* name,
              grpc::Status (proto::milvus::MilvusService::Stub::*func)(grpc::ClientContext*, const Request&, Response*),
              const Request& request, Response& response, const GrpcContextOptions& options) {
+        try {
```
yhmo, line 567: src/impl/MilvusConnection.h line:563
Low ---- The PR description does not describe this PR main change: the body only mentions the rbac.cpp object-name fix and a SearchAggregation.h AggDirection alias, but SearchAggregation.h is not modified in this diff and the exception barrier added in this hunk (and in src/impl/utils/ConnectionHandler.h) is not mentioned at all. Please update the description to match the actual scope so the change is reviewable from the PR page alone.

src/impl/MilvusConnection.h
```diff
@@ -560,6 +560,22 @@ class MilvusConnection {
...
+            return grpcCallInternal(name, func, request, response, options);
+        } catch (const std::exception& e) {
+            return {StatusCode::UNKNOWN_ERROR, "Unexpected SDK exception: " + std::string(e.what())};
+        } catch (...) {
+            return {StatusCode::UNKNOWN_ERROR, "Unexpected SDK exception"};
+        }
+    }
+
+ private:
```
yhmo: src/impl/MilvusConnection.h line:572
Low ---- Redundant access specifier: grpcCall and grpcCallInternal are already inside the private section (private: at line 533 above), so this second private: is unnecessary. The same redundant private: was added before invokeInternal in src/impl/utils/ConnectionHandler.h (line 224). Harmless, but can be dropped for cleanliness.

src/impl/utils/ConnectionHandler.h
```diff
@@ -212,6 +212,24 @@ class ConnectionHandler {
                Status (MilvusConnection::*rpc)(const Request&, Response&, const GrpcOpts&),
                std::function<Status(const Response&)> wait_for_status, std::function<Status(const Response&)> post,
                uint64_t rpc_timeout_ms = 0) {
+        try {
```
yhmo, line 215: src/impl/utils/ConnectionHandler.h line:215
Low ---- The new exception barrier has no test coverage. The sibling barrier implementations are tested for the throwing path (test/ut/utils/TestSchemaCache.cpp forces a throwing schema loader and asserts the UNKNOWN_ERROR status; OptimizeTask uses the same pattern), but nothing here exercises the catch branches. Since apiHandler accepts std::function callbacks, a mocked test could pass a validate/wait_for_status/post lambda that throws and assert the resulting "Unexpected SDK exception" Status, covering both the std::exception and catch(...) branches.

examples/src/v2/rbac.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusConnection.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusConnection.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/ConnectionHandler.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusConnection.h
```diff
@@ -560,6 +560,21 @@ class MilvusConnection {
     grpcCall(const char* name,
              grpc::Status (proto::milvus::MilvusService::Stub::*func)(grpc::ClientContext*, const Request&, Response*),
              const Request& request, Response& response, const GrpcContextOptions& options) {
+        try {
+            return grpcCallInternal(name, func, request, response, options);
+        } catch (const std::exception& e) {
+            return {StatusCode::UNKNOWN_ERROR, "Unexpected SDK exception: " + std::string(e.what())};
```
yhmo: src/impl/MilvusConnection.h line:566
Medium ---- This catch branch (and the catch(...) branch below) is never exercised: TestExceptionBarrier.cpp only throws from the apiHandler validate lambda, so codecov flags these 4 lines as the uncovered part of the patch. The earlier thread on ConnectionHandler.h:215 (resolved) added tests for the apiHandler barrier only, not for grpcCall. Because every grpcCall caller except ConnectionHandler::GetLoadingProgress already runs inside apiHandler's barrier, this second barrier is largely defense-in-depth - please add a test, or a comment explaining which direct callers rely on these branches, so they do not silently drift.

src/impl/utils/ConnectionHandler.h
```diff
@@ -212,6 +212,23 @@ class ConnectionHandler {
...
+            return invokeInternal(validate, pre, rpc, wait_for_status, post, rpc_timeout_ms);
+        } catch (const std::exception& e) {
+            return {StatusCode::UNKNOWN_ERROR, "Unexpected SDK exception: " + std::string(e.what())};
```
yhmo: src/impl/utils/ConnectionHandler.h line:218
Low ---- The exception-to-Status conversion block (try/catch std::exception + catch(...)) is now duplicated verbatim here in apiHandler and in MilvusConnection::grpcCall (src/impl/MilvusConnection.h:563-570); SchemaCache.cpp:116-118 carries a similar conversion. If the message wording or status code ever changes, the copies must be kept in sync. Consider extracting a small shared helper (e.g. a Status-from-exception factory) so the barrier has a single definition.

src/impl/MilvusConnection.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/ConnectionHandler.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusConnection.h
```diff
@@ -560,6 +561,24 @@ class MilvusConnection {
     grpcCall(const char* name,
              grpc::Status (proto::milvus::MilvusService::Stub::*func)(grpc::ClientContext*, const Request&, Response*),
              const Request& request, Response& response, const GrpcContextOptions& options) {
+        // Defense-in-depth: most callers already run inside ConnectionHandler::apiHandler's
+        // barrier, but lifecycle APIs such as ConnectionHandler::GetLoadingProgress call
```
yhmo, line 565: src/impl/MilvusConnection.h line:565
Medium ---- This comment (and the PR summary's claim that Connect/Disconnect/CheckHealth are covered by grpcCall) overstates the actual coverage. MilvusConnection::Connect performs its handshake via stub->Connect directly (MilvusConnection.cpp:203), outside every try block and outside grpcCall, and the streaming public API MilvusClientV2::DumpMessages calls MilvusConnection::DumpMessages, which reads the server stream directly without grpcCall or apiHandler. An exception thrown while building the client info in Connect, or from the on_message callback / SDK conversion inside the DumpMessages read loop, still escapes the public MilvusClient::Connect / MilvusClientV2::DumpMessages calls. Since the whole point of this PR is a uniform "public calls never throw" boundary, please route these two paths through a guarded wrapper as well, or narrow the comment/summary to state exactly which residual paths may still throw.

src/impl/utils/ConnectionHandler.h
```diff
@@ -212,6 +212,23 @@ class ConnectionHandler {
                Status (MilvusConnection::*rpc)(const Request&, Response&, const GrpcOpts&),
                std::function<Status(const Response&)> wait_for_status, std::function<Status(const Response&)> post,
                uint64_t rpc_timeout_ms = 0) {
+        try {
+            return invokeInternal(validate, pre, rpc, wait_for_status, post, rpc_timeout_ms);
```
yhmo, line 216: src/impl/utils/ConnectionHandler.h line:216
Low ---- The barrier converts a post-RPC exception into UNKNOWN_ERROR, which is indistinguishable from a transport failure even when the server-side operation already succeeded. For example, if a response-conversion post lambda throws after a successful insert/upsert/delete, the caller sees UNKNOWN_ERROR and a retry would duplicate the mutation. The SDK already throws std::runtime_error from response conversion code (e.g. SearchResults.cpp and DqlUtils.cpp binary-string parsing), so this is not purely hypothetical for the read path. Consider returning OK (with a log) for exceptions that escape post after an OK RPC, or at least documenting that UNKNOWN_ERROR from the barrier does not imply the server operation was not applied.

src/impl/MilvusConnection.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/utils/ConnectionHandler.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #601, opened 2026-09-09
fix: harden client telemetry (#587, #588)

yhmo:
- Treat ttl_seconds as an ignored push_config field (#587)
- Harden client telemetry control flow: make very large heartbeat intervals safe and cancellable across platforms, fence command batches across telemetry transport rebinding, canonicalize ACK IDs, wake the unsupported backoff on channel replacement, and bound show_errors truncation including UTF-8 payloads (#588)

### DIFF
--- a/src/impl/ClientTelemetry.cpp
+++ b/src/impl/ClientTelemetry.cpp
@@ -68,12 +68,18 @@ constexpr size_t kSnapshotLimit = 4096;
 constexpr uint64_t kSamplingScale = 1000000000ULL;
 constexpr size_t kMaxReplyBytes = 1024 * 1024;
 constexpr uint64_t kMaxUnsupportedBackoffMs = 30 * 60 * 1000;
+constexpr uint64_t kMaxHeartbeatIntervalMs = static_cast<uint64_t>(std::numeric_limits<int64_t>::max());
+// condition_variable implementations commonly form an absolute deadline internally.
+// Keep each addition comfortably inside every supported platform clock and represent
+// longer intervals as stop/rebind-aware chunks.
+constexpr uint64_t kMaxWaitChunkMs = 24ULL * 60 * 60 * 1000;
 
 TelemetryConfig
 NormalizedTelemetryConfig(TelemetryConfig config) {
     if (config.heartbeat_interval_ms == 0) {
         config.heartbeat_interval_ms = 10000;
     }
+    config.heartbeat_interval_ms = std::min(config.heartbeat_interval_ms, kMaxHeartbeatIntervalMs);
     config.sampling_rate = std::max(0.0, std::min(1.0, config.sampling_rate));
     if (config.error_max_count == 0) {
         config.error_max_count = 100;
@@ -92,6 +98,27 @@ int64_t
 NowMillis() {
     return std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch())
         .count();
+}
+
+int64_t
+SaturatingWindowStart(int64_t end_time, uint64_t interval_ms) {
+    const auto signed_interval = static_cast<int64_t>(std::min(interval_ms, kMaxHeartbeatIntervalMs));
+    if (end_time < std::numeric_limits<int64_t>::min() + signed_interval) {
+        return std::numeric_limits<int64_t>::min();
+    }
+    return end_time - signed_interval;
+}
+
+size_t
+Utf8SafePrefixLength(const std::string& value, size_t requested) {
+    auto length = std::min(requested, value.size());
+    if (length == value.size()) {
+        return length;
+    }
+    while (length > 0 && (static_cast<unsigned char>(value[length]) & 0xC0) == 0x80) {
+        --length;
+    }
+    return length;
 }
 
 std::string
@@ -443,10 +470,10 @@ FailedReply(const std::string& command_id, const std::string& error) {
 class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientTelemetryManager::Impl> {
  public:
     Impl(const TelemetryConfig& value, const std::string& runtime_client_id)
-        : config(NormalizedTelemetryConfig(value)),
-          connection_config(config),
-          stable_client_id(!value.client_id.empty()),
-          client_id(stable_client_id ? value.client_id
+        : config_(NormalizedTelemetryConfig(value)),
+          connection_config_(config_),
+          stable_client_id_(!value.client_id.empty()),
+          client_id_(stable_client_id_ ? value.client_id
                                      : (runtime_client_id.empty() ? RandomUuid() : runtime_client_id)) {
         RegisterDefaultHandlers();
     }
@@ -469,123 +496,134 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
         std::string new_sdk_version = version;
         std::string new_connection_scope = scope;
 
-        std::lock_guard<std::recursive_mutex> command_lock(command_mutex);
-        std::lock_guard<std::mutex> lock(mutex);
-        stub = std::move(new_stub);
-        ++channel_generation;
-        username = std::move(new_username);
-        database = std::move(new_database);
-        uri = std::move(new_uri);
-        sdk_version = std::move(new_sdk_version);
-        connection_scope = std::move(new_connection_scope);
+        std::lock_guard<std::recursive_mutex> command_lock(command_mutex_);
+        std::lock_guard<std::mutex> lock(mutex_);
+        stub_ = std::move(new_stub);
+        ++channel_generation_;
+        // A newly attached transport must be probed immediately. Carrying an
+        // UNIMPLEMENTED backoff from the retired endpoint can otherwise keep a
+        // supporting endpoint silent for up to thirty minutes.
+        unsupported_streak_ = 0;
+        ++wait_revision_;
+        username_ = std::move(new_username);
+        database_ = std::move(new_database);
+        uri_ = std::move(new_uri);
+        sdk_version_ = std::move(new_sdk_version);
+        connection_scope_ = std::move(new_connection_scope);
+        condition_.notify_all();
     }
 
     void
     Start() {
-        std::unique_lock<std::mutex> lock(mutex);
-        if (ready) {
+        std::unique_lock<std::mutex> lock(mutex_);
+        if (ready_) {
             return;
         }
-        const bool called_from_worker = worker_running && worker_id == std::this_thread::get_id();
+        const bool called_from_worker = worker_running_ && worker_id_ == std::this_thread::get_id();
         if (called_from_worker) {
             // Stop()/Start() is used by reconnect handlers. Reuse this worker only
             // when no external caller has claimed it for joining. An external stop
             // always wins over a self-restart that races with it.
-            if (join_in_progress || external_stop_requested) {
+            if (join_in_progress_ || external_stop_requested_) {
                 return;
             }
-            ready = true;
-            control_plane_activated = control_plane_activated || config.enabled;
-            if (!control_plane_activated) {
-                stopped = true;
+            ready_ = true;
+            control_plane_activated_ = control_plane_activated_ || config_.enabled;
+            if (!control_plane_activated_) {
+                stopped_ = true;
                 return;
             }
-            stopped = false;
+            stopped_ = false;
             return;
         }
 
-        condition.wait(lock, [this]() { return !join_in_progress; });
-        if (ready) {
+        condition_.wait(lock, [this]() { return !join_in_progress_; });
+        if (ready_) {
             return;
         }
-        if (worker.joinable()) {
-            join_in_progress = true;
-            auto previous_worker = std::move(worker);
+        if (worker_.joinable()) {
+            join_in_progress_ = true;
+            auto previous_worker = std::move(worker_);
             lock.unlock();
             previous_worker.join();
             lock.lock();
-            worker_id = {};
-            worker_running = false;
-            join_in_progress = false;
-            condition.notify_all();
-            if (ready) {
+            worker_id_ = {};
+            worker_running_ = false;
+            join_in_progress_ = false;
+            condition_.notify_all();
+            if (ready_) {
                 return;
             }
         }
-        external_stop_requested = false;
-        ready = true;
+        external_stop_requested_ = false;
+        ready_ = true;
         // Initial enabled=false is an explicit opt-out and creates no control-plane traffic.
         // Once activated, keep the control plane sticky across dynamic disable and
         // Stop/Start reconnects so the server can later re-enable telemetry.
-        control_plane_activated = control_plane_activated || config.enabled;
-        if (!control_plane_activated) {
-            stopped = true;
+        control_plane_activated_ = control_plane_activated_ || config_.enabled;
+        if (!control_plane_activated_) {
+            stopped_ = true;
             return;
         }
-        stopped = false;
-        worker_running = true;
+        stopped_ = false;
+        worker_running_ = true;
         try {
             auto self = shared_from_this();
-            worker = std::thread([self = std::move(self)]() { self->HeartbeatLoop(); });
-            worker_id = worker.get_id();
+            worker_ = std::thread([self = std::move(self)]() { self->HeartbeatLoop(); });
+            worker_id_ = worker_.get_id();
         } catch (...) {
             // Telemetry startup is best-effort. Restore a clean, retryable stopped state instead
             // of leaking an exception through Connect() or leaving worker_running without a thread.
-            ready = false;
-            stopped = true;
-            worker_running = false;
-            worker_id = {};
-            condition.notify_all();
+            ready_ = false;
+            stopped_ = true;
+            worker_running_ = false;
+            worker_id_ = {};
+            condition_.notify_all();
         }
     }
 
     void
     Stop() {
-        std::unique_lock<std::mutex> lock(mutex);
-        stopped = true;
-        ready = false;
-        condition.notify_all();
+        std::unique_lock<std::mutex> lock(mutex_);
+        stopped_ = true;
+        ready_ = false;
+        condition_.notify_all();
 
-        const bool called_from_worker = worker_running && worker_id == std::this_thread::get_id();
+        const bool called_from_worker = worker_running_ && worker_id_ == std::this_thread::get_id();
         if (called_from_worker) {
             return;
         }
 
-        external_stop_requested = true;
-        condition.wait(lock, [this]() { return !join_in_progress; });
-        if (!worker.joinable()) {
-            worker_id = {};
-            worker_running = false;
+        external_stop_requested_ = true;
+        condition_.wait(lock, [this]() { return !join_in_progress_; });
+        if (!worker_.joinable()) {
+            worker_id_ = {};
+            worker_running_ = false;
             return;
         }
 
         // Move the thread object while holding the state mutex. This gives
         // exactly one external caller ownership of join(); other Stop()/Start()
         // calls wait on join_in_progress instead of touching the same std::thread.
-        join_in_progress = true;
-        auto current_worker = std::move(worker);
+        join_in_progress_ = true;
+        auto current_worker = std::move(worker_);
         lock.unlock();
         current_worker.join();
         lock.lock();
-        worker_id = {};
-        worker_running = false;
-        join_in_progress = false;
-        condition.notify_all();
+        worker_id_ = {};
+        worker_running_ = false;
+        join_in_progress_ = false;
+        condition_.notify_all();
     }
 
     void
     HeartbeatLoop() {
         while (true) {
+            uint64_t heartbeat_revision;
+            {
+                std::lock_guard<std::mutex> lock(mutex_);
+                heartbeat_revision = wait_revision_;
+            }
             try {
                 CreateSnapshot();
                 SendHeartbeat();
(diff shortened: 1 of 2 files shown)


src/impl/ClientTelemetry.cpp
```diff
@@ -824,9 +884,18 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
...
             }
             auto reply = HandleCommand(command);
-            std::lock_guard<std::mutex> lock(mutex);
-            executed_commands[command.command_id] = command.create_time;
-            pending_replies.push_back(std::move(reply));
+            {
+                std::lock_guard<std::mutex> lock(mutex);
+                executed_commands[command.command_id] = command.create_time;
+                pending_replies.push_back(std::move(reply));
+                if (expected_generation != 0 && expected_generation != channel_generation) {
```
yhmo: src/impl/ClientTelemetry.cpp line:891
Medium ---- The generation fencing can make the server receive a spurious SUCCESS ack for an already-processed command. When the transport rebinds mid-batch, this return leaves last_command_timestamp unchanged, so the server redelivers the whole batch; on redelivery the completed command hits the skip path (line 878) which pushes SuccessReply(command_id) unconditionally -- even if the real reply that was already queued and sent for that command was a failure. The server therefore sees the command's actual (failed) result followed by a contradictory success ack. Skipped redeliveries should not mint a fresh SuccessReply when the command's real reply was already sent; consider re-queueing the original reply, or skipping the ack entirely for fenced redeliveries.

src/impl/ClientTelemetry.cpp
```diff
@@ -68,12 +68,18 @@ constexpr size_t kSnapshotLimit = 4096;
...
+// Keep each addition comfortably inside every supported platform clock and represent
+// longer intervals as stop/rebind-aware chunks.
+constexpr uint64_t kMaxWaitChunkMs = 24ULL * 60 * 60 * 1000;
 
 TelemetryConfig
 NormalizedTelemetryConfig(TelemetryConfig config) {
     if (config.heartbeat_interval_ms == 0) {
         config.heartbeat_interval_ms = 10000;
     }
+    config.heartbeat_interval_ms = std::min(config.heartbeat_interval_ms, kMaxHeartbeatIntervalMs);
```
yhmo, line 82: src/impl/ClientTelemetry.cpp line:82
Low ---- This clamp only protects the constructor path. The push_config handler (line 934) still reads heartbeat_interval_ms with payload["heartbeat_interval_ms"].get<int64_t>() and assigns it un-clamped, so a pushed unsigned value above int64 max (e.g. 18446744073709551615) makes get<int64_t>() throw json::out_of_range and the whole command fails with a FailedReply instead of being accepted or clamped -- inconsistent with the PR's goal of making very large heartbeat intervals safe. Since the test MaximumHeartbeatIntervalsAreSafeAndDoNotSpin only exercises exactly int64 max, consider clamping in push_config too (and reading as uint64_t).

src/impl/ClientTelemetry.cpp
yhmo: src/impl/ClientTelemetry.cpp line:891
Low ---- On this fenced abort the watermark/config-hash advance and the `executed_commands` prune below are skipped, so markers for commands that ran before the fence stay until a later *full* batch completes. Under repeated rebinds that abort batches, the executed-ID set can grow beyond the usual watermark-based bound. This is a sound trade-off (advancing the watermark could drop the unprocessed suffix), so just confirming the transient growth is acceptable -- consider a one-line note if rebind flapping is expected.

src/impl/ClientTelemetry.cpp
```diff
@@ -631,18 +684,29 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
...
+        bool all_collections_enabled_snapshot = false;
+        std::unordered_set<std::string> enabled_collections_snapshot;
+        {
+            std::lock_guard<std::mutex> lock(mutex);
+            if (!config.enabled) {
+                return;
+            }
+            snapshot.end_time = NowMillis();
+            snapshot.timestamp = last_snapshot_end == 0 || last_snapshot_end > snapshot.end_time
+                                     ? SaturatingWindowStart(snapshot.end_time, config.heartbeat_interval_ms)
```
yhmo: src/impl/ClientTelemetry.cpp line:699
Low ---- With a clamped maximum interval (e.g. int64 max after push_config), the FIRST snapshot window computes snapshot.timestamp = SaturatingWindowStart(end_time, ~int64-max) ≈ -9.2e18 ms (roughly year -292M). That value is exposed verbatim through show_latency_history (timestamp field, line 1192) and MetricsSnapshots(), so a consumer querying recent history sees a nonsensical window start (and the query filter snapshot.timestamp <= end always matches it). Consider capping the reported window start to something meaningful (e.g. end_time - kMaxHistoryRangeMs) when the interval exceeds a sane bound, even though the wait itself is correctly chunked.

src/impl/ClientTelemetry.cpp
```diff
@@ -604,6 +642,12 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
             if (stopped) {
                 break;
             }
+            if (heartbeat_revision != wait_revision) {
+                // AttachChannel may have replaced the transport while the RPC
+                // was in flight, before there was a waiter to notify. Probe the
+                // replacement immediately instead of sleeping the old interval.
+                continue;
```
yhmo, line 649: src/impl/ClientTelemetry.cpp line:649
Low ---- The revision-change `continue` discards the remaining sleep and the accumulated unsupported backoff, then immediately re-runs CreateSnapshot + SendHeartbeat. Under a reconnect storm that calls AttachChannel repeatedly (each call bumps wait_revision), the single worker thread runs full heartbeat cycles back-to-back with no throttling, and each cycle re-serializes the entire latest_snapshot.metrics and pending_replies while holding the mutex (the existing TODO at line 769). A short settle/coalesce window before re-probing would avoid hammering a flapping endpoint while preserving the immediate-probe fix.

src/impl/ClientTelemetry.cpp
```diff
@@ -997,12 +1082,52 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
...
             auto encoded = result.dump();
-            while (encoded.size() > kMaxReplyBytes && result.size() == 1 &&
-                   result.at(0).value("error_msg", std::string{}).size() > 1) {
-                auto message = result.at(0).at("error_msg").get<std::string>();
-                result.at(0)["error_msg"] =
-                    message.substr(0, std::max<size_t>(1, message.size() / 2)) + "...(truncated)";
+            constexpr const char* kTruncated = "...(truncated)";
+            constexpr std::array<const char*, 4> kTruncatableFields = {"error_msg", "collection", "operation",
+                                                                       "request_id"};
+            while (encoded.size() > kMaxReplyBytes && result.size() == 1) {
```
yhmo, line 1107: src/impl/ClientTelemetry.cpp line:1088
Low ---- The show_errors truncation runs on the single telemetry worker thread and re-serializes the whole JSON on every iteration: the array-halving loop re-calls result.dump() each halving (line 1081) and the per-field truncation loop re-dumps after every truncation (line 1121). With a large error_max_count and large error payloads this is avoidable O(N) serialization work on the control-plane thread, during which pending_replies/snapshots access waits behind the mutex. max_count is also unbounded (only the stored error count caps it). Consider computing the dump incrementally or bounding the iterations.

src/impl/ClientTelemetry.cpp
yhmo: Fixed: ProcessCommands now stores each executed command's actual reply (ExecutedCommand{create_time, reply}) and re-queues that reply on fenced redelivery instead of minting a SuccessReply, so an already-reported failure is never contradicted by a success ACK. Covered by the new FencedRedeliveryRequeuesOriginalFailedReply test (verified to fail against the old code path).

src/impl/ClientTelemetry.cpp
yhmo: Fixed: push_config now reads heartbeat_interval_ms as uint64_t and clamps it to kMaxHeartbeatIntervalMs before applying, so a pushed unsigned value above int64 max is accepted and clamped instead of failing the whole command. Covered by the new PushConfigClampsMaximumInterval test.

src/impl/ClientTelemetry.cpp
yhmo: Fixed: CreateSnapshot now caps the reported window start at end_time - kMaxHistoryRangeMs, so a maximum heartbeat interval no longer surfaces a saturated INT64_MIN timestamp in show_latency_history / MetricsSnapshots() and the timestamp <= end filter stays meaningful. Covered by the new SnapshotTimestampBoundedForMaximumInterval test.

src/impl/ClientTelemetry.cpp
yhmo: Agreed this is a sound trade-off and the growth is transient and self-limiting: entries are pruned as soon as a full batch completes (max_timestamp advances past them), and the set is bounded by the distinct command IDs seen. Advancing the watermark on the fenced abort could drop the unprocessed suffix and risk re-execution, so keeping the marker set is intentional. Leaving as-is.

src/impl/ClientTelemetry.cpp
yhmo: Acknowledged. The revision-change continue fires at most once per cycle and only when AttachChannel bumped the revision during the in-flight RPC, and every re-probe includes a heartbeat RPC to the new endpoint, so it does not busy-spin; under a reconnect storm the cycle rate is bounded by the RPC round-trip. A short settle/coalesce window would be a reasonable future hardening, but we keep the immediate-probe behavior for now since telemetry is best-effort. Deferring.

src/impl/ClientTelemetry.cpp
yhmo: Acknowledged. The repeated result.dump() in the truncation loop is only noticeable with very large error payloads combined with a large max_count; with the default error_max_count=100 and typical payloads it is negligible, and the handler runs outside the plain mutex (only under command_mutex), so RecordOperation is not blocked. If error sizes grow, an incremental-serialization rework would be the right follow-up. Deferring.

src/impl/ClientTelemetry.cpp
```diff
@@ -612,7 +656,16 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
                 }
                 delay = std::max(delay, backed_off);
             }
-            condition.wait_for(lock, std::chrono::milliseconds(delay), [this]() { return stopped; });
+            const auto observed_revision = wait_revision;
+            while (!stopped && observed_revision == wait_revision && delay > 0) {
```
yhmo: src/impl/ClientTelemetry.cpp line:660
Low ---- With the new large-interval support, the unsupported-server backoff cap is effectively defeated for configured intervals above 30 minutes: delay is computed as std::max(heartbeat_interval_ms, backed_off), so for a clamped interval (e.g. int64 max, or any value > 30 min) the probe cadence on an UNIMPLEMENTED server becomes the configured interval (chunked to 24h) instead of the intended 30-minute cap. That slows automatic recovery during a rolling upgrade for exactly the clients #588 wants to wake quickly on channel replacement (the AttachChannel streak reset only helps when a reattach actually occurs). The design doc caps backoff at 30 minutes while also saying it never shortens below the configured interval; for intervals above the cap those two clauses conflict and the code lets the interval win. Consider capping delay at kMaxUnsupportedBackoffMs while unsupported_streak > 0, or documenting the deviation.

yhmo: Fixed: the unsupported-server re-probe cadence is now capped at kMaxUnsupportedBackoffMs even when the configured heartbeat interval is larger. The worker now takes the capped exponential backoff directly (delay = backed_off) instead of std::max(heartbeat_interval_ms, backed_off), so an UNIMPLEMENTED server is re-probed at most every 30 minutes instead of waiting out the full (possibly huge) configured interval; for intervals at or below the cap the exponential ramp is unchanged. Covered by the new UnsupportedServerCapsLargeIntervalProbeCadence test (maximum interval against an UNIMPLEMENTED server stays quiescent at the single initial heartbeat).

test/ut/TestClientTelemetry.cpp
```diff
@@ -474,6 +751,225 @@ TEST(ClientTelemetryTest, ReusesHeartbeatWorkerWhenReconnectRunsInCommandHandler
...
+    }
+    ASSERT_FALSE(manager.IsSupported());
+    ASSERT_EQ(service.heartbeats.load(), 1);
+    std::this_thread::sleep_for(std::chrono::milliseconds(20));
+    manager.Stop();
+    server->Shutdown();
+
+    // The worker must stay quiescent after the single initial probe: the re-probe
+    // cadence is the capped backoff, not the full (maximum) configured interval.
+    EXPECT_EQ(service.heartbeats.load(), 1);
```
yhmo, line 925: test/ut/TestClientTelemetry.cpp line:925
Low ---- This test would also pass against the pre-fix code: with the previous delay = std::max(interval, backed_off), an int64-max interval produced a 24-hour chunked wait, so no second heartbeat fires within the 20 ms window either. As written it verifies quiescence only, not that the re-probe cadence is capped at kMaxUnsupportedBackoffMs, so a future regression of the delay = backed_off change would not be caught. If the cap is the behavior under test, consider a time-controlled assertion (inject/override a clock or expose the computed delay) or a short-interval cadence check; otherwise the name/comment could say it guards against spinning.

src/impl/ClientTelemetry.cpp
```diff
@@ -604,15 +642,33 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
...
+            }
+            const auto observed_revision = wait_revision;
+            while (!stopped && observed_revision == wait_revision && delay > 0) {
+                const auto chunk = std::min(delay, kMaxWaitChunkMs);
+                if (condition.wait_for(
+                        lock, std::chrono::milliseconds(static_cast<int64_t>(chunk)),
+                        [this, observed_revision]() { return stopped || wait_revision != observed_revision; })) {
+                    break;
+                }
+                delay -= chunk;
```
yhmo, line 670: src/impl/ClientTelemetry.cpp line:670
Low ---- The chunked wait subtracts a full `chunk` on every wake where the predicate is still false, whether that wake is a timeout or a spurious/notify wakeup. The `push_config` handler calls `condition.notify_all()` unconditionally (line 1010), even when only `enabled`/`sampling_rate` changed or the interval was *increased* (which deliberately does not bump `wait_revision`). Each such command therefore shortens the remaining sleep by one whole chunk and can fire an early heartbeat; with a default 10s interval every config push collapses the current wait. Consider tracking elapsed time with a steady_clock deadline and only notifying when the interval actually changes, so spurious wakeups cannot shorten the effective cadence.

src/impl/ClientTelemetry.cpp
```diff
@@ -867,10 +965,11 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
                     !payload["heartbeat_interval_ms"].is_number_unsigned()) {
                     throw std::invalid_argument("heartbeat_interval_ms must be an integer");
                 }
-                interval = payload["heartbeat_interval_ms"].get<int64_t>();
-                if (interval <= 0) {
+                interval = payload["heartbeat_interval_ms"].get<uint64_t>();
```
yhmo, line 973: src/impl/ClientTelemetry.cpp line:968
Low ---- Negative heartbeat_interval_ms values are no longer rejected. The old code read the field with get<int64_t>() and threw "heartbeat_interval_ms must be positive" for interval <= 0; the new get<uint64_t>() makes a negative signed integer wrap around to a huge unsigned value (nlohmann get_arithmetic_value does a plain static_cast for number_integer), so the interval == 0 check is skipped and std::min then clamps the wrapped value to kMaxHeartbeatIntervalMs, which is reported as applied. A push_config like {"heartbeat_interval_ms": -1} therefore silently maxes out the heartbeat cadence instead of failing the command. Consider rejecting negatives explicitly (e.g. reject when the value is a signed number_integer that is < 0) and add a negative-value case to the push_config tests.

yhmo: Fixed: push_config now explicitly rejects signed negative heartbeat_interval_ms (is_number_integer && !is_number_unsigned && value < 0) before reading as uint64_t, so {"heartbeat_interval_ms": -1} fails the command with "must be positive" instead of wrapping to a maximum cadence. The PushConfigClampsMaximumInterval test now also pushes -1 and asserts a failed reply with the config unchanged.

test/ut/TestClientTelemetry.cpp
yhmo: Renamed the test to UnsupportedServerLargeIntervalDoesNotSpin: the 20ms assertion can only verify quiescence (both delay = max(interval, backoff) and delay = backed_off stay silent in that window for an int64-max interval), so the name now matches what it actually guards. The backoff-cap behavior itself is covered by code + comment; a time-controlled cadence assertion would need clock injection, deferred as noted.

src/impl/ClientTelemetry.cpp
yhmo: This does not occur: the wait uses condition_variable::wait_for with a predicate, whose predicate overload keeps re-waiting to the same deadline on spurious/notify wakeups while the predicate is false, so a notify_all() that does not bump wait_revision (enabled/sampling_rate changes or an interval increase) cannot shorten the remaining sleep or fire an early heartbeat. Only an interval decrease bumps wait_revision, which is the intended immediate re-probe. No code change.

src/impl/ClientTelemetry.cpp
```diff
@@ -604,15 +642,33 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
...
             if (unsupported_streak > 0) {
                 uint64_t backed_off = std::min(delay, kMaxUnsupportedBackoffMs);
                 for (int index = 0; index < unsupported_streak && backed_off < kMaxUnsupportedBackoffMs; ++index) {
                     backed_off = backed_off > kMaxUnsupportedBackoffMs / 2 ? kMaxUnsupportedBackoffMs : backed_off * 2;
                 }
-                delay = std::max(delay, backed_off);
+                // Cap the probe cadence at the documented backoff ceiling even when the
+                // configured interval is larger, so an UNIMPLEMENTED server is re-probed
+                // within the cap instead of waiting out the full (possibly huge) interval.
+                delay = backed_off;
```
yhmo: src/impl/ClientTelemetry.cpp line:660
Medium ---- Removing the std::max(delay, backed_off) floor means a configured heartbeat interval above 30 minutes now re-probes an UNIMPLEMENTED server every <= 30 min -- more often than configured. The Go SDK reference keeps the floor (nextHeartbeatDelay: "a client configured to heartbeat less often than the cap must not start heartbeating *more* often because the server is rejecting it"), and the design doc says the backoff "never shortening below the configured interval". For a 24h-interval client on an old cluster that is 48x more probe RPCs. AttachChannel already resets unsupported_streak and wakes the wait, so immediate probing on transport replacement is unaffected. I understand the cap fix from the earlier thread; is also dropping the interval floor intentional for parity?

test/ut/TestClientTelemetry.cpp
```diff
@@ -1011,6 +1507,67 @@ TEST(ClientTelemetryTest, PushConfigIsAtomicAndReportsAppliedAndIgnoredKeys) {
...
+    EXPECT_LT(payload.front().at("collection").get<std::string>().size(), oversized_collection.size());
+}
+
+TEST(ClientTelemetryTest, ShowErrorsTruncatesOversizedUtf8FieldsAtCodePointBoundaries) {
+    milvus::TelemetryConfig config;
+    milvus::ClientTelemetryManager manager(config);
+    std::string oversized_utf8;
+    oversized_utf8.reserve(1200000);
+    for (size_t index = 0; index < 400000; ++index) {
+        oversized_utf8 += u8"界";
```
yhmo: test/ut/TestClientTelemetry.cpp line:1555
Low ---- u8 string literals become const char8_t[] in C++20, so `oversized_utf8 += u8"界";` no longer compiles there -- the default build is C++14, but scripts/build.sh and the root CMake explicitly allow CMAKE_CXX_STANDARD 20. Consider appending hex-escaped bytes or a plain char array so the UT target also builds under C++20.

src/impl/ClientTelemetry.cpp
yhmo: Fixed: restored the configured-interval floor, delay = max(delay, backed_off), so the UNIMPLEMENTED backoff caps how often a short-interval client probes but never makes a long-interval client probe more often than configured -- matching the Go SDK nextHeartbeatDelay. AttachChannel already handles immediate probing on transport replacement.

test/ut/TestClientTelemetry.cpp
yhmo: Fixed: the test now appends the UTF-8 bytes of U+754C as a hex-escaped char array ("\xe7\x95\x8c") instead of a u8 string literal, so it compiles under both C++14 and C++20 builds.

src/impl/ClientTelemetry.cpp
```diff
@@ -898,27 +996,33 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
...
-            std::lock_guard<std::mutex> lock(mutex);
+        handlers_["collection_metrics"] = [this](const TelemetryCommand& command) {
+            std::lock_guard<std::mutex> lock(mutex_);
             if (command.payload.empty()) {
-                std::vector<std::string> names(enabled_collections.begin(), enabled_collections.end());
+                std::vector<std::string> names(enabled_collections_.begin(), enabled_collections_.end());
                 std::sort(names.begin(), names.end());
-                nlohmann::json result = {{"enabled_collections", names},
-                                         {"all_collections_enabled", all_collections_enabled}};
+                nlohmann::json result = {{"enabled_collections_", names},
```
yhmo: src/impl/ClientTelemetry.cpp line:1024
High ---- The mechanical `_`-suffix member rename leaked into the JSON string keys of the command-reply payloads, silently changing the operator-visible contract: `collection_metrics` now emits `"enabled_collections_"` / `"all_collections_enabled_"` (this hunk), `get_config` emits `"username_"` / `"enabled_collections_"` / `"all_collections_enabled_"` (lines 1162, 1167-1168), and `show_latency_history` detail mode emits `"snapshots_"` (lines 1203, 1210) instead of `"username"` / `"enabled_collections"` / `"all_collections_enabled"` / `"snapshots"`. These keys are the reply payload surfaced by the server's /config and /history REST endpoints and read by the telemetry WebUI, and the in-repo consumer test/e2e/ClientTelemetryE2E.cpp:349 reads `user_config.value("all_collections_enabled", false)`, which would now fail. Neighboring keys that don't match member names (`"total_snapshots"`, `"db_name"`, `"address"`, `"timestamp"`, `"end_time"`) were untouched, confirming the rename is unintentional. Please revert the string literals to the underscore-free names and add a unit test that pins the exact reply-payload keys so a future rename cannot slip through again.

yhmo: Fixed: reverted the JSON string-literal keys to their underscore-free operator-visible contract (collection_metrics: enabled_collections / all_collections_enabled; get_config: username / enabled_collections / all_collections_enabled; show_latency_history detail: snapshots). Added ReplyPayloadKeysStayUnderscoreFree, which pins the exact reply-payload keys for all three handlers so a future mechanical rename cannot slip through undetected.

test/ut/TestClientTelemetry.cpp
```diff
@@ -474,6 +751,225 @@ TEST(ClientTelemetryTest, ReusesHeartbeatWorkerWhenReconnectRunsInCommandHandler
...
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    }
+    ASSERT_FALSE(manager.IsSupported());
+    ASSERT_EQ(service.heartbeats.load(), 1);
+    std::this_thread::sleep_for(std::chrono::milliseconds(20));
+    manager.Stop();
+    server->Shutdown();
+
+    // The worker must stay quiescent after the single initial probe: the re-probe
+    // cadence is the capped backoff, not the full (maximum) configured interval.
```
yhmo: test/ut/TestClientTelemetry.cpp line:924
Low ---- This comment states that "the re-probe cadence is the capped backoff, not the full (maximum) configured interval", but HeartbeatLoop derives delay = std::max(delay, backed_off) (ClientTelemetry.cpp:660), so for an int64-max interval the delay is the full configured interval (chunked to 24h), not the 30-minute backoff cap. The 20ms assertion can only verify quiescence, which holds under either derivation, so the comment describes a cadence this test cannot establish (without clock injection) and it contradicts the actual delay computation. Consider rewording the comment to state only that the worker stays quiescent in the window, or drop the cadence claim.

src/impl/ClientTelemetry.cpp
```diff
@@ -778,34 +853,38 @@ class ClientTelemetryManager::Impl : public std::enable_shared_from_this<ClientT
...
+            auto reply = handler(command);
+            // The incoming command ID is the protocol correlation key. Extension
+            // handlers_ control the result, but cannot redirect or drop its ACK.
+            reply.command_id = command.command_id;
+            return reply;
         } catch (const std::exception& exception) {
             return FailedReply(command.command_id, exception.what());
         } catch (...) {
-            // Command handlers are extensible application code. Telemetry is best-effort and
+            // Command handlers_ are extensible application code. Telemetry is best-effort and
```
yhmo: src/impl/ClientTelemetry.cpp line:872
Low ---- The mechanical `_`-suffix member rename leaked into prose comments in this file: "Command handlers_ are extensible application code" (here), plus "reconnect handlers_" / "Reuse this worker_ only" (line 524), "retryable stopped_ state" and "leaving worker_running_ without a thread" (lines 575-576), "holding the state mutex_" (line 605), and "the worker_'s shared_ptr keeps Impl alive" (line 680). These read as member names inside sentences where the plain words are meant, which looks stale and can confuse a reader. This is the same class of rename leak as the JSON-key issue already fixed; worth cleaning up so future readers do not trust the identifiers literally.

test/ut/TestClientTelemetry.cpp
yhmo: Fixed: reworded the comment to state only that the worker stays quiescent in the window. After restoring delay = max(delay, backed_off), an int64-max interval is the full chunked interval, so the previous cadence claim was wrong for this test.

src/impl/ClientTelemetry.cpp
yhmo: Fixed: stripped the underscore suffix from all affected prose comments in ClientTelemetry.cpp (handlers / worker / stopped state / worker_running / state mutex / join_in_progress / worker's / collectors / latest_snapshot), matching the JSON-key cleanup. Verified no code member references were touched.

PR #602, opened 2026-09-10
Auto generate release notes from changelog

yhmo:
- Add scripts/extract_changelog.sh, which prints the CHANGELOG.md section for a version and fails when the entry is missing
- Use it in create-release.yaml to build the GitHub release body from the changelog instead of a static two-line body
- Add a validate job to release.yaml that requires the tag commit to be an ancestor of master or a numeric release branch and that the Ubuntu/Windows CI checks passed on that exact commit before any package is built


.github/workflows/release.yaml
```diff
@@ -61,10 +61,65 @@ concurrency:
...
+            done < <(git for-each-ref --format='%(refname:short)' 'refs/remotes/origin/')
+          fi
+          if [[ -z "${approved_branch}" ]]; then
+            echo "Release tag commit ${tag_sha} is not contained in master or a numeric release branch." >&2
+            exit 1
+          fi
+          echo "Release commit ${tag_sha} is contained in ${approved_branch}."
+
+          check_runs="$(gh api "repos/${GITHUB_REPOSITORY}/commits/${tag_sha}/check-runs?per_page=100")"
+          for check_name in "Build and test AMD64 Ubuntu 20.04" "Build and test windows"; do
```
yhmo: .github/workflows/release.yaml line:109
Medium ---- This check requires a successful "Build and test AMD64 Ubuntu 20.04" and "Build and test windows" run on the exact tag commit, but the usual release tag points at a "Prepare for vX.Y.Z" commit that only modifies markdown (CHANGELOG.md, README.md, DEVELOPMENT.md). The main.yaml push path filter explicitly excludes `!**.md`, so that workflow never runs on such commits and the required check runs never exist -- validate would fail and block the release. For example the v3.0.2 prepare commit (d77b3bca) touched only md files and has no "Build and test" check runs. Consider checking the last CI-verified ancestor on the release branch instead, or exempting markdown-only prepare commits from the exact-commit CI requirement.

scripts/extract_changelog.sh
```diff
@@ -0,0 +1,138 @@
...
+  awk -v version="${version}" -v include_heading="${include_heading}" '
+    function heading_level(line,    lvl) {
+      lvl = 0
+      while (substr(line, lvl + 1, 1) == "#") lvl++
+      return lvl
+    }
+    BEGIN {
+      ver = version
+      gsub(/\./, "\\.", ver)
+      pattern = "(^|[^0-9A-Za-z.])" ver "([^0-9A-Za-z.]|$)"
```
yhmo: scripts/extract_changelog.sh line:105
Low ---- The version boundary classes `[^0-9A-Za-z.]` treat `-` and `_` as valid boundaries, so a prerelease-style heading such as `## milvus-sdk-cpp 3.0.3-rc1` would match a search for `3.0.3`, even though the header comment (lines 21-22) only promises protection against longer numeric versions like `3.0.10`. The current changelog has no prerelease headings, but excluding `-`/`_` from the boundary set (or documenting the limitation) would make the matcher robust to that heading style.

.github/workflows/release.yaml
```diff
@@ -61,10 +61,65 @@ concurrency:
...
 jobs:
+  # Validate the release commit before any package is built: the tag must point at
+  # a commit contained in master or a numeric release branch, and the required CI
+  # checks must have passed on that exact commit. This prevents packaging an
+  # unreviewed or non-release-line commit.
+  validate:
+    name: Validate release commit
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
```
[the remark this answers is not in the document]
yhmo: Fixed: the validate job now declares job-level permissions {contents: read, checks: read}, so the gh api check-runs call can read check run conclusions.

.github/workflows/release.yaml
yhmo: Fixed: validate now walks back from the tag commit to the nearest ancestor that carries the required successful checks (bounded to 50 commits). Release tags point at markdown-only Prepare-for-vX.Y.Z commits that main.yaml skips, so the exact-commit requirement would have blocked every release; the nearest CI-verified ancestor (usually the merged code commit) is used instead.

scripts/extract_changelog.sh
yhmo: Fixed: the awk version-boundary class now excludes - and _ on both sides ([^0-9A-Za-z._-]), so a prerelease heading like 3.0.3-rc1 no longer matches a search for 3.0.3; header comment updated to document the prerelease protection. Verified against a changelog containing both 3.0.3-rc1 and 3.0.3 headings.

.github/workflows/release.yaml
```diff
@@ -61,10 +61,86 @@ concurrency:
...
+          fi
+          echo "Release commit ${tag_sha} is contained in ${approved_branch}."
+
+          # Release tags point at a markdown-only "Prepare for vX.Y.Z" commit, which
+          # main.yaml skips (the push path filter excludes **.md), so that commit has
+          # no check runs. Walk back to the nearest ancestor that carries the required
+          # successful checks instead of requiring them on the exact tag commit.
+          sha_to_check="${tag_sha}"
+          verified_sha=""
+          for depth in $(seq 0 50); do
```
yhmo, line 123: .github/workflows/release.yaml line:117
Medium ---- The walk-back loop treats a commit whose required checks FAILED the same as a commit with no check runs: both set all_ok=0 and move to the parent. So if the actual release code commit (or the "Prepare for vX.Y.Z" commit) has a failed "Build and test AMD64 Ubuntu 20.04" or "Build and test windows" run, the loop silently skips past it and accepts the nearest older green ancestor, and packages are still built from the tag (post-failure code). That defeats the stated purpose of this gate ("prevent packaging ... a commit whose required CI checks failed"). Consider failing (exit 1) when an ancestor carries a completed but non-successful run for a required check, and only walking past commits that have no check runs at all or only incomplete ones.

.github/workflows/release.yaml
```diff
@@ -61,10 +61,86 @@ concurrency:
...
+            check_runs="$(gh api "repos/${GITHUB_REPOSITORY}/commits/${sha_to_check}/check-runs?per_page=100")"
+            all_ok=1
+            for check_name in "Build and test AMD64 Ubuntu 20.04" "Build and test windows"; do
```
yhmo: .github/workflows/release.yaml line:120
Low ---- The required check names are hardcoded here and duplicated from the job names in .github/workflows/main.yaml ("Build and test AMD64 ${{ matrix.os.distro }} ${{ matrix.os.version }}" and "Build and test windows"). If the CI matrix or a job name ever changes (for example Ubuntu 20.04 is dropped), the validate job will either block every release with a confusing "No ancestor of <tag> has the required successful checks" error, or stop gating entirely when the matching name disappears. A comment documenting the coupling to main.yaml job names would make the dependency discoverable.

.github/workflows/release.yaml
```diff
@@ -61,10 +61,86 @@ concurrency:
...
+              fi
+            done < <(git for-each-ref --format='%(refname:short)' 'refs/remotes/origin/')
+          fi
+          if [[ -z "${approved_branch}" ]]; then
+            echo "Release tag commit ${tag_sha} is not contained in master or a numeric release branch." >&2
+            exit 1
+          fi
+          echo "Release commit ${tag_sha} is contained in ${approved_branch}."
+
+          # Release tags point at a markdown-only "Prepare for vX.Y.Z" commit, which
```
yhmo: .github/workflows/release.yaml line:111
Low ---- This comment premise is not always true: the "Prepare for v3.0.3" commit (#600, now master head da7acff7) touched tutorial/**/conanfile.py files, which the main.yaml push path filter does not exclude (only **.md, doc/**, .gitignore, codecov.yml, LICENSE, OWNERS, libmilvus.spec.rpkg), so main.yaml did trigger on that commit (a "Build and test" run is currently pending for it). The walk-back logic is robust either way, but the stated reason ("main.yaml skips ... so that commit has no check runs") can mislead a future maintainer about why the ancestor walk-back exists.

.github/workflows/release.yaml
yhmo: Fixed: the walk-back loop now fails hard (exit 1) whenever a required check completed non-successfully on the commit being examined, and only walks past commits with no check runs or only incomplete ones. A failing release commit (or its failing ancestor) can no longer be silently bypassed to an older green commit.

.github/workflows/release.yaml
yhmo: Fixed: added a comment next to the hardcoded check names documenting that they must stay in sync with the job names in main.yaml (including the matrix expansion), so a CI rename/drop is discoverable instead of silently blocking or disabling the gate.

.github/workflows/release.yaml
yhmo: Fixed: reworded the walk-back comment to state the general reason (the tag commit may lack completed successful checks for any reason, e.g. a markdown-only push or a run still in flight) instead of claiming prepare commits are always markdown-only, which the actual Prepare for v3.0.3 commit disproves.

.github/workflows/release.yaml
```diff
@@ -61,10 +61,104 @@ concurrency:
...
+                jq -c --arg name "${check_name}" --arg sha "${sha_to_check}" \
+                  '[.check_runs[] | select(.name == $name and .head_sha == $sha and .conclusion == "success")] | length' \
+                  <<<"${check_runs}"
+              )"
+              if [[ "${successful}" -gt 0 ]]; then
+                continue
+              fi
+              completed_non_success="$(
+                jq -c --arg name "${check_name}" --arg sha "${sha_to_check}" \
+                  '[.check_runs[] | select(.name == $name and .head_sha == $sha and .status == "completed" and .conclusion != "success")] | length' \
```
yhmo: .github/workflows/release.yaml line:136
Medium ---- The walk-back's fail-hard predicate `conclusion != "success"` is broader than a real failure: it also matches `cancelled`, `skipped` and `neutral`. Because main.yaml sets `concurrency: cancel-in-progress: true` (grouped per branch/ref), a superseded push leaves earlier commits with a completed-but-cancelled check run; if such a commit sits in the 50-commit ancestry of a release tag, this step exits 1 and permanently blocks the release even though that commit never failed. `skipped` (path-filtered jobs on a docs-only commit) and `neutral` (`continue-on-error`) have the same effect. Recommend failing hard only on genuine failure conclusions (`failure`, `timed_out`, `startup_failure`, `action_required`) and treating cancelled/skipped/neutral as "not present" so they fall through to the walk-back.

.github/workflows/release.yaml
```diff
@@ -61,10 +61,104 @@ concurrency:
...
+          # The required check names below must stay in sync with the job names in
+          # main.yaml ("Build and test AMD64 ${{ matrix.os.distro }} ${{ matrix.os.version }}"
+          # and "Build and test windows"): renaming or dropping a CI job there breaks
+          # this gate, blocking every release or silently disabling validation.
+          sha_to_check="${tag_sha}"
+          verified_sha=""
+          for depth in $(seq 0 50); do
+            check_runs="$(gh api "repos/${GITHUB_REPOSITORY}/commits/${sha_to_check}/check-runs?per_page=100")"
+            all_present=1
+            for check_name in "Build and test AMD64 Ubuntu 20.04" "Build and test windows"; do
```
yhmo: .github/workflows/release.yaml line:125
Medium ---- The required-check gate only covers "Build and test AMD64 Ubuntu 20.04" and "Build and test windows", but main.yaml also runs "Build and test AMD64 Fedora 39", "Build and test macOS 15" and "Test all with coverage" on the same commit, and the release packages built downstream include Fedora, macOS, ARM, and other Ubuntu targets. A tag commit whose Fedora or macOS CI is red would still pass this gate and publish broken packages for those platforms. Consider requiring every check that guards a shipped package (or at least the Fedora/macOS matrix legs) rather than just the Ubuntu 20.04 + Windows pair, so the gate covers the full set of release artifacts.

.github/workflows/release.yaml
yhmo: Fixed: the walk-back now fails hard only on genuine failure conclusions (failure / timed_out / startup_failure / action_required) and treats cancelled, skipped and neutral runs as not present, so superseded pushes (cancel-in-progress), docs-only skipped jobs and continue-on-error neutral outcomes walk past instead of permanently blocking a release.

.github/workflows/release.yaml
yhmo: Fixed: the required-check list now covers every CI-checked packaging platform: Build and test AMD64 Ubuntu 20.04, Build and test AMD64 Fedora 39, Build and test macOS 15, and Build and test windows (main.yaml currently runs Ubuntu 20.04 + Fedora 39 in its linux matrix, macOS 15, and windows).

.github/workflows/create-release.yaml
```diff
@@ -39,15 +45,23 @@ jobs:
           echo "version=${version}" >> "$GITHUB_OUTPUT"
           echo "release_date=${release_date}" >> "$GITHUB_OUTPUT"
           echo "milvus_minor=${milvus_minor}" >> "$GITHUB_OUTPUT"
+
+          changelog="$(./scripts/extract_changelog.sh "${version}")"
```
yhmo: .github/workflows/create-release.yaml line:49
Medium ---- This step runs `./scripts/extract_changelog.sh` from the checked-out TAG tree, but the script is only added by this PR on the master line. The `2.6` maintenance branch is still active (v2.6.7 released 2026-09-03) and does not contain `scripts/extract_changelog.sh` (verified against the 2.6 branch), and pre-existing tags such as v2.6.7 point at commits whose trees predate the script. Re-pushing or tagging on those lines fails here with "No such file or directory", aborting the whole job before the release is created or the package build is dispatched - a behavior regression vs the old static-body flow, which worked for any vX.Y.Z tag. Consider backporting the script to the release branches or falling back to the static body (Release date / Compatible with Milvus) when the script is not present in the tag's tree.

.github/workflows/release.yaml
```diff
@@ -61,10 +61,109 @@ concurrency:
...
+              fi
+              all_present=0
+            done
+            if [[ "${all_present}" -eq 1 ]]; then
+              verified_sha="${sha_to_check}"
+              break
+            fi
+            sha_to_check="$(git rev-parse "${sha_to_check}~1")"
+          done
+          if [[ -z "${verified_sha}" ]]; then
```
yhmo, line 162: .github/workflows/release.yaml line:156
Low ---- The ancestry + required-checks gate now hard-rejects manual dispatch of release tags on legacy lines (v2.0-v2.5): those commits never ran the current main.yaml job names (the push branch list is only master/2.6/3.0), so the walk-back exhausts the 50-commit bound and the job only surfaces the generic "No ancestor ... has the required successful checks (searched 50 commits)" message. The workflow input docstring just says "Minimum supported: v2.6.3", so the new master/2.6/3.0-only restriction is not discoverable and a blocked dispatch looks like a transient CI miss rather than a permanent restriction. Consider documenting the supported lines next to the gate and printing which lineages were checked when it fails.

.github/workflows/create-release.yaml
yhmo: Fixed: create-release.yaml now falls back to the static two-line body (Release date / Compatible with Milvus) when scripts/extract_changelog.sh is not present in the checked-out tag tree, so legacy tags (e.g. v2.6.x on the 2.6 branch, whose trees predate the script) still create releases and dispatch package builds instead of aborting.

.github/workflows/release.yaml
yhmo: Fixed: documented next to the gate that only lineages running main.yaml (master / 2.6 / 3.0) can be validated, and the failure message now reports the approved lineage and states that older lines (v2.0-v2.5) should be released via the create-release.yaml fallback instead of manual dispatch here.

PR #604, opened 2026-09-14
Expose query/get cost and scanned-bytes stats on QueryResponse

yhmo:
### What this PR does

Aligns `MilvusClientV2::Query()` / `Get()` responses with pymilvus `.extra` output by surfacing the cost / scanned-bytes / cache-hit-ratio metadata that the server already returns in the RPC status `extra_info`.

- `QueryResponse` (alias `GetResponse`) gains `Cost()`, `ScannedRemoteBytes()`, `ScannedTotalBytes()` and `CacheHitRatio()` accessors, mirroring the existing `SearchResponse` surface (defaults `-1`).
- `FillSearchResponseExtraInfo` is templatized over the response type and explicitly instantiated for `SearchResponse` and `QueryResponse`, so query/get reuse the same `extra_info` parsing.
- The `Query` post-hook now fills these fields (covers plain `Query`, `Get`, and session-routed queries); the `Get` empty-ids short-circuit resets them to `-1` to avoid stale values on reused responses.

### Tests

- Unit: `QueryResponse` accessor coverage + `FillSearchResponseExtraInfo` on `QueryResponse` (incl. malformed/comma-float fallback already covered for `SearchResponse`).
- Mocked: `V2QuerySurfacesCostExtraInfo` asserts cost metadata flows through the real `client->Query()` wire path.
- `testing-ut`: 911 passed; `testing-it`: 319 passed; clang-format clean.


src/impl/utils/DqlUtils.cpp
```diff
@@ -109,8 +110,9 @@ AppendOrderByFields(const std::vector<OrderByField>& order_by_fields,
 
 }  // namespace
 
+template <typename T>
 void
-FillSearchResponseExtraInfo(const proto::common::Status& status, SearchResponse& response) {
+FillSearchResponseExtraInfo(const proto::common::Status& status, T& response) {
```
yhmo: src/impl/utils/DqlUtils.cpp line:115
Low ---- The templated FillSearchResponseExtraInfo is now called from the Search, HybridSearch and Query post-hooks (and its QueryResponse instantiation) in MilvusClientV2Impl.cpp, so the "Search" name no longer matches the code it fills. Because every supported response type has to be explicitly instantiated, consider renaming it to a response-agnostic name (for example `FillResponseExtraInfo`) or documenting that it applies to all DQL responses, so a future maintainer adding a third response type does not assume the helper is search-specific.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -2404,6 +2405,10 @@ MilvusClientV2Impl::getWithoutTelemetry(const GetRequest& request, GetResponse&
         }
         response.SetResults(QueryResults{});
         response.SetSessionTs(0);
+        response.SetCost(-1);
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:2408
Low ---- The -1 reset here duplicates the reset block at the top of `FillSearchResponseExtraInfo`, so the default cost/scanned-bytes/cache-hit-ratio values now live in three places: the member initializers in `QueryResponse.h`, the shared helper, and this short-circuit. Consider reusing `FillSearchResponseExtraInfo(proto::common::Status{}, response)` alongside `SetResults`/`SetSessionTs` so the reset defaults have a single source of truth and cannot drift when the defaults change.

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -2404,6 +2405,10 @@ MilvusClientV2Impl::getWithoutTelemetry(const GetRequest& request, GetResponse&
...
+        response.SetScannedRemoteBytes(-1);
+        response.SetScannedTotalBytes(-1);
+        response.SetCacheHitRatio(-1.0f);
```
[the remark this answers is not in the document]

src/impl/utils/DqlUtils.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientV2Impl.cpp

PR #608, opened 2026-09-18
feat: add IndexParam for index creation and deprecate IndexDesc-based request input

yhmo:
## What this PR does / why we need it

Introduce a dedicated `IndexParam` type for index-creation input so the C++ SDK matches the Java/Rust SDKs, where `createIndex` takes `IndexParam` and `describeIndex` returns `IndexDesc`. Previously the C++ SDK reused `IndexDesc` (the describe type) as the create-index input.

## What changes

- New public type `milvus::IndexParam` (`src/include/milvus/types/IndexParam.h`): creation-only fields `field_name`/`index_name`/`index_type`/`metric_type`/`extra_params`.
- `CreateIndexRequest` and `CreateCollectionRequest` (inline index creation) gain `IndexParams()` / `SetIndexParams()` / `WithIndexParams()` / `AddIndexParam()` to build multiple indexes from `IndexParam`.
- The previous `IndexDesc`-based accessors (`Indexes()` / `SetIndexes()` / `WithIndexes()` / `AddIndex()`) are marked `[[deprecated]]` but remain functional: deprecated setters forward into the `IndexParam` list, and the deprecated `Indexes()` getter derives the `IndexDesc` view on demand, so existing callers keep working without a source break (single source of truth).
- `MilvusClientV2::CreateIndex` / `CreateCollection` signatures are unchanged; the impl builds the create-index RPC from `IndexParam` entries.
- Examples, tutorials, and unit/mocked/system tests updated to the new accessors.

## Verification

- `testing-ut`: 923 tests passed
- `testing-it` (mocked): CreateIndex/CreateCollection suites passed
- `testing-st` (system tests) compiles
- All 40 V2 examples build
- All 6 tutorials compile against the PR SDK
- All changed files pass clang-format

## Notes

- `AddFunctionFieldRequest::WithIndex` still uses `IndexDesc` (unchanged).
- The deprecated `Indexes()` getter on `CreateIndexRequest`/`CreateCollectionRequest` returns by value (previously a `const&`). `auto& v = req.Indexes();` no longer compiles and `const auto& v` now binds a snapshot rather than a live view; code should migrate to `IndexParams()`. Updating the deprecated accessors' shape is intentional so the two getters share one source of truth.
- Removing the deprecated accessors is left to a future major version.


src/include/milvus/request/index/CreateIndexRequest.h
```diff
@@ -44,32 +45,64 @@ class MILVUS_SDK_API CreateIndexRequest : public IndexRequestBase<CreateIndexReq
...
+    CreateIndexRequest&
+    AddIndexParam(IndexParam&& index_param);
+
     /**
      * @brief Get indexes.
      * @return the indexes.
+     * @deprecated Use IndexParams() instead.
      */
-    const std::vector<IndexDesc>&
+    [[deprecated("use IndexParams() instead")]] const std::vector<IndexDesc>&
```
yhmo, line 81: src/include/milvus/request/index/CreateIndexRequest.h line:81
Low ---- The class-level @code example above (lines 32-38) still demonstrates the deprecated IndexDesc + WithIndexes() pattern this PR is migrating away from. Users who copy it will compile with deprecation warnings and learn the old API. Please update the example to IndexParam + WithIndexParams()/AddIndexParam() so the header matches the rest of the PR (examples/, tutorial/, tests).

src/impl/request/index/CreateIndexRequest.cpp
```diff
@@ -28,6 +63,10 @@ CreateIndexRequest::Indexes() const {
 void
 CreateIndexRequest::SetIndexes(std::vector<IndexDesc>&& indexes) {
     indexes_ = std::move(indexes);
+    index_params_.clear();
```
yhmo, line 58: src/impl/request/index/CreateIndexRequest.cpp line:66
Low ---- The deprecated setters keep a mirrored std::vector<IndexDesc> that can drift from the IndexParam list actually consumed by MilvusClientV2Impl::CreateIndex(). For example, req.AddIndexParam(p).WithIndexes(v) silently discards p because SetIndexes clears index_params_, while SetIndexParams(v) leaves the deprecated indexes_ mirror stale so Indexes() no longer reflects what is sent over the wire (only index_params_ is serialized). Since both accessors coexist during the deprecation window, consider making the deprecated getter derive from index_params_ (single source of truth) or documenting that the two APIs must not be mixed on the same request.

test/it/v2/TestIndex.cpp
```diff
@@ -102,3 +103,50 @@ TEST_F(UnconnectMilvusMockedTest, CreateIndexRejectsEmptyIndexes) {
     EXPECT_FALSE(status.IsOk());
     EXPECT_EQ(status.Code(), milvus::StatusCode::INVALID_ARGUMENT);
 }
+
+TEST_F(UnconnectMilvusMockedTest, CreateIndexWithIndexParams) {
```
yhmo, line 107: test/it/v2/TestIndex.cpp line:107
Low ---- The compatibility promise of this PR is that the deprecated setters (SetIndexes/WithIndexes/AddIndex) forward into the IndexParam list and produce the same CreateIndex RPC. That forwarding is only exercised at the request-DTO level (size assertions in test/ut/request/TestIndexRequests.cpp), never at the mocked-RPC wire level, so a regression in ConvertToIndexParam (for example a dropped metric type or extra param) would go unnoticed. Consider adding a mocked test that builds the request through the deprecated path and asserts the same INDEX_TYPE/METRIC_TYPE/PARAMS keys as the new-API test above.

tutorial/1_quickstart/src/main.cpp
```diff
@@ -53,7 +53,7 @@ main() {
         milvus::CreateCollectionRequest()
             .WithCollectionName(collection)
             .WithCollectionSchema(schema)
-            .AddIndex(milvus::IndexDesc("embedding", "", milvus::IndexType::AUTOINDEX, milvus::MetricType::COSINE)));
+            .AddIndexParam(milvus::IndexParam("embedding", "", milvus::IndexType::AUTOINDEX, milvus::MetricType::COSINE)));
```
yhmo, line 56: tutorial/1_quickstart/src/main.cpp line:56
High ---- CreateCollectionRequest has no AddIndexParam() member, so this tutorial does not compile. In this PR only CreateIndexRequest gained AddIndexParam()/WithIndexParams(); CreateCollectionRequest still uses IndexDesc with AddIndex()/WithIndexes() (unchanged per the PR notes). The current call chain CreateCollectionRequest().AddIndexParam(milvus::IndexParam(...)) produces "no member named 'AddIndexParam' in 'milvus::CreateCollectionRequest'", which is exactly what the failing CI (Build and test windows / AMD64 Ubuntu 20.04) reports. Revert to .AddIndex(milvus::IndexDesc(...)) here, or create the index via CreateIndexRequest. The same mistake exists in tutorial/2_collection, tutorial/5_dml and tutorial/6_dql.

tutorial/2_collection/src/main.cpp
```diff
@@ -70,7 +70,7 @@ main() {
             .WithCollectionName(collection)
             .WithCollectionSchema(schema)
             .WithConsistencyLevel(milvus::ConsistencyLevel::BOUNDED)
-            .AddIndex(milvus::IndexDesc("embedding", "", milvus::IndexType::AUTOINDEX, milvus::MetricType::COSINE)));
+            .AddIndexParam(milvus::IndexParam("embedding", "", milvus::IndexType::AUTOINDEX, milvus::MetricType::COSINE)));
```
yhmo, line 73: tutorial/2_collection/src/main.cpp line:73
High ---- CreateCollectionRequest has no AddIndexParam() member, so this tutorial does not compile. In this PR only CreateIndexRequest gained AddIndexParam()/WithIndexParams(); CreateCollectionRequest still uses IndexDesc with AddIndex()/WithIndexes() (unchanged per the PR notes). The call chain CreateCollectionRequest().AddIndexParam(milvus::IndexParam(...)) produces "no member named 'AddIndexParam' in 'milvus::CreateCollectionRequest'", which is exactly what the failing CI (Build and test windows / AMD64 Ubuntu 20.04) reports. Revert to .AddIndex(milvus::IndexDesc(...)) here, or create the index via CreateIndexRequest.

tutorial/5_dml/src/main.cpp
```diff
@@ -50,7 +50,7 @@ main() {
         milvus::CreateCollectionRequest()
             .WithCollectionName(collection)
             .WithCollectionSchema(schema)
-            .AddIndex(milvus::IndexDesc("embedding", "", milvus::IndexType::AUTOINDEX, milvus::MetricType::COSINE)));
+            .AddIndexParam(milvus::IndexParam("embedding", "", milvus::IndexType::AUTOINDEX, milvus::MetricType::COSINE)));
```
yhmo, line 53: tutorial/5_dml/src/main.cpp line:53
High ---- CreateCollectionRequest has no AddIndexParam() member, so this tutorial does not compile. In this PR only CreateIndexRequest gained AddIndexParam()/WithIndexParams(); CreateCollectionRequest still uses IndexDesc with AddIndex()/WithIndexes() (unchanged per the PR notes). The call chain CreateCollectionRequest().AddIndexParam(milvus::IndexParam(...)) produces "no member named 'AddIndexParam' in 'milvus::CreateCollectionRequest'", which is exactly what the failing CI (Build and test windows / AMD64 Ubuntu 20.04) reports. Revert to .AddIndex(milvus::IndexDesc(...)) here, or create the index via CreateIndexRequest.

tutorial/6_dql/src/main.cpp
```diff
@@ -95,7 +95,7 @@ main() {
         milvus::CreateCollectionRequest()
             .WithCollectionName(collection)
             .WithCollectionSchema(schema)
-            .AddIndex(milvus::IndexDesc("dense", "", milvus::IndexType::AUTOINDEX, milvus::MetricType::COSINE))
+            .AddIndexParam(milvus::IndexParam("dense", "", milvus::IndexType::AUTOINDEX, milvus::MetricType::COSINE))
```
yhmo: tutorial/6_dql/src/main.cpp line:98
High ---- CreateCollectionRequest has no AddIndexParam() member, so this tutorial does not compile. In this PR only CreateIndexRequest gained AddIndexParam()/WithIndexParams(); CreateCollectionRequest still uses IndexDesc with AddIndex()/WithIndexes() (unchanged per the PR notes). The call chain CreateCollectionRequest().AddIndexParam(milvus::IndexParam("dense", ...)) produces "no member named 'AddIndexParam' in 'milvus::CreateCollectionRequest'", which is exactly what the failing CI (Build and test windows / AMD64 Ubuntu 20.04) reports. Note the neighbouring sparse line still correctly uses .AddIndex(milvus::IndexDesc(...)); revert the dense line to the same pattern, or create the index via CreateIndexRequest.

src/include/milvus/request/index/CreateIndexRequest.h
yhmo: Fixed in a239d11: the @code example now uses IndexParam + WithIndexParams().

src/impl/request/index/CreateIndexRequest.cpp
yhmo: Addressed in a239d11: the mirrored std::vector<IndexDesc> member was removed entirely; the deprecated Indexes() getter now derives the IndexDesc view from index_params_ on demand, so there is a single source of truth and the two accessor families cannot drift.

test/it/v2/TestIndex.cpp
yhmo: Added UnconnectMilvusMockedTest.CreateIndexDeprecatedIndexDescForwarding: it builds the request through the deprecated WithIndexes() path and asserts the wire INDEX_TYPE/METRIC_TYPE/PARAMS keys, same as the new-API test.

tutorial/1_quickstart/src/main.cpp
yhmo: Fixed in a239d11: reverted to CreateCollectionRequest().AddIndex(milvus::IndexDesc(...)) for inline index creation; only CreateIndexRequest gained IndexParam accessors.

tutorial/2_collection/src/main.cpp
yhmo: Fixed in a239d11: reverted to CreateCollectionRequest().AddIndex(milvus::IndexDesc(...)).

tutorial/5_dml/src/main.cpp
yhmo: Fixed in a239d11: reverted to CreateCollectionRequest().AddIndex(milvus::IndexDesc(...)).

tutorial/6_dql/src/main.cpp
yhmo: Fixed in a239d11: the dense line now uses .AddIndex(milvus::IndexDesc(...)) matching the sparse line.

src/impl/request/collection/CreateCollectionRequest.cpp
```diff
@@ -20,6 +20,28 @@
 
 namespace milvus {
 
+namespace {
+
+IndexParam
+ConvertToIndexParam(const IndexDesc& desc) {
```
yhmo: src/impl/request/collection/CreateCollectionRequest.cpp line:26
Medium ---- The ConvertToIndexParam()/ConvertToIndexDesc() helper pair added here is byte-for-byte identical to the pair added in src/impl/request/index/CreateIndexRequest.cpp. Both anonymous-namespace copies must be kept in sync manually; a future field added to IndexParam/IndexDesc (e.g. new type-specific params) would have to be updated in two places, and a mismatch would silently change the wire params between CreateCollection inline-index and CreateIndex paths. Consider hoisting them into a shared internal header (e.g. src/impl/utils/) or a common anonymous helper used by both request classes.

yhmo: Fixed in dce4ee7: hoisted the conversion pair into a shared header src/impl/utils/IndexUtils.h (inline ConvertToIndexParam/ConvertToIndexDesc in namespace milvus); both CreateCollectionRequest.cpp and CreateIndexRequest.cpp now include it, so the two conversions stay in one place.

src/impl/request/index/CreateIndexRequest.cpp
```diff
@@ -18,16 +18,47 @@
...
+CreateIndexRequest&
+CreateIndexRequest::AddIndexParam(IndexParam&& index_param) {
+    index_params_.emplace_back(std::move(index_param));
+    return *this;
+}
+
 const std::vector<IndexDesc>&
 CreateIndexRequest::Indexes() const {
-    return indexes_;
+    indexes_cache_.clear();
```
yhmo, line 49: src/impl/request/index/CreateIndexRequest.cpp line:49
Low ---- The deprecated Indexes() getter now clears and rebuilds a mutable scratch cache on every call and returns a reference to it. Previously the returned reference pointed at the stable stored member, so it stayed valid across calls; now any second Indexes() call invalidates the first reference. The const getter also mutates a mutable member, which is a data race if a request object is read concurrently. The identical pattern was introduced in CreateCollectionRequest.cpp:211. Since this is the compatibility path that existing callers still rely on, consider returning the vector by value from the deprecated getter (lifetime-safe for const auto& bindings) or documenting that the reference is only valid until the next Indexes() call.

test/it/v2/TestIndex.cpp
```diff
@@ -102,3 +103,79 @@ TEST_F(UnconnectMilvusMockedTest, CreateIndexRejectsEmptyIndexes) {
...
+    auto status = client->CreateIndex(milvus::CreateIndexRequest()
+                                          .WithDatabaseName("db")
+                                          .WithCollectionName("collection")
+                                          .AddIndexParam(std::move(param_vec))
+                                          .AddIndexParam(std::move(param_scalar))
+                                          .WithSync(false));
+    EXPECT_TRUE(status.IsOk());
+}
+
+TEST_F(UnconnectMilvusMockedTest, CreateIndexDeprecatedIndexDescForwarding) {
```
yhmo, line 154: test/it/v2/TestIndex.cpp line:154
Low ---- This mocked test covers the deprecated IndexDesc-to-wire forwarding for CreateIndexRequest only. The CreateCollectionRequest inline-index path (MilvusClientV2Impl::CreateCollection now reads IndexParams() and issues one CreateIndex RPC per entry) and its deprecated WithIndexes()/AddIndex() forwarding have no mocked-RPC coverage, only DTO-level size assertions. Since the compatibility promise in this PR explicitly covers CreateCollection too, consider a sibling mocked test that builds the collection request through the deprecated path and asserts the same INDEX_TYPE/METRIC_TYPE/PARAMS wire keys as this test.

src/include/milvus/types/IndexParam.h
```diff
@@ -0,0 +1,137 @@
...
+     * @param field_name field name which the index belong to
+     * @param index_name index name
+     * @param index_type index type see IndexType
+     * @param metric_type metric type see MetricType, no need to set this for scalar field index
+     */
+    IndexParam(std::string field_name, std::string index_name, milvus::IndexType index_type,
+               milvus::MetricType metric_type = milvus::MetricType::INVALID);
+
+    /**
+     * @brief Filed name which the index belong to.
```
yhmo, line 52: src/include/milvus/types/IndexParam.h line:52
Low ---- Doc typo in the new public header: "Filed name which the index belong to" should be "Field name ...". Since IndexParam is now the documented way to create indexes, this typo is likely to be copied into downstream examples and docs (it was inherited from the same wording in IndexDesc.h).

src/impl/request/index/CreateIndexRequest.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

test/it/v2/TestIndex.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/include/milvus/types/IndexParam.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

examples/src/v2/general.cpp
```diff
@@ -133,10 +133,10 @@ main(int argc, char* argv[]) {
         milvus::FieldSchema(field_face, milvus::DataType::FLOAT_VECTOR, "face signature").WithDimension(dimension));
 
     // define indexes
-    milvus::IndexDesc index_vector(field_face, "", milvus::IndexType::IVF_FLAT, milvus::MetricType::COSINE);
-    index_vector.AddExtraParam(milvus::NLIST, "100");
-    milvus::IndexDesc index_sort(field_age, "", milvus::IndexType::STL_SORT);
-    milvus::IndexDesc index_varchar(field_name, "", milvus::IndexType::TRIE);
+    milvus::IndexParam index_vector(field_face, "", milvus::IndexType::IVF_FLAT, milvus::MetricType::COSINE);
+    index_vector.AddExtraParam("nlist", "1024");
```
yhmo: examples/src/v2/general.cpp line:137
Low ---- This mechanical IndexDesc -> IndexParam rename also changed `AddExtraParam(milvus::NLIST, "100")` into `AddExtraParam("nlist", "1024")`. That silently alters the example: IVF_FLAT is now built with nlist=1024 instead of 100, and it drops the public `milvus::NLIST` constant that sibling examples still use (e.g. `struct_field.cpp` and `NPROBE` at general.cpp:360). If the value change is intentional, a note would help; otherwise please restore `milvus::NLIST` and "100".

src/impl/MilvusClientV2Impl.cpp
```diff
@@ -4159,26 +4160,44 @@ MilvusClientV2Impl::RemovePrivilegesFromGroup(const RemovePrivilegesFromGroupReq
 Status
 MilvusClientV2Impl::createIndex(const std::string& db_name, const std::string& collection_name, const IndexDesc& desc,
                                 bool sync, int64_t timeout_ms) {
-    auto pre = [&db_name, &collection_name, &desc](proto::milvus::CreateIndexRequest& rpc_request) {
+    return createIndex(db_name, collection_name, desc.FieldName(), desc.IndexName(), desc.IndexType(),
```
yhmo: src/impl/MilvusClientV2Impl.cpp line:4163
Low ---- After migrating all internal callers to the `IndexParam` overload, this private `createIndex(db_name, collection_name, const IndexDesc&, bool, int64_t)` overload no longer has any caller: CreateCollection and CreateIndex both go through the `IndexParam` variant now. This overload (plus its declaration in MilvusClientV2Impl.h) is dead code; consider removing it so the two overloads cannot drift apart.

src/include/milvus/request/index/CreateIndexRequest.h
```diff
@@ -44,32 +45,64 @@ class MILVUS_SDK_API CreateIndexRequest : public IndexRequestBase<CreateIndexReq
...
+    CreateIndexRequest&
+    AddIndexParam(IndexParam&& index_param);
+
     /**
      * @brief Get indexes.
      * @return the indexes.
+     * @deprecated Use IndexParams() instead.
      */
-    const std::vector<IndexDesc>&
+    [[deprecated("use IndexParams() instead")]] std::vector<IndexDesc>
```
yhmo, line 81: src/include/milvus/request/index/CreateIndexRequest.h line:81
Low ---- The deprecated `Indexes()` getter changes return type from `const std::vector<IndexDesc>&` to `std::vector<IndexDesc>` (by value). Most callers are unaffected, but `auto& v = req.Indexes();` no longer compiles, and `const auto& v = req.Indexes();` now binds a snapshot copy instead of a live view into the request. The PR description says existing callers keep working "without a source break"; this return-type change is itself a (deprecated) source break for those call patterns, so it may be worth acknowledging explicitly in the deprecation/release notes.

examples/src/v2/general.cpp
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/include/milvus/request/index/CreateIndexRequest.h
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/impl/MilvusClientV2Impl.cpp

PR #609, opened 2026-09-21
doc: tag Zilliz Cloud only APIs with @zillizCloudOnly

yhmo:


Doxyfile
```diff
@@ -12,6 +12,8 @@ RECURSIVE              = YES
 EXCLUDE_PATTERNS       = */thirdparty/*
 FILE_PATTERNS          = *.h *.hpp
 
+ALIASES                += "zillizCloudOnly=\par \b Zilliz Cloud only"
```
yhmo, line 17: Doxyfile line:15
Low ---- Consider documenting the @zillizCloudOnly convention next to this alias so future Zilliz Cloud-only APIs get tagged. This line is currently the only trace of the convention in the repo: the Doxyfile header comment, AGENTS.md, and DEVELOPMENT.md never mention it, and CI only checks that index.html exists, so an untagged API would silently lose the "Zilliz Cloud only" badge from the published docs. The Java SDK uses the same @zillizCloudOnly tag (registered via the javadoc -tag option in pom.xml), so a one-line note here would anchor the cross-SDK convention for contributors.

yhmo: Added a note above the ALIASES line in Doxyfile documenting that Zilliz Cloud only APIs should carry @zillizCloudOnly, and that it mirrors the Java SDK tag. Force-pushed as 1f7bdcd1.

